PR #5, opened 2020-04-23
Feature/api group manage joined cleaned

knight212:
- Group manage api integration
- Group joined api integration


smadbe:
When I visit the managed group page from an external page, for instance an activity, I still have this empty (only the player icon) page that I showed you. Don't you?

PR #13, opened 2020-05-06
Feature/dev group module

knight212:
- [x] in the menu, "groups you manage", add 2 groups: "group 50" and "group 51", with "blue arrow" linking respectively to `/groups/managed/50` and `/groups/managed/51` 
- page `/groups/managed/:id` shows:
  - [x] on the right pane (page excluding the left menu and the breadcrumb), only display components which can be implemented and show correct information.
  - [x] header: group name (so no link/img/url/address)
  - [x] between header and tabs:  the "pending requests" for this group:
    - [x] list requests (`GET /groups/50/requests`), **only keep** those with `action=="join_request_created"`
    - [x] if the list is empty, do not display the blue panel at all 
    - [x] remove picture, long text 
    - [x] if "grade" is not null for the joining_user, display "Grade: <value>" where "Terminale" is displayed in the designs
    - [x] Hide "this group only"/"all subgroups" 
    - [x] add sorting to this list (backend-side, cfr the `sort` in the API)

### DIFF
--- a/projects/dev/src/app/group/group-content/group-content.component.ts
+++ b/projects/dev/src/app/group/group-content/group-content.component.ts
@@ -1,0 +1,46 @@
+import { Component, OnInit, ElementRef, EventEmitter, Input, Output } from '@angular/core';
+
+@Component({
+  selector: 'app-group-content',
+  templateUrl: './group-content.component.html',
+  styleUrls: ['./group-content.component.scss']
+})
+export class GroupContentComponent implements OnInit {
+
+  @Input() data;
+  @Input() empty;
+
+  @Output() expandWholeWidth = new EventEmitter<void>();
+
+  constructor(
+    private elementRef: ElementRef
+  ) { }
+
+  ngOnInit() {
+  }
+
+  onTabChange(e) {
+    const tabs = this.elementRef.nativeElement.querySelectorAll('.mat-tab-labels .mat-tab-label');
+    const activeTab = this.elementRef.nativeElement.querySelector('.mat-tab-labels .mat-tab-label.mat-tab-label-active');
+    tabs.forEach((tab) => {
+      tab.classList.remove('mat-tab-label-before-active');
+    });
+    
+    let i;
+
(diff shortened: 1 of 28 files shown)


knight212:
> If I go to page 51 and then 50, I have these requests:
> 
> ```
> GEThttp://127.0.0.1:8000/groups/51
> GEThttp://127.0.0.1:8000/groups/51/members?sort=-member_since,id
> GEThttp://127.0.0.1:8000/groups/51/requests?sort=-at,member_id
> <clicking on 50>
> GEThttp://127.0.0.1:8000/groups/50
> GEThttp://127.0.0.1:8000/groups/51/members?sort=-member_since,id
> GEThttp://127.0.0.1:8000/groups/50/requests?sort=-at,member_id
> ```
> 
> Why do the request before last is to 51? Because members are not implemented and it is left as it was?

Yes, it should be commented but didn't do it yet.

projects/dev/src/app/group/group-content/group-content.component.ts
```diff
@@ -0,0 +1,46 @@
...
+    });
+    
+    let i;
+
+    for (i = 0 ; i < tabs.length ; i++) {
+      if (tabs[i] === activeTab) {
+        break;
+      }
+    }
+
```
smadbe, line 36: As you have lodash installed, do it properly with lodash.
Something like `let i = _.findIndex(tabs, activeTab)` (to be tested)

knight212: Oh, my bad.
Why did I do that? It can be done with pure javascript itself.

knight212: @smadbe 
I cleaned the code.

smadbe: There are 5 places where there is the same kind of code, fix all of them. If they were not part of this PR, fix the other ones in a branch from master directly.

PR #17, opened 2020-05-08
Feature/project env

knight212:


src/environments/environment.prod.ts
```diff
@@ -1,3 +1,4 @@
 export const environment = {
-  production: true
+  production: true,
+  apiUrl: 'http://dev.algorea.org/api'
```
smadbe, line 3: Should even be api as it runs on the same domain. (will handle http/https better)
```suggestion
  apiUrl: 'api'
```

knight212: Okay

smadbe: You could have just commit the proposal just above ("commit suggestion")

PR #18, opened 2020-05-11
Code lint

knight212:
1. Code linting using Prettier
2. Remove unnecessary lines, imports, comments, and `console.log`s in `core` library and `dev` sub app.


smadbe:
I would expect `ng lint` not to return a single error anymore. 
It seems their recommendations is to use single quotes instead of double quotes, it is probably in our own interest to follow their "standards".

knight212:
I am using the Prettier so I will change the linter to use double quotes.

knight212:
@smadbe 
Fixed the lint and there is no error if you run `ng lint`.
Please make sure you run `ng build core` before you run `ng lint`.
This branch was checked out from `master` so it doesn't have functionality for accept/reject.

smadbe:
And why not using "tslint" which seems more standard for Angular ? (and which is handling a specific linting for unit tests, e2e, ...)

knight212:
@smadbe 
The new commits are using tslint so you can review.
:)

tslint.json
```diff
@@ -29,7 +29,7 @@
     "max-classes-per-file": false,
     "max-line-length": [
       true,
-      140
+      256
```
smadbe, line 32: So you made 3 changes to the default config.  Is there good reason for changing them?
* For the line length, "140" looks good enough I would say.   In the backend, we set it to 140. Looks like a good trade-off.
* `no-switch-case-fall-through` looks a good way to avoid difficult-to-debug bugs with `/* falls through */` as a way to enforce it for cases where you know what you are doing.
* `no-output-on-prefix`: I have no opinion.

knight212: - The first was changed because it needs for multiple line test as it was required for frontend design.
- The second was to optimize the code but if you want I can enable it, no problem.
- The third one was for output event emitters with on as prefix. I use this to emphasize the events.

PR #19, opened 2020-05-14
initial circle ci config for linting

smadbe:
Linting which will be run for every pull-request and merge on Github. Will be enforced as soon as the ongoing branches without linting are merged.


knight212:
@smadbe 
Could you downgrade the node version to 12.16?
To use the LTS version of Node.js.

PR #58, opened 2020-07-12
keep only relevant part of the UI in the topbar

smadbe:


smadbe:
Merged manually (as GH was down)

PR #80, opened 2020-09-16
fixes #78

### LINKED ISSUE #78: Bug: "Pending request" not filtered correctly
The "pending requests" shown on a manager's groups seem not to be filtered by group id. As a consequence, the reject/accept request are sent on the wrong group.

Fix this by filtering on the group. Check if the service supports it (https://france-ioi.github.io/algorea-devdoc/api/), if not it should probably be done in the service. 

To reproduce, use "devapi" with "token20" you'll see you have the same requests for both groups (while there are only for group 50 in the db), (as a consequence, the requests on group 51 fail)

AyeRlS:
Fixes `group_id` filter not being added to the url params when looking for users requests.


smadbe (approved):
👍

PR #81, opened 2020-09-16
Add "create group invitation" http-service

GabrielVidal1:
Added the http-service for the group-invite-users component.
This service is calling the [groupInvitationsCreate](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupInvitationsCreate) in the Backend.

### DIFF
--- a/src/app/modules/group/http-services/create-group-invitations.service.ts
+++ b/src/app/modules/group/http-services/create-group-invitations.service.ts
@@ -2,9 +2,15 @@ import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs';
 import { environment } from 'src/environments/environment';
 import { HttpClient } from '@angular/common/http';
-import { ActionResponse, successData, objectToMap } from 'src/app/shared/http-services/action-response';
+import { ActionResponse, successData } from 'src/app/shared/http-services/action-response';
 import { map } from 'rxjs/operators';
 
+export enum InvitationResult {
+  success,
+  error,
+  already_invited,
+  not_found,
+}
 
 @Injectable({
   providedIn: 'root'
@@ -16,7 +22,7 @@ export class CreateGroupInvitationsService {
   createInvitations(
     groupId: string,
     logins : string[]
-  ) : Observable<Map<string, any>>
+  ) : Observable<Map<string, InvitationResult>>
   {
     return this.http
       .post<ActionResponse<Object>>(
@@ -24,7 +30,16 @@ export class CreateGroupInvitationsService {
         {logins: logins}, {})
       .pipe(
         map(successData),
-        map(objectToMap)
+        map(function (data: Object): Map<string, InvitationResult> {
+          return new Map<string, InvitationResult>(
+            Object.entries(data).map(
+              ([key, value]) => {
+                const result = InvitationResult[value as keyof typeof InvitationResult];
+                if (result == undefined)
+                  throw new Error(`Invitation of user ${key} returned an unexpected result`);
+                return [key, value];
+          }));
+        })
       );
   }
 }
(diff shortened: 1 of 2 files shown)


smadbe (changes requested):
Very good.  👍  
I would suggest a change to make you component cleaner (which was not done in `RequestActionsService` but probably should as well):
It is not really great to obtain a `Map<string,any>` for the complement as any could be anything. After your two rx "map(..)"'s, you should again your `Map<string, any>` to `Map<string, InvitationResult>` where `InvitationResult` would be a union type of what is supposed in the service (the enum  [cycle, invalid, success, unchanged, not_found]) but reinterpreted for our needs, so for instance mapping to 'success'|'error'|'already_invited'|'not_found', and throwing an error if you get something unexpected.

src/app/modules/group/http-services/create-group-invitations.service.ts
```diff
@@ -2,9 +2,15 @@ import { Injectable } from '@angular/core';
...
-import { ActionResponse, successData, objectToMap } from 'src/app/shared/http-services/action-response';
+import { ActionResponse, successData } from 'src/app/shared/http-services/action-response';
 import { map } from 'rxjs/operators';
 
+export enum InvitationResult {
+  success,
+  error,
+  already_invited,
+  not_found,
+}
```
smadbe, line 13: I was suggesting an union type but in this case probably an enum is better, right 👍 
However, enum options should be in PascalCase, so `Success`,  `AlreadyInvited`, ...

src/app/modules/group/http-services/create-group-invitations.service.ts
```diff
@@ -16,15 +22,24 @@ export class CreateGroupInvitationsService {
...
         map(successData),
-        map(objectToMap)
+        map(function (data: Object): Map<string, InvitationResult> {
+          return new Map<string, InvitationResult>(
+            Object.entries(data).map(
+              ([key, value]) => {
+                const result = InvitationResult[value as keyof typeof InvitationResult];
+                if (result == undefined)
+                  throw new Error(`Invitation of user ${key} returned an unexpected result`);
+                return [key, value];
```
smadbe, line 40: Ah I didn't know this weird syntax `InvitationResult[value as keyof typeof InvitationResult];`
Actually, I have not been clear but the service returns (following the API) `enum: [cycle, invalid, success, unchanged, not_found]` so you have to map these input values to the enum (cycle and invalid can be both mapped to error, the UI does not need to know more). So I expect more a switch-case which return the different values of the enum.

src/app/modules/group/http-services/create-group-invitations.service.ts
```diff
@@ -16,15 +22,24 @@ export class CreateGroupInvitationsService {
...
       .pipe(
         map(successData),
-        map(objectToMap)
+        map(function (data: Object): Map<string, InvitationResult> {
+          return new Map<string, InvitationResult>(
+            Object.entries(data).map(
+              ([key, value]) => {
+                const result = InvitationResult[value as keyof typeof InvitationResult];
+                if (result == undefined)
+                  throw new Error(`Invitation of user ${key} returned an unexpected result`);
```
smadbe, line 39: in your error, you should print the value as well.

smadbe (changes requested):

src/app/modules/group/http-services/create-group-invitations.service.ts
GabrielVidal1: I used `JSON.stringify(value)` because the value is of type `any`

smadbe: Great 👍

smadbe (approved):
👍

PR #91, opened 2020-09-18
Add "get group managers" service

GabrielVidal1:
Added the http-service for the get-group-managers component.
This service is calling the [groupsMembersView](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupsMembersView) in the Backend.


smadbe (approved):
👍

PR #92, opened 2020-09-18
Initial version of breadcrumbs

smadbe:
Initial version of the bit-less-static breadcrumbs which can be modified by the pages displaying content.
To be reviewed commit per commit (easier).

(while the item breadcrumb service is implemented, it is still not used by the item details page, it will be for a future PR)


AyeRlS (approved):
👍

GabrielVidal1 (approved):

PR #94, opened 2020-09-21
Load item breadcrumbs

smadbe:
Replace static breadcrumbs for items to the service call.

Note that: 
* 3 errors are still not well handled (requires services being implemented on the backend)
* `breadcumbRequest` is a separate fct (and not in fetchItem) as the next step is to add other parallel calls

(note the first commit is things that may have been noticed by the previous review)

### DIFF
--- a/src/app/shared/services/nav-types.ts
+++ b/src/app/shared/services/nav-types.ts
@@ -39,3 +39,7 @@ export function itemFromDetailParams(params: ParamMap): NavItem|undefined {
     itemPath: pathAsString === null || pathAsString.length === 0 ? [] : pathAsString.split(','),
   };
 }
+
+export function pathGiven(params: ParamMap): boolean {
+  return params.has(pathParamName);
+}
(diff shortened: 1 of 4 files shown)


src/app/shared/services/nav-types.ts
```diff
@@ -39,3 +39,7 @@ export function itemFromDetailParams(params: ParamMap): NavItem|undefined {
     itemPath: pathAsString === null || pathAsString.length === 0 ? [] : pathAsString.split(','),
   };
 }
+
+export function pathGiven(params: ParamMap): boolean {
```
AyeRlS, line 43:
```suggestion
export function isPathGiven(params: ParamMap): boolean {
```
That would require some other changes, but I feel like it would make what the fonction does clearer

GabrielVidal1 (approved):
👍

src/app/shared/services/nav-types.ts
smadbe: 👍  done

PR #99, opened 2020-09-24
Add progress-level component

GabrielVidal1:
Added the progress-level component that will be used by the group manager list component to show different levels of permission.

### DIFF
--- a/src/app/modules/shared-components/components/progress-level/progress-level.component.html
+++ b/src/app/modules/shared-components/components/progress-level/progress-level.component.html
@@ -1,0 +1,6 @@
+<div class="progress-level-container">
+  <ng-container *ngFor="let i of totalArray()">
+    <span class="progress-level-item" [ngClass]="{active: i < filled, revert: theme === 'revert'}"></span>
+    <span *ngIf="i + 1 < total" class="progress-level-item-link" [ngClass]="{active: i + 1 < filled, revert: theme === 'revert'}"></span>
+  </ng-container>
+</div>
--- a/src/app/modules/shared-components/components/progress-level/progress-level.component.html
+++ b/src/app/modules/shared-components/components/progress-level/progress-level.component.html
@@ -1,6 +1,6 @@
 <div class="progress-level-container">
-  <ng-container *ngFor="let i of totalArray()">
-    <span class="progress-level-item" [ngClass]="{active: i < filled, revert: theme === 'revert'}"></span>
-    <span *ngIf="i + 1 < total" class="progress-level-item-link" [ngClass]="{active: i + 1 < filled, revert: theme === 'revert'}"></span>
+  <ng-container *ngFor="let i of totalArray">
+    <span class="progress-level-item" [ngClass]="{active: i < value, revert: theme === 'revert'}"></span>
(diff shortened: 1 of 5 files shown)


src/app/modules/shared-components/components/progress-level/progress-level.component.html
```diff
@@ -0,0 +1,6 @@
+<div class="progress-level-container">
+  <ng-container *ngFor="let i of totalArray()">
```
smadbe, line 2: You cannot call a function from your html, otherwise angular call it each time there is a tiny unrelated change on the page. Add a console log in your function and you'll see.

src/app/modules/shared-components/components/progress-level/progress-level.component.ts
```diff
@@ -0,0 +1,21 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'alg-progress-level',
+  templateUrl: './progress-level.component.html',
+  styleUrls: ['./progress-level.component.scss']
+})
+export class ProgressLevelComponent {
+
+  @Input() size = 12;
```
smadbe, line 10: not used? if not, remove

src/app/modules/shared-components/components/progress-level/progress-level.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  @Input() theme : 'default' | 'revert' = 'default';
+  @Input() filled = 0;
```
smadbe, line 12: This attribute name is not clear... from what I understand it is just the number of dots filled... ok so it may just be called "value". And if you call "total" -> "maximum", you have the same kind of properties as a numeric slider... which is actually quite close to what this widget is.

GabrielVidal1: fiixed

src/app/modules/shared-components/components/progress-level/progress-level.component.html
GabrielVidal1: ok now I store the array in the component, and initialize it in the constructor.

src/app/modules/shared-components/components/progress-level/progress-level.component.ts
GabrielVidal1: fixed

src/app/modules/shared-components/components/progress-level/progress-level.component.ts
```diff
@@ -9,13 +9,13 @@ export class ProgressLevelComponent {
 
   @Input() size = 12;
   @Input() theme : 'default' | 'revert' = 'default';
-  @Input() filled = 0;
-  @Input() total = 4;
+  @Input() value = 0;
+  @Input() maximum = 4;
 
-  constructor() {}
+  totalArray : number[] = [];
```
smadbe, line 14: Detail: No need to put an initial value as you initial value in the constructor.
You should add explaining what this strange variable is.
```suggestion
  totalArray : number[]; // array containing [0, 1, 2, ...] as the template cannot iterate by itself
```

smadbe (changes requested):

src/app/modules/shared-components/components/progress-level/progress-level.component.ts
smadbe: You just have "commit suggestion" directly in GH.

smadbe (approved):

Issue #85, opened 2020-09-17
Pending request: change the loading behavior


smadbe:
Change the way the blue panel "pending request" is displayed:
- always display the "pending request"  panel when displaying a group IF the group has `is_public=true`.
- keep the blue panel collapsed by default
- while loading, grey-out the "this-group/subgroups" switch and display a loading spinner
- when loaded, re-enable the switch and: 
  * if loading error: display an error message in place of the table 
  * if empty: display a message that it is empty in place of the table 
  * otherwise, display the table and un-collapse the panel

PR #101, opened 2020-09-28
Subgroup table

### LINKED ISSUE #86: Pending request: add this group only / all subgroups switch, step 2: adapt table display
As it is done in the design branch (`http://dev.algorea.org/branch/design/`), add to the "pending request" box the switch "group only / all subgroups".

Step2: 
This parameter is now supported by the backend (see API) and has to be added in the service.
When "include subgroups" switch is enabled or disabled:
- redisplay the loading (done in #85) spinner, grey out the switch while processing
- use the same error/empty behavior when response is received
- if "including subgroups" is enabled, display an additional column (first) with the group name (group that the user wants to join)

Make sure you do not break any of the current behaviors.

AyeRlS:
Fixes #86
Adapt table display when subgroups' pending requests are shown as requested.

Please note that the failing tests will be fixed once the branch is merge with "subgroup-loading".

### DIFF
--- a/src/app/modules/group/components/pending-request/pending-request.component.html
+++ b/src/app/modules/group/components/pending-request/pending-request.component.html
@@ -71,8 +71,10 @@
             <p-tableCheckbox [value]="rowData" [disabled]="ongoingActivity !== Activity.None"></p-tableCheckbox>
           </td>
           <td *ngFor="let col of columns">
-            <ng-container *ngIf="col.field === 'group.name'">
-              {{ rowData.group.name }}
+            <ng-container *ngIf="col.field === 'group.name'" >
+              <span style="white-space: initial;">
+                {{ rowData.group.name }}
+              </span>
             </ng-container>
             <ng-container *ngIf="col.field === 'at'">
               {{ rowData.at | date: "d/MM/y" }}
--- a/src/app/modules/group/components/pending-request/pending-request.component.html
+++ b/src/app/modules/group/components/pending-request/pending-request.component.html
@@ -16,8 +16,8 @@
   <div class="pending-request-container" *ngIf="status !== 'loading'; else loading" [ngSwitch]="status">
     <alg-grid
       *ngSwitchCase="'loaded'"
-      [selectedColumns]="columns"
-      [columns]="columns"
+      [selectedColumns]="getGroupGrid()"
+      [columns]="getGroupGrid()"
       [data]="requests"
       [groupInfo]="panel"
       [(selection)]="selection"
@@ -47,6 +47,16 @@
                 <p-sortIcon [field]="col.field"></p-sortIcon>
               </div>
             </th>
+            <th
+              *ngIf="col.field === 'group.name'"
+              [pSortableColumn]="col.field"
+              style="width: 13.3333rem; background-color: #d3dbe7;"
+            >
+              <div class="header-container">
+                {{ col.header }}
+                <p-sortIcon [field]="col.field"></p-sortIcon>
+              </div>
+            </th>
           </ng-container>
         </tr>
       </ng-template>
@@ -61,6 +71,9 @@
             <p-tableCheckbox [value]="rowData" [disabled]="ongoingActivity !== Activity.None"></p-tableCheckbox>
           </td>
           <td *ngFor="let col of columns">
+            <ng-container *ngIf="col.field === 'group.name'">
+              {{ rowData.group.name }}
+            </ng-container>
             <ng-container *ngIf="col.field === 'at'">
               {{ rowData.at | date: "d/MM/y" }}
             </ng-container>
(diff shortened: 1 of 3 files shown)


smadbe:
I wait for the tests to pass before starting the review.

src/app/modules/group/components/pending-request/pending-request.component.html
```diff
@@ -16,8 +16,8 @@
   <div class="pending-request-container" *ngIf="status !== 'loading'; else loading" [ngSwitch]="status">
     <alg-grid
       *ngSwitchCase="'loaded'"
-      [selectedColumns]="columns"
-      [columns]="columns"
+      [selectedColumns]="getGroupGrid()"
+      [columns]="getGroupGrid()"
```
smadbe, line 20: You cannot depend on a function in Angular templates, otherwise they are called hundred times, each time angular has to refresh anything on the view.

smadbe (changes requested):
The requested behavior of #85 does not seem be applied here: 

> when loaded, re-enable the switch and:
>  ... display the table and un-collapse the panel

src/app/modules/group/components/pending-request/pending-request.component.html
AyeRlS: fixed

src/app/modules/group/components/pending-request/pending-request.component.ts
```diff
@@ -50,6 +50,7 @@ export class PendingRequestComponent implements OnInit, OnChanges {
     { field: 'user.login', header: 'USER' },
     { field: 'at', header: 'REQUESTED ON' },
   ];
+  groupColumn = { field: 'group.name', header: 'GROUP'};
```
smadbe, line 53: These two variables `columns` and `groupColumn` aren't the same. `columns` is an instance variable that may change and has a default value. However, you use `groupColumn` as a constant from what I see. So it should be a constant (moved over the class definition).

src/app/modules/group/components/pending-request/pending-request.component.ts
```diff
@@ -193,6 +197,11 @@ export class PendingRequestComponent implements OnInit, OnChanges {
 
   onSubgroupSwitch(selectedIdx: number) {
     this.includeSubgroup = this.subgroupSwitchItems[selectedIdx].includeSubgroup;
+
+    this.columns = this.columns.filter(elm => elm !== this.groupColumn);
```
smadbe, line 201: Beware that ===/!== does not work on objects in JS. Here it works only because you use your groupColumn as a constant and so the reference is the same. But actually if you move it as a clear constant,  probably that is ok for here as well.

src/app/modules/group/components/pending-request/pending-request.component.ts
```diff
@@ -89,6 +91,8 @@ export class PendingRequestComponent implements OnInit, OnChanges {
         (reqs: PendingRequest[]) => {
           this.requests = reqs;
           this.status = reqs.length ? 'loaded' : 'empty';
+          if (this.status === 'loaded')
+            this.collapsed = false;
```
smadbe, line 95: As this status has been set at the previous line, it is clearer (especially in a subscribe) to base it on the "source", so just put `if (reqs.length)`.
```suggestion
          if (reqs.length) this.collapsed = false;
```

src/app/modules/group/components/pending-request/pending-request.component.html
```diff
@@ -47,6 +47,16 @@
                 <p-sortIcon [field]="col.field"></p-sortIcon>
               </div>
             </th>
+            <th
```
smadbe, line 50: if `col.field === 'group.name'` you do the same as `at`? So just put it with "at".
But actually, the 3 columns are the same except the `width: 13.3333rem; `, so just use one code and use a condition for this width.

smadbe (changes requested):

src/app/modules/group/components/pending-request/pending-request.component.ts
AyeRlS: It's still works when I move groupColums to the class definition, can I keep !== ?

src/app/modules/group/components/pending-request/pending-request.component.ts
AyeRlS: fixed

src/app/modules/group/components/pending-request/pending-request.component.html
AyeRlS: Done

src/app/modules/group/components/pending-request/pending-request.component.ts
smadbe: Yes, that's even better as here, it is really the same reference as it is a constant

src/app/modules/group/components/pending-request/pending-request.component.html
smadbe: much better

smadbe (approved):

PR #103, opened 2020-09-30
Add missing columns of group manager list 

### LINKED ISSUE #23: Group administration page: v1 - readonly
In the group "administration" tabs: (based on `/design` implementation)
- rename the tab to "Managers"
- title: "Managers of this group"
- columns:
  * name (useless to display the column caption (as it is now)) 
  * level: which can be "read-only", "memberships", or "full", + the blue "bubble" level    (from "none", "memberships" and "memberships_and_group" received from backend)
  * "can grant group access": green "check-circle" if true, red "times-circle" if false
  * "can watch members": green "check-circle" if true, red "times-circle" if false
- content obtained from [groupManagerView service](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupManagersView) 
- don't handle selection / select all for now
- ~allow sort by name~  actually, not possible as there are no column header on name

GabrielVidal1:
Fixes #23 
Add the 3 missing columns of the manager list.
The 'Can Manage' column have an additional widget sometimes (the progress-level) but I'm not sure if this is really necessary.
The other 2 columns are represented by an open or closed lock

### DIFF
--- a/src/app/modules/group/components/group-manager-list/group-manager-list.component.html
+++ b/src/app/modules/group/components/group-manager-list/group-manager-list.component.html
@@ -41,8 +41,9 @@
               </div>
             </div>
           </th>
+          <th style="width: 110px;"></th>
       </tr>
-    </ng-template>
+      </ng-template>
 
       <ng-template
         #bodyTemplate
@@ -81,6 +82,7 @@
               <i *ngIf="rowData[rowData.can_watch_members]" class="fa fa-lock-open"></i>
             </span>
           </td>
+          <td style="width: 110px;"></td>
         </tr>
       </ng-template>
     </alg-grid>
--- a/src/app/modules/group/components/group-manager-list/group-manager-list.component.html
+++ b/src/app/modules/group/components/group-manager-list/group-manager-list.component.html
@@ -11,37 +11,78 @@
     [scrollable]="true"
     class="slanted-grid"
     >
-
-    <ng-template #headerTemplate let-columns>
-      <tr style="border: none;">
-        <th style="min-width: 110px;">
-          <div class="slanted-header">
-            <div class="slanted-header-content">
-              Name
+      <ng-template #headerTemplate let-columns>
+        <tr style="border: none;">
+          <th style="width: 160px;">
+            <div class="slanted-header">
+              <div class="slanted-header-content">
+                Name
+              </div>
             </div>
-          </div>
-        </th>
-        <th style="width: 160px; transform: none;">
+          </th>
+          <th style="min-width: 110px;">
+            <div class="slanted-header">
+              <div class="slanted-header-content">
+                Can Manage
+              </div>
+            </div>
+          </th>
+          <th style="width: 5rem;">
+            <div class="slanted-header">
+              <div class="slanted-header-content">
+                Can grant group access
+              </div>
+            </div>
+          </th>
+          <th style="width: 5rem;">
+            <div class="slanted-header">
+              <div class="slanted-header-content">
+                Can watch members
+              </div>
+            </div>
+          </th>
       </tr>
     </ng-template>
 
-    <ng-template
-      #bodyTemplate
-      let-rowData
-      let-columns="columns"
-      let-rowIndex="rowIndex"
-    >
-      <tr [pSelectableRow]="rowData" [pSelectableRowIndex]="rowIndex">
-        <td
-          style="text-align: left;width: 160px;padding-left: 0.5rem;"
-          tooltipPosition="top"
-          tooltipStyleClass="tooltip-custom"
-        >
-          {{ rowData.name }}
-        </td>
-      </tr>
-    </ng-template>
-  </alg-grid>
-
+      <ng-template
+        #bodyTemplate
+        let-rowData
+        let-columns="columns"
+        let-rowIndex="rowIndex"
+      >
+        <tr [pSelectableRow]="rowData" [pSelectableRowIndex]="rowIndex">
+          <td
+            style="text-align: left;width: 160px;padding-left: 0.5rem;"
+            tooltipPosition="top"
+            tooltipStyleClass="tooltip-custom"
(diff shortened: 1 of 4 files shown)


src/app/modules/group/components/group-manager-list/group-manager-list.component.html
```diff
@@ -11,37 +11,78 @@
...
+              <div class="slanted-header-content">
+                Can grant group access
+              </div>
+            </div>
+          </th>
+          <th style="width: 5rem;">
+            <div class="slanted-header">
+              <div class="slanted-header-content">
+                Can watch members
+              </div>
```
smadbe, line 41: The problem is that these labels go out of the screen and so are not visible. Maybe we could do it in the other direction? Could you try or suggest an idea?

src/app/modules/group/components/group-manager-list/group-manager-list.component.html
```diff
@@ -11,37 +11,78 @@
...
+            style="text-align: left;width: 160px;padding-left: 0.5rem;"
+            tooltipPosition="top"
+            tooltipStyleClass="tooltip-custom"
+          >
+            {{ rowData.name }}
+          </td>
+          <td style="min-width: 110px;">
+            {{ rowData.can_manage }}
+          </td>
+          <td style="width: 5rem;">
```
smadbe, line 64: Why are they clickable?

smadbe (changes requested):
Try to polish the elements as they would be pushed in prod right now. Don't hesitate to ask if you have questions.

src/app/modules/group/components/group-manager-list/group-manager-list.component.html
GabrielVidal1: fixed, they are no longer clickable

src/app/modules/group/components/group-manager-list/group-manager-list.component.html
GabrielVidal1: In the design, this is made possible by the last, column which is the 'Edit' column, that adds enough space for all the headers to be visible. For now I just added an empty column at the end (that will later be replaced by the 'edit' column)

smadbe: Ok thx. We will still have to fix the column repartition (80% is taken by the second column), but that's something else.

smadbe (approved):

GabrielVidal1:
I rebased this branch on master to have the progress-level component and also the latest linter changes.
Also changed the icons showed (now a check and a cross instead of locks)
The text for the 'Can Manage' is now more user friendly (instead of just the value of the corresponding field from the backend)

smadbe:
👍 you can merge

src/app/modules/group/components/group-manager-list/group-manager-list.component.ts
```diff
@@ -22,15 +22,28 @@ export class GroupManagerListComponent implements OnChanges {
...
+  private convertText(can_manage: string): string {
+    switch(can_manage) {
+      case 'none':
+        return 'Read-only';
+      case 'memberships':
+        return 'Memberships';
+      case 'memberships_and_group':
+        return 'Memberships and group';
+      default:
+        return '';
```
GabrielVidal1, line 34: Just one last thing, I don't exactly know what to return if the `can_manage` value is not one of the one of the desired value.

smadbe: Hum indeed, this is not really correct. You cannot really consider that `can_manage` is a string as what is returned by the service is a union type (you "downcast" it, you lose information). You should probably define an exported type in the service for this union type, use this type here, and you will not have to handle any default case.

smadbe (changes requested):

src/app/modules/group/http-services/get-group-managers.service.ts
```diff
@@ -7,7 +7,7 @@ export interface Manager {
   id: string,
   name: string,
 
-  can_manage: string,
+  can_manage: 'none' | 'memberships' | 'memberships_and_group',
```
GabrielVidal1, line 10: I now restrict the type of `can_manage` to only possible values instead of just a `string`

src/app/modules/group/components/group-manager-list/group-manager-list.component.html
```diff
@@ -11,37 +11,80 @@
...
+        <tr [pSelectableRow]="rowData" [pSelectableRowIndex]="rowIndex">
+          <td
+            style="text-align: left;width: 160px;padding-left: 0.5rem;"
+            tooltipPosition="top"
+            tooltipStyleClass="tooltip-custom"
+          >
+            {{ rowData.name }}
+          </td>
+          <td style="min-width: 110px;">
+            {{ getManagerLevel(rowData) }}
```
GabrielVidal1, line 63: The value of the `can_manage` field is now "translated" here to be more user-friendly

smadbe: You cannot do it this way, otherwise this function will be called hundreds times. No function calls can be done in the rendering. 
One way to solve that is after `.getGroupManagers(this.group.id)`  to map each managers to a dictionary where you have added a "can_manage_as_text" for instance. And so having `managers: Manager[] = [];` -> `managers: (Manager&{can_manage_as_text:string})[] = [];` for instance.
You can check what I did with functions such as "withCodeAdditions" but here it is just something internal to this class, it can probably be done more easily.

smadbe (changes requested):

src/app/modules/group/components/group-manager-list/group-manager-list.component.html
GabrielVidal1: Ok I fixed it (I didn't know how to do this properly).

smadbe (approved):

PR #109, opened 2020-10-01
Handle links in breadcrumbs and show group panel if we are on a group page

smadbe:
Handle links in breadcrumbs and show group panel if we are on a group page. 
Review commit per commit as the first commit is preparation for the next ones.

### DIFF
--- a/src/app/core/components/breadcrumb/breadcrumb.component.ts
+++ b/src/app/core/components/breadcrumb/breadcrumb.component.ts
@@ -1,5 +1,5 @@
 import { Component, Input } from '@angular/core';
-import { PageInfo } from 'src/app/shared/services/current-content.service';
+import { ContentBreadcrumb } from 'src/app/shared/services/current-content.service';
 
 @Component({
   selector: 'alg-breadcrumb',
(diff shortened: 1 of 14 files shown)


src/app/core/components/breadcrumb/breadcrumb.component.ts
```diff
@@ -8,9 +8,9 @@ import { PageInfo } from 'src/app/shared/services/current-content.service';
 })
 export class BreadcrumbComponent {
 
-  @Input() pageInfo: PageInfo|null
+  @Input() contentBreadcrumb?: ContentBreadcrumb
```
AyeRlS, line 12: why did you put the `contentBreadcrumb` as optional instead of letting `|null` ?

src/app/modules/item/pages/item-details/item-details.component.spec.ts
```diff
@@ -18,8 +17,7 @@ describe('ItemDetailsComponent', () => {
       set: {
         providers: [
           { provide: CurrentContentService, useValue: {
-            setCurrent: (_i: NavItem) => {},
-            setPageInfo: (_p: any) => {},
+            setCurrent: (_i: any) => {},
```
AyeRlS, line 20: should `_i` be of type `ContentInfo` ?

smadbe: For tests, we do not care (especially as we do not do anything with it).

src/app/core/components/breadcrumb/breadcrumb.component.ts
smadbe: Because the breadcrumb attribute of ContentInfo is optional and not null.

AyeRlS (approved):

PR #113, opened 2020-10-02
Add item header

### LINKED ISSUE #106: Item details: header
Start the item page with (according to designs)
- the title
- the subtitle (if any)
- a tab "content" with a "description" section if type === 'Chapter'

GabrielVidal1:
Fixes #106 
Add item header component.

### DIFF
--- a/src/app/modules/item/components/item-header/item-header.component.html
+++ b/src/app/modules/item/components/item-header/item-header.component.html
@@ -1,0 +1,11 @@
+<div class="task-header-container">
+  <div class="task-header">
+    <div class="task-user-image">
+      <img src="assets/images/_messi.jpg" style="width: 4rem; height: 4rem;" />
+    </div>
+    <div class="task-title">
+      <p>{{item.string.title}}</p>
+      <span>Subtitle for the task</span>
+    </div>
+  </div>
+</div>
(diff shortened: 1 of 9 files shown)


src/app/modules/item/http-services/get-item-by-id.service.ts
```diff
@@ -8,6 +8,7 @@ export interface Item {
   requires_explicit_entry: boolean
   string: {
     title: string|null,
+    subtitle: string|null,
```
smadbe, line 11: Subtitle is both nullable and omitable (cfr API, "nullable" and not required), so it is `subtitle?: string|null`

smadbe (changes requested):

src/app/modules/item/components/item-header/item-header.component.html
```diff
@@ -0,0 +1,11 @@
+<div class="task-header-container">
+  <div class="task-header">
+    <div class="task-user-image">
+      <img src="assets/images/_messi.jpg" style="width: 4rem; height: 4rem;" />
```
smadbe, line 4: ah and don't put an image of messi by default ;-)

GabrielVidal1: fixed, I removed the image all together

src/app/modules/item/http-services/get-item-by-id.service.ts
GabrielVidal1: fixed

smadbe (approved):

PR #117, opened 2020-10-06
Accept join request of different group/subgroup in the same time

### LINKED ISSUE #87: Pending request: add this group only / all subgroups switch, step 3: adapt acceptation/rejection
As it is done in the design branch (http://dev.algorea.org/branch/design/), add to the "pending request" box the switch "group only / all subgroups".

Step3:
Create a new pseudo service with accept or reject several requests even if they are related to different groups. Handle with rxJS the call to the multiple services and aggregate the results at the end.

For the user it should be transparent that it is one or several services.

AyeRlS:
Fixes #87

The way the request are handled has been changed in order to be able to accept/request join request for a group and its subgroups


src/app/modules/group/components/pending-request/pending-request.component.ts
```diff
@@ -143,20 +145,31 @@ export class PendingRequestComponent implements OnInit, OnChanges {
...
 
+  processRequests(action: Action) {
+    const requestMap = new Map<string, string[]>();
+    this.selection.forEach(elm => {
+      const groupID = elm.group.id;
+      const memberID = elm.user.group_id;
+
+      const value = requestMap.get(groupID);
+      if (value) requestMap.set(groupID, value.concat([memberID]));
+      else       requestMap.set(groupID, [memberID]);
```
smadbe, line 156: We don't space indent in the middle of the line to align in this project.
```suggestion
      else requestMap.set(groupID, [memberID]);
```

src/app/modules/group/components/pending-request/pending-request.component.ts
```diff
@@ -143,20 +145,31 @@ export class PendingRequestComponent implements OnInit, OnChanges {
...
+    });
+    return forkJoin(
+      Array.from(requestMap.entries()).map(elm => {
+        if (action === Action.Accept) return this.requestActionService.acceptJoinRequest(elm[0], elm[1]);
+        else                          return this.requestActionService.rejectJoinRequest(elm[0], elm[1]);
```
smadbe, line 161:
```suggestion
        else return this.requestActionService.rejectJoinRequest(elm[0], elm[1]);
```

smadbe (approved):
Good just fix these two details before merging.

PR #118, opened 2020-10-06
Add item-get-children service

GabrielVidal1:
Add the service used to display the list of children in the chapter content page.

### DIFF
--- a/src/app/modules/item/http-services/get-item-children.service.ts
+++ b/src/app/modules/item/http-services/get-item-children.service.ts
@@ -4,6 +4,13 @@ import { Observable } from 'rxjs';
 import { environment } from 'src/environments/environment';
 
 export interface ItemChild {
+  string: {
+    title: string,
+  },
+  permissions: {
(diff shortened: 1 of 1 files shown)


src/app/modules/item/http-services/get-item-children.service.ts
```diff
@@ -0,0 +1,28 @@
+import { HttpClient } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { environment } from 'src/environments/environment';
+
+export interface ItemChild {
+  string: {
+    title: string,
```
smadbe, line 8: The title is nullable. Always read carefully the doc to set the type.

smadbe (changes requested):

src/app/modules/item/http-services/get-item-children.service.ts
GabrielVidal1: fixed

smadbe (approved):

PR #122, opened 2020-10-07
Add item content tab

### LINKED ISSUE #107: Item details: children (part 1)
# Objective
Display children as shown [here](https://france-ioi.github.io/algorea-designs/03.Activities_01.Content_a.Content-presentation-(full-text)).

# Details
* Do a separate component which is only visible for chapters (item.type 'Chapter') and with can_view >= content
* Add the "To validate this chapter, solve at least all tasks with Validation type " with the same style as designs
* Use the usual loading/error/ready state for fetching children data (https://france-ioi.github.io/algorea-devdoc/api/#operation/itemChildrenView)
* just display the "ready"  string when the data is loaded, we'll do the display in another issue

GabrielVidal1:
Fixes #107 
Add the item content tab and the loading of the item children if the item is a chapter.
For now only 'Ready' is displayed when the children of the chapter are successfully loaded.

### DIFF
--- a/src/app/modules/item/components/chapter-detail/chapter-detail.component.ts
+++ b/src/app/modules/item/components/chapter-detail/chapter-detail.component.ts
@@ -1,0 +1,42 @@
+import { Component } from '@angular/core';
+import { EMPTY } from 'rxjs';
+import { switchMap } from 'rxjs/operators';
+import { GetItemChildrenService, ItemChild } from '../../http-services/get-item-children.service';
+import { ItemDataSource } from '../../services/item-datasource.service';
+
+@Component({
+  selector: 'alg-chapter-detail',
+  templateUrl: './chapter-detail.component.html',
+  styleUrls: ['./chapter-detail.component.scss']
+})
+export class ChapterDetailComponent {
+
+  item$ = this.itemDataSource.item$;
+
+  state: 'loading' | 'ready' | 'error' = 'loading';
+
+  constructor(
+    private itemDataSource: ItemDataSource,
+    private getItemChildrenService: GetItemChildrenService,
+  ) {
+    this.reloadData();
+  }
+
+  children: ItemChild[] = [];
+
+  private reloadData() {
+    this.state = 'loading';
+    this.itemDataSource.itemData$.pipe(
+      switchMap(itemData => {
+        if (!itemData.attemptId) return EMPTY;
+        return this.getItemChildrenService.get(itemData.item.id, itemData.attemptId);
+      }),
+    ).subscribe((children: ItemChild[]) => {
+      this.children = children;
+      this.state = 'ready';
+    },
+      _err => {
+        this.state = 'error';
+      });
+  }
+}
--- a/src/app/modules/item/item-routing.module.ts
+++ b/src/app/modules/item/item-routing.module.ts
@@ -3,6 +3,7 @@ import { RouterModule } from '@angular/router';
 import { ItemByIdComponent } from './pages/item-by-id/item-by-id.component';
 import { ItemDetailsComponent } from './pages/item-details/item-details.component';
 import { ItemEditComponent } from './pages/item-edit/item-edit.component';
+import { ItemContentComponent } from './pages/item-content/item-content.component';
 
 @NgModule({
   imports: [RouterModule.forChild([
@@ -13,6 +14,12 @@ import { ItemEditComponent } from './pages/item-edit/item-edit.component';
         {
           path: 'details',
           component: ItemDetailsComponent,
+          children: [
+            {
+              path: '',
+              component: ItemContentComponent,
+            }
+          ]
         },
         {
           path: 'edit',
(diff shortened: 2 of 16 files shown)


src/app/modules/item/components/chapter-detail/chapter-detail.component.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ItemDataSource } from '../../services/item-datasource.service';
+
+@Component({
+  selector: 'alg-chapter-detail',
+  templateUrl: './chapter-detail.component.html',
+  styleUrls: ['./chapter-detail.component.scss']
+})
+export class ChapterDetailComponent {
+
+  item$ = this.itemDataSource.item$;
```
smadbe, line 14: The item data source is actually dedicated to communication between the item-by-id page and its "route children" (in router-outlet) as it is the only way to pass information between these. For classical components, it is preferable to use the classical Angular communication between a component and its subcomponents: inputs.

Then, there are actually two possible way to pass the input:
- either you pass the observable (of state or item): it means the subcomponent needs to know what to do when its parent is in loading/error to hide itself (here in practice, nothing as its parents do not display it when in error/loading, but the child does not know that) and what to do with this stream of items where some items are not suitable for this component.
- either you pass the item itself in case it has to fetch its children.

I think the second option is cleaner (in terms of software design) for subcomponents of item-content. In practice, it means that at init and onchange, the previous service subscription should be be cancelled if any, the state should be to set to loading and then call the service. (with a true reloadData probably)

src/app/modules/item/components/chapter-detail/chapter-detail.component.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  state: 'loading' | 'ready' | 'error' = 'loading';
```
smadbe, line 16: keep all the instance variable together, `children` should be here as well.

src/app/modules/item/components/chapter-detail/chapter-detail.component.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.reloadData();
+  }
+
+  children: ItemChild[] = [];
+
+  private reloadData() {
+    this.state = 'loading';
+    this.itemDataSource.itemData$.pipe(
+      switchMap(itemData => {
+        if (!itemData.attemptId) return EMPTY;
```
smadbe, line 31: Have you tested it? I think it means your component will stay in "loading" forever in this case (which may happen as I explained).

src/app/modules/item/components/chapter-detail/chapter-detail.component.ts
```diff
@@ -0,0 +1,42 @@
...
+        return this.getItemChildrenService.get(itemData.item.id, itemData.attemptId);
+      }),
+    ).subscribe((children: ItemChild[]) => {
```
smadbe, line 34: The type should not be required. And `children` should be aligned on "_err".

src/app/modules/item/item-routing.module.ts
```diff
@@ -13,6 +14,12 @@ import { ItemEditComponent } from './pages/item-edit/item-edit.component';
         {
           path: 'details',
           component: ItemDetailsComponent,
+          children: [
+            {
+              path: '',
+              component: ItemContentComponent,
+            }
+          ]
```
smadbe, line 22: As I changed the router, probably you have handle to conflict properly. But actually you just have to do the same with the new "detail" page which is one level deeper.

src/app/modules/item/pages/item-content/item-content.component.html
```diff
@@ -0,0 +1,4 @@
+<ng-container *ngIf="item$ | async as item">
+  <alg-chapter-detail *ngIf="item.type === 'Chapter'">
```
smadbe, line 2: We discussed yesterday other cases where the this component should not be displayed (or behave correctly in this case at least).

src/app/modules/item/pages/item-details/item-details.component.html
```diff
@@ -8,7 +8,18 @@
...
-    <p>item-details works!</p>
+    <nav mat-tab-nav-bar>
+      <a mat-tab-link
+        [routerLink]="'./'"
+        routerLinkActive #rla1="routerLinkActive"
+        [routerLinkActiveOptions]="{exact: true}"
+        [active]="rla1.isActive">
+        Content
+      </a>
+    </nav>
```
smadbe, line 19: It does not work correctly as the tab is not selected by default.

smadbe (changes requested):
I have raised several important design choices to be changed. I think in this case, what would be easier would be to start a new branch from master (the updated one) and to re-do everything from scratch, copying what is good to keep and redo what needs to be rethought. 
Good work still 👍

smadbe (changes requested):
Also, I would not call the component "ChapterDetailComponent" but "ChapterChildrenComponent" as chapter detail may be compared with item detail while it is not related at all.

smadbe:
Redone in #127

PR #137, opened 2020-10-16
Replace auto-starting result by fetching results and start one if none 

smadbe:
Replace in the item state the attempt_id by a detailed version of it a list of results and the current result, which allows to know the current score (and more)
So, instead of starting a result after fetching the item: fetch the list of results, find the more appropriate result and if none start a new one.

### DIFF
--- a/src/app/modules/item/services/item-datasource.service.ts
+++ b/src/app/modules/item/services/item-datasource.service.ts
@@ -1,14 +1,16 @@
 import { Injectable, OnDestroy } from '@angular/core';
 import { BehaviorSubject, concat, EMPTY, forkJoin, Observable, of, Subject } from 'rxjs';
 import { catchError, filter, map, mapTo, switchMap, tap } from 'rxjs/operators';
+import { bestAttemptFromResults, implicitResultStart } from 'src/app/shared/helpers/attempts';
 import { errorState, FetchError, Fetching, fetchingState, isReady, Ready, readyState } from 'src/app/shared/helpers/state';
 import { ResultActionsService } from 'src/app/shared/http-services/result-actions.service';
 import { NavItem } from 'src/app/shared/services/nav-types';
 import { canCurrentUserViewItemContent } from '../helpers/item-permissions';
 import { BreadcrumbItem, GetBreadcrumbService } from '../http-services/get-breadcrumb.service';
 import { GetItemByIdService, Item } from '../http-services/get-item-by-id.service';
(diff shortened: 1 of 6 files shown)


src/app/modules/item/services/item-datasource.service.ts
```diff
@@ -68,21 +70,54 @@ export class ItemDataSource implements OnDestroy {
...
+    return this.getResultsService.getResults(nav.itemId, attempt).pipe(
+      switchMap(results => {
+        // 1) if attempt_id was given as arg, try to select the matching result
+        if (nav.attemptId) {
+          const currentResult = results.find(r => r.attemptId === nav.attemptId);
+          if (currentResult) return of({ results: results, currentResult: currentResult});
+        }
+        // 2) if there are already results on this item, select the best one
+        const currentResult = bestAttemptFromResults(results);
+        if (currentResult !== null) return of({ results: results, currentResult: currentResult});
```
GabrielVidal1, line 103:
```suggestion
        if (currentResult) return of({ results: results, currentResult: currentResult});
```

src/app/modules/item/services/item-datasource.service.ts
```diff
@@ -68,21 +70,54 @@ export class ItemDataSource implements OnDestroy {
...
+        // 4) otherwise, start a result
+        const attemptId  = nav.attemptId || nav.parentAttemptId;
+        if (!attemptId) return EMPTY; // unexpected
+        return this.resultActionsService.start(nav.itemPath.concat([nav.itemId]), attemptId).pipe(
+          // once a result has been created, fetch it
+          switchMap(() => this.getResultsService.getResults(nav.itemId, attempt).pipe(
+            map(results => {
+              // this time we are sure to have a started result as we just started it
+              const currentResult = bestAttemptFromResults(results);
+              if (currentResult === null) throw new Error('Unexpected: result just created not found');
```
GabrielVidal1, line 115:
```suggestion
              if (!currentResult) throw new Error('Unexpected: result just created not found');
```

GabrielVidal1 (changes requested):

src/app/modules/item/services/item-datasource.service.ts
smadbe: For the moment, "everywhere" in the code, nulls are checked if `=== null` and `!== null`. I should recheck if it is really appropriate to do this, but at least for this PR it is consistent with the rest of the code.

AyeRlS (approved):

PR #140, opened 2020-10-22
Item header score

### LINKED ISSUE #136: Display score in item header
Display the score (as a score ring) on the left of the title in the item header as shown on https://france-ioi.github.io/algorea-designs/03.Activities_00.Header_b2.Access-to-activity.

Only display it if the user has a result on this item (use current score and validated from this result).

GabrielVidal1:
Fixes #136 
Added the score ring to the item header.
The score is round down to the nearest integer to avoid a really long floating point number.

### DIFF
--- a/src/app/modules/item/components/item-header/item-header.component.ts
+++ b/src/app/modules/item/components/item-header/item-header.component.ts
@@ -1,5 +1,4 @@
 import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
-import { bestAttemptFromResults } from 'src/app/shared/helpers/attempts';
 import { ItemData } from '../../services/item-datasource.service';
 
 @Component({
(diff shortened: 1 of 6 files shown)


src/app/modules/item/components/item-header/item-header.component.ts
```diff
@@ -1,11 +1,19 @@
...
-export class ItemHeaderComponent {
-  @Input() item: Item;
+export class ItemHeaderComponent implements OnChanges {
+  @Input() itemData: ItemData;
+
+  current_score = 0;
+
+  ngOnChanges(_changes: SimpleChanges): void {
+    if (this.itemData.results && this.itemData.currentResult) {
+      this.current_score = Math.floor(this.itemData.currentResult.score);
```
smadbe, line 16: Probably it would be more useful to do this rounding in the score ring component directly.
And cleaner/easier way to do it is to just use a pipe : https://angular.io/api/common/DecimalPipe

smadbe (changes requested):

src/app/modules/item/components/item-header/item-header.component.ts
GabrielVidal1: fixed

smadbe: Good but you can remove this code then.

smadbe (changes requested):

smadbe (approved):
👍

PR #144, opened 2020-10-23
Item edition service

AyeRlS:
Clicking on save will now save the changes and update the displayed values, but not on the left menu.

Also the toast aren't working currently.


src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+  styleUrls: ['./item-edit.component.scss'],
+  providers: [MessageService]
 })
 export class ItemEditComponent implements OnDestroy {
+  item: Item;
   itemForm = this.formBuilder.group({
     // eslint-disable-next-line @typescript-eslint/unbound-method
     title: ['', [Validators.required, Validators.minLength(3),]],
   });
   itemLoadingState$ = this.itemDataSource.state$;
```
smadbe, line 27: Why "loading state" actually?  State includes if it is loading but it is not a loading state.

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
   templateUrl: './item-edit.component.html',
-  styleUrls: ['./item-edit.component.scss']
+  styleUrls: ['./item-edit.component.scss'],
+  providers: [MessageService]
 })
 export class ItemEditComponent implements OnDestroy {
+  item: Item;
   itemForm = this.formBuilder.group({
     // eslint-disable-next-line @typescript-eslint/unbound-method
     title: ['', [Validators.required, Validators.minLength(3),]],
```
smadbe, line 25: I don't see these validators applied (you should have tested that yourself by hand!)

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+    this.subscriptions.forEach(subscription => subscription.unsubscribe());
   }
 
   getCurrentItem(): void {
-    this.itemSubscription = this.itemLoadingState$.pipe(
+    this.subscriptions.push(this.itemLoadingState$.pipe(
       filter<Ready<ItemData> | Fetching | FetchError, Ready<ItemData>>(isReady),
     ).subscribe(state => {
-      const item = state.data.item;
+      this.item = state.data.item;
```
smadbe, line 52: I would avoid storing an item in addition to the observable.
You should probably just store the id... but even probably it should be in the form.

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
       this.itemForm.patchValue({
-        title: item.string.title
+        title: this.item.string.title
       });
+    }));
+  }
+
+  onSubmit(): void {
```
smadbe, line 59: I would expect an `onSubmit` fct to be the function called when pushing a "submit" button...

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+    this.onSubmit();
   }
 
   ngOnDestroy() {
     this.currentContent.editState.next('non-editable');
-    this.itemSubscription.unsubscribe();
+    this.subscriptions.forEach(subscription => subscription.unsubscribe());
   }
 
   getCurrentItem(): void {
```
smadbe, line 48: Wrong naming. This function does not get current item!

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+  successToast() : void {
+    this.messageService.add({
+      severity: 'success',
+      summary: 'Success',
+      detail: 'Changes successfully saved.',
+      life: TOAST_LENGTH,
+    });
+  }
+
+  saveInput(): void {
```
smadbe, line 83: "save input", what input ?  ;-)

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+    this.updateItemStringService.updateItem(
+      this.item.id,
+      this.itemForm.value
```
smadbe, line 86: It might work at first (except if you add the id as I suggest above) but probably you'll need to extract properly the value (I think, that might be done later maybe).

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+    ).subscribe(
+      _status => {
+        this.successToast();
+        this.itemDataSource.refreshItem();
+        this.currentContent.editAction.next(EditAction.Cancel);
+      },
+      _err => this.errorToast()
+    );
```
smadbe, line 94: and while it is processing? UI should be frozen.

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+
+  saveInput(): void {
+    this.updateItemStringService.updateItem(
+      this.item.id,
+      this.itemForm.value
+    ).subscribe(
+      _status => {
+        this.successToast();
+        this.itemDataSource.refreshItem();
+        this.currentContent.editAction.next(EditAction.Cancel);
```
smadbe, line 91: look normal to use to simulate a call to "cancel" to leave the page? 🤨

smadbe (changes requested):

src/app/modules/item/http-services/update-item-string.service.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  updateItem(
+    itemId: string,
+    changes: object,
+    languageTag?: string
+  ) {
+    // Fixme: Get languageTag properly
+    const tag = languageTag || 'default';
+    return this.http.put<ActionResponse<Object>>(
```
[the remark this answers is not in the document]
AyeRlS: Fixed

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+    }));
+  }
+
+  onSubmit(): void {
+    this.subscriptions.push(this.currentContent.editAction$
+      .pipe(filter(action => action === EditAction.Save))
+      .subscribe(_action => this.saveInput()));
+  }
+
+  errorToast() :void {
```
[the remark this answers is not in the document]
AyeRlS: The ': void' have been removed

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: I've used the same name you used in item detail

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: It's fixed

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: This function as been removed

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: moved the creation of observer into the constructor

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: moved the creation of observer into the constructor

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: If all the inputs are named correctly it should work as itemForm.value return an object that match what http.put needs. But I could be missing something.

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: The idea is to save the new values => refresh the item => simulate a call to cancel to leave the page. To make things more clear I go rename it to "close" or I would need to add another EditAction that would basically have the same behavior than "cancel" but with a different name.

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: I've disabled the form on processing

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -1,49 +1,97 @@
...
+import { MessageService } from 'primeng/api';
+import { ERROR_MESSAGE } from '../../../../shared/constants/api';
 
 
 @Component({
   selector: 'alg-item-edit',
   templateUrl: './item-edit.component.html',
-  styleUrls: ['./item-edit.component.scss']
+  styleUrls: ['./item-edit.component.scss'],
+  providers: [MessageService]
```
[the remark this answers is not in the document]
AyeRlS: They are working now but in the case of success, the page is changed too fast to actually see it. This could be an easy fix if I have to add a new EditAction.

smadbe (changes requested):
Few remarks. Some other feedbacks do not seem to be fixed (? not sure). I'll recheck at next review.

src/app/modules/item/pages/item-by-id/item-by-id.component.ts
```diff
@@ -80,7 +80,7 @@ export class ItemByIdComponent implements OnDestroy {
         if (isItemInfo(currentInfo)) {
           void this.router.navigate(itemDetailsRoute(currentInfo.data, action === EditAction.StartEditing));
         }
-      })
+      }),
     );
```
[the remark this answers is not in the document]
smadbe: Not answered

src/app/modules/item/pages/item-edit/item-edit.component.ts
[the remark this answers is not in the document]
smadbe: Not fixed?

src/app/modules/item/pages/item-edit/item-edit.component.ts
smadbe: That's a great name then! 😁

src/app/modules/item/pages/item-edit/item-edit.component.ts
smadbe: Wouldn't it make sense to store it in the form?

src/app/modules/item/pages/item-edit/item-edit.component.ts
smadbe: Yes, that should be ok for now.

src/app/modules/item/pages/item-edit/item-edit.component.ts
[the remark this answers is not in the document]
AyeRlS: I haven't been able to make it work when the component is changed, but if an error happens, the notification will corectly show up

src/app/modules/item/pages/item-by-id/item-by-id.component.ts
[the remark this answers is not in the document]
AyeRlS: Those must be a mix of remnants of tests I did and auto formatting, I can remove them if you want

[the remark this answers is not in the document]
smadbe: Remove the commit if you know how to do.  You should not commit such things.

src/app/shared/services/current-content.service.ts
```diff
@@ -32,7 +32,7 @@ export function isGroupInfo(info: ContentInfo|null): info is GroupInfo {
 }
 
 export type EditState = 'non-editable'|'editable'|'editing'|'editing-noaction'; // editingNoAction is for temporary disabled actions
-export enum EditAction { StartEditing, Save, Cancel }
+export enum EditAction { StartEditing, Save, FinishEditing }
```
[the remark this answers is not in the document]
AyeRlS: Changed

src/app/modules/item/pages/item-edit/item-edit.component.html
```diff
@@ -6,9 +6,12 @@
 
   <div *ngIf="state.tag === 'ready'" style="margin: 1rem 0 0.5rem 0">
     <form [formGroup]="itemForm">
-      <alg-input [parentForm]="itemForm" name="title" size="large" [isDark]='false'></alg-input>
+      <alg-input [parentForm]="itemForm" name="title" size="large" [isDark]='false'>
+        <ng-template #errorTemplate>
+          Title should be at least 3 characters long
```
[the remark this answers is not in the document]
AyeRlS: My idea was to allow custom error message, but I've changed the way errors are displayed, and no need to write the same thing 2 times now.

src/app/modules/shared-components/components/input/input.component.html
```diff
@@ -12,3 +12,7 @@
     <i class="fa fa-{{buttonIcon}}"></i>
   </span>
 </div>
+
+<ng-container *ngIf="parentForm.get(name) as input">
```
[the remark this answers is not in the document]
AyeRlS: Fixed

src/app/modules/shared-components/components/input/input.component.ts
```diff
@@ -15,18 +17,20 @@ import { FormGroup } from '@angular/forms';
   styleUrls: ['./input.component.scss'],
 })
 export class InputComponent implements OnInit, OnChanges {
-  @Input() name : string | number | null; // name of the input in the parent form
-  @Input() parentForm : FormGroup;
+  @Input() name: string; // name of the input in the parent form
+  @Input() parentForm: FormGroup;
```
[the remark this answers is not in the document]
AyeRlS: Considering there is conflits with master, and probably other new rules will be broken, So I'm planning to fix everything in one go.

src/app/modules/item/pages/item-edit/item-edit.component.ts
smadbe: That's not a great name but I'll check this later.

src/app/modules/item/pages/item-edit/item-edit.component.ts
smadbe: Still not answered

src/app/modules/item/pages/item-edit/item-edit.component.ts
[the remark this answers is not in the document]
smadbe: But so it does not work well as it is not displayed correctly on success. Do you want me to check?

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -75,7 +75,10 @@ export class ItemEditComponent implements OnDestroy {
   }
 
   saveInput() {
-    if (this.itemForm.invalid) return;
+    if (this.itemForm.invalid) {
+      this.errorToast('The form isn\'t valid');
```
smadbe, line 79: "You need to solve all the errors displayed in the form to save changes."

smadbe (changes requested):
It is time to merge master as well. 
Directly after this PR, you should also make the left menu refresh the name when it has been edited.

src/app/modules/item/pages/item-edit/item-edit.component.html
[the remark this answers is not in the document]
smadbe: much better now 👍

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: I'm confused because I was sure to already have answered this. My bad

So I did some tests and putting the id in the form would require a lot of changes. So, instead of storing the item, I choose to only store the id of the item

src/app/modules/item/pages/item-edit/item-edit.component.ts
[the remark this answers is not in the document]
AyeRlS: Yes please

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -14,11 +14,11 @@ import { ERROR_MESSAGE } from '../../../../shared/constants/api';
...
   selector: 'alg-item-edit',
   templateUrl: './item-edit.component.html',
-  styleUrls: ['./item-edit.component.scss'],
-  providers: [MessageService]
+  styleUrls: [ './item-edit.component.scss' ],
+  providers: [ MessageService ]
 })
 export class ItemEditComponent implements OnDestroy {
-  itemId : string;
+  itemId : string | null = null;
```
smadbe, line 21: Use undefined instead if you have no reason to use null. 
So just `itemId?: string`

src/app/modules/shared-components/components/input/input.component.html
```diff
@@ -3,7 +3,7 @@
     [type]="inputType"
     [placeholder]="placeholder"
     [ngClass]="{'has-button': buttonIcon}"
-    [formControlName]="name"
+    [formControlName]="name!"
```
smadbe, line 6: I told yesterday I forbid you to use exclamation mark except for setting variable to ngprime component. 
Here input is not a ng-prime component, so it could cause problem.
If it must be a string  (initially you put ` string | number | null`? ), just put an empty string as default value.

src/app/core/components/item-nav/item-nav.component.ts
```diff
@@ -115,13 +115,16 @@ export class ItemNavComponent implements OnInit, OnDestroy {
...
+        switchMap(content => {
+          if (content === null) return EMPTY;
+          const item = content.data;
           if (item === null || !item.result) return EMPTY; // only update if result (score) is given
           if (!isReady(this.state) || this.state.data.selectedElement?.itemId !== item.nav.itemId) return EMPTY; // and same item id
           return of(this.state.data.withUpdatedElement(item.nav.itemId, {
             attemptId: item.result.attemptId,
-            score: { best: item.result.bestScore, current: item.result.currentScore, validated: item.result.validated }
+            score: { best: item.result.bestScore, current: item.result.currentScore, validated: item.result.validated },
+            title: content.title
```
smadbe, line 127: I would have some remarks on this code, but I changed this as well in my current work, so that's not important (either remove the commit or let it there).

smadbe (changes requested):

src/app/modules/shared-components/components/input/input.component.html
AyeRlS: `parentForm.get` doesn't allow number | null, so I've removed it as well to avoid issues.
I guess I could add the support for numbers but for the moment I thought this was enough to only support strings.
I've put "" as the default value.

src/app/modules/item/pages/item-edit/item-edit.component.ts

src/app/core/components/item-nav/item-nav.component.ts
AyeRlS: It has been removed, but it seems like out already took care of the refreshing of the item value in the menu

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: Done

src/app/modules/item/http-services/update-item-string.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  providedIn: 'root'
+})
+export class UpdateItemStringService {
+
+  constructor(private http: HttpClient) {
+  }
+
+  updateItem(
+    itemId: string,
+    changes: object,
```
GabrielVidal1, line 18: why not have 
`{
description: string|null,
image_url: string|null,
subtitle: string|null,
title: string|null
}` instead of `object` here ?

GabrielVidal1 (changes requested):

src/app/modules/item/http-services/update-item-string.service.ts
smadbe: Yes, true. "object" type should not be used in Typescript. (we could use {string -> unknown} dictionary at least.)
The problem is that we should not repeat all attributes in each class and here it does not really help preventing  future problems as its output data. 
But at least for item strings where the supported attribute are not many, we should add them.

Note that all these attributes should be optional in addition to be nullable.

src/app/modules/shared-components/components/input/input.component.html
smadbe: But at the end, it has not been changed!

smadbe (changes requested):
2 comments above still need to be fixed (object and '!')

src/app/modules/shared-components/components/input/input.component.html
AyeRlS: This is finally fixed now !

smadbe (changes requested):
Good. Just let me check the MessageService.

GabrielVidal1 (approved):

smadbe (approved):
Ok I see how to fix MessageService but as it is something more global, it better fits out of this PR, so you can merge and I'll do it afterwards.

Issue #124, opened 2020-10-08
Item progress tab: log view (first version)


smadbe:
# Overview 

In the progress page (for instance http://dev.algorea.org/en/#/activities/by-id/4702;parentAttempId=0;path=/details/progress), display a list (table) of latest activities of the current user on this item. (as bit as drafted at the bottom of [this design](https://france-ioi.github.io/algorea-designs/02.Skills_02.Progression) , but with more relevant content)

# Details
* In ItemLogViewComponent (existing component):
   -  add a table to display the result of `state.data` (check MemberListComponent for an example of table)
   - this table should have the following columns:
      * "Action", which values which could be (see ([API for the service](https://france-ioi.github.io/algorea-devdoc/api/#operation/itemActivityLogForItem) for explanation of the input) "Submission" (if "submission" with score undefined), "Submission (score: <score>)" (if "submission" with score given), "Activity started" (if "result_started") or "Activity validated" (if "result_validated")
      * "Content", which is the "title" of the item with a link to that item
      * Time with minute precision  (check MemberListComponent for an example)
* May require to update a bit the `ActivityLog` interface returned by the service. ([API for the service](https://france-ioi.github.io/algorea-devdoc/api/#operation/itemActivityLogForItem) used)
* No need to support any sort for now.

smadbe:
Note that item-log-view component should probably be included in item-current-situation component by routing.

smadbe:
It would be nice to create a PR when you have the structure of the components so that other devs can start on the tabs in parallel.

So the PRs should be: 
1) the pages structure with a simple message
2) the page which may be in error, loading or ready state, on ready state (data loaded), just display "data loaded"
3) the data is displayed as expected

mblockelet:
Hi, sorry again for the delay, I should be working on it either tonight if I feel well enough, or tomorrow. Sure for the three pull requests.

smadbe:
The service has been completely reworked to match the most recent needs, see https://france-ioi.github.io/algorea-devdoc/api/#operation/itemActivityLog. 
As I also mentioned somewhere else, I have (before the new service) re-written most of your component to match changes related with routing and the clean things that I missed during the review.
The third and last PR on this issue should then be to adapt the service and to display the expected data in the table.

smadbe:
Description rewritten.

PR #148, opened 2020-10-24
First version of item-progress with subcomponents

mblockelet:
First pull request for #124 : adds a simple pages structure with item-progress, item-current-situation and item-log-view.

### DIFF
--- a/src/app/modules/item/pages/item-current-situation/item-current-situation.component.html
+++ b/src/app/modules/item/pages/item-current-situation/item-current-situation.component.html
@@ -1,3 +1,4 @@
-<h1>User situation</h1>
-<alg-selection (change)="onViewChanged($event)" [items]="viewItems" [selected]="viewSelected"></alg-selection>
-<router-outlet></router-outlet>
+<alg-section icon="fa fa-list" label="User situation">
+  <div class="selection-center"><alg-selection (change)="onViewChanged($event)" [items]="viewItems" [selected]="viewSelected"></alg-selection></div>
+  <router-outlet></router-outlet>
+</alg-section>
--- a/src/app/modules/item/pages/item-current-situation/item-current-situation.component.html
+++ b/src/app/modules/item/pages/item-current-situation/item-current-situation.component.html
@@ -1,3 +1,3 @@
 <h1>User situation</h1>
-<alg-selection (onChange)="onViewChange($event)" [items]="viewItems" [selected]="viewSelected"></alg-selection>
+<alg-selection (onChange)="onViewChanged($event)" [items]="viewItems" [selected]="viewSelected"></alg-selection>
 <router-outlet></router-outlet>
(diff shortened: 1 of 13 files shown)


src/app/modules/item/pages/item-current-situation/item-current-situation.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  selector: 'alg-item-current-situation',
+  templateUrl: './item-current-situation.component.html',
+  styleUrls: ['./item-current-situation.component.scss']
+})
+export class ItemCurrentSituationComponent implements OnInit {
+
+  viewItems = [
+    {label: 'Chapter view', route: './chapter'},
+    {label: 'Log view', route: './'}
+    ];
```
smadbe, line 13: Styling for consistency with the rest of the code:
```suggestion
  viewItems = [
    { label: 'Chapter view', route: './chapter' },
    { label: 'Log view', route: './' }
  ];
```

src/app/modules/item/pages/item-current-situation/item-current-situation.component.html
```diff
@@ -0,0 +1,3 @@
+<h1>User situation</h1>
+<alg-selection (change)="onViewChanged($event)" [items]="viewItems" [selected]="viewSelected"></alg-selection>
```
smadbe, line 2: Arf, the items parameter is still a any type :-/  (from the design)
This is the kind of things I try to fix directly when I see it. I'll fix this alg-selection when we'll be done with this PR...

src/app/modules/item/pages/item-current-situation/item-current-situation.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  viewSelected = 1;
+
+  constructor() {}
+
+  onViewChanged(selectedIdx: number) {
+    // TODO :: Implement routing from alg-selection
+    this.viewSelected = selectedIdx;
+  }
+
+  ngOnInit(): void {
```
smadbe, line 24: For the moment, we do not put return type for functions returning nothing.

src/app/modules/item/pages/item-progress/item-progress.component.html
```diff
@@ -0,0 +1,2 @@
+<h1>Progress</h1>
+<router-outlet></router-outlet>
```
smadbe, line 2: Try to have a correct style directly. Normally, that should not be too difficult as it has been done for group pages. (and the design branch)

smadbe (changes requested):
Ok. Just fix the design so that we have a good basis, should not be complicated.
I would recommend you to run the linter locally before pushing. Also, using a IDE such as VSCode (with Angular and Eslint extensions) helps a lot (good support for nav between references and linting while typing).

smadbe (approved):

PR #149, opened 2020-10-26
Adding Yourself page

AyeRlS:
Adding a simple Yourself page and routing to it.


src/app/core/app-routing.module.ts
```diff
@@ -10,6 +10,10 @@ const routes: Routes = [
     path: 'items',
     loadChildren: () => import('../modules/item/item.module').then(m => m.ItemModule)
   },
+  {
+    path: 'yourself',
+    loadChildren: () => import('../modules/yourself/yourself.module').then(m => m.YourselfModule)
+  }
```
smadbe, line 16: Users are groups and multiple things in yourself will be group related stuffs. So it should be in the group module and probably under the url `groups/me`.

src/app/core/components/group-nav/group-nav.component.ts
```diff
@@ -20,10 +21,12 @@ export class GroupNavComponent {
 
   joinedGroups: GroupData = [];
   managedGroups: GroupData = [];
+  selectedGroup: number;
```
smadbe, line 24: "0" for yourself while I expect a selected group (so a group id) is misleading. If you just need it for yourself for now, use a boolean.

src/app/core/components/group-nav/group-nav.component.html
```diff
@@ -5,12 +5,12 @@
...
   </div> -->
-  <!-- <div class="group-self" (click)="onSelectYourself($event, 0)">
+  <div class="group-self" (click)="onSelectYourself()">
     <span class="anonym-group">
       <img src="assets/images/Male.svg" style="width: 3.3333rem; height: 3.3333rem;" />
     </span>
-    <span class="group-subtitle" [ngClass]="{'selected': selectedGroup == 0 ? true : false}">Yourself</span>
-  </div> -->
+    <span class="group-subtitle" [ngClass]="{'selected': selectedGroup == 0}">Yourself</span>
+  </div>
```
smadbe, line 13: The click and the link selection should be handle 100% in html which now if you current page matches the link.

src/app/modules/yourself/pages/yourself/yourself.component.ts
```diff
@@ -0,0 +1,28 @@
...
+import { CurrentContentService } from 'src/app/shared/services/current-content.service';
+
+const YourselfBreadcrumbCat = 'Yourself';
+
+@Component({
+  selector: 'alg-yourself',
+  templateUrl: './yourself.component.html',
+  styleUrls: ['./yourself.component.scss'],
+})
+export class YourselfComponent implements OnDestroy {
```
smadbe, line 11: Call it "CurrentUserComponent". "Yourself" is just the current name of the link.

smadbe (commented):
I am afraid you will have to redo it for scratch 🙄😕 You should ask when you are not 100% how to do it.

src/app/modules/yourself/pages/yourself/yourself.component.ts
AyeRlS: Done

src/app/core/components/group-nav/group-nav.component.ts
AyeRlS: This variable is used to highlight which group tab has been selected, the variable name makes in the file (the tabs are called group-self, group-joined, and group-managed) but I can see why it can be misleading.

How about renaming it to selectedTab ? Also, this variable isn't currently used in the other element, but according to the design it should be, should I add it in another PR or in this one ?

src/app/core/app-routing.module.ts
AyeRlS: I've moved it to the group module and under that url.

src/app/core/components/group-nav/group-nav.component.html
AyeRlS: Done

smadbe (approved):
Looks good now 👍 .
It is unfortunate that 95% of the code of the PR  is garbage (code which has been changed/removed afterwards), which make review more complicated (so possibly missing real error), git history messy (last change on modules will be a removal of something which have been added just before). 
You should have followed the recommendation of really redoing it from scratch... it would have been both cleaner and faster for you. Also, it seems you commit too early without having tested a minimum what you commit :-/.

smadbe (changes requested):
BTW, you still have to merge breaking changes in master.

smadbe (approved):

PR #150, opened 2020-10-26
Update primeng

smadbe:
CSS fixing after primeng update.
(will require everybody to update package `npm i`)


smadbe:
Let's forget this PR for now.

PR #151, opened 2020-10-27
enable strictPropertyInitialization and fix code where it must be

smadbe:
enable strictPropertyInitialization and fix code where it must be. 
I used forced unwrapping (`varname!`) for argument of primeng component are their type is non-optional while they actually support undefined values.

### DIFF
--- a/src/app/modules/group/components/group-join-by-code/group-join-by-code.component.ts
+++ b/src/app/modules/group/components/group-join-by-code/group-join-by-code.component.ts
@@ -19,10 +19,10 @@ import { CodeActionsService } from '../../http-services/code-actions.service';
 
 export class GroupJoinByCodeComponent implements OnChanges {
 
-  @Input() group: Group
+  @Input() group?: Group
   @Output() refreshRequired = new EventEmitter<void>();
 
-  groupExt: Group & CodeAdditions; // group extended with code related attributes
+  groupExt?: Group & CodeAdditions; // group extended with code related attributes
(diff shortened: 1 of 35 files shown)


src/app/modules/group/components/group-join-by-code/group-join-by-code.component.ts
```diff
@@ -32,7 +32,7 @@ export class GroupJoinByCodeComponent implements OnChanges {
   ) { }
 
   ngOnChanges() {
-    this.groupExt = withCodeAdditions(this.group);
+    this.groupExt = this.group ? withCodeAdditions(this.group) : undefined;
```
AyeRlS, line 35: This can be changed to `this.groupExt = this.group && withCodeAdditions(this.group)` for clarity

AyeRlS (approved):

GabrielVidal1 (approved):

src/app/modules/group/components/group-join-by-code/group-join-by-code.component.ts
smadbe: This is more elegant but I would not say it brings clarity. This is very cryptic. If you show ` this.group && withCodeAdditions(this.group)`, 95% of developers will wrongly guess what it returns exactly. In most languages, it will result in an error or a boolean. 
At least it is not consistent with the rest of the code as I never use this JS/TS construct. I have to think if is a good idea to use these.

PR #155, opened 2020-10-27
Add lock icon when item can't be started in left menu

### LINKED ISSUE #145: Flag item which cannot be started in left menu
## Motivations
Be able to quickly identify which item is "locked" and cannot be started now.
Example: https://france-ioi.github.io/algorea-designs/00.Global_06.Navigation-panel

## Subtasks
* indicate in grey and with a lock icon the items with permissions can_view=info

GabrielVidal1:
Fixes #145 
I wasn't sure if I was supposed to handle the click on a locked item, so when a user clicks on a locked item, it doesn't do anything.

### DIFF
--- a/src/app/core/components/item-nav-tree/item-nav-tree.component.html
+++ b/src/app/core/components/item-nav-tree/item-nav-tree.component.html
@@ -44,7 +44,7 @@
         </span>
         <span (click)="selectNode(node)" class="node-label">
           <span class="label-container">
-            <span class="node-label-title">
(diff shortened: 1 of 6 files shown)


smadbe (changes requested):
👍 
Could you try with lighter gray?
Yes this is better not to let the user select the item (even it can show some info on the right), but here it select it in white but does nothing, leaving two selections in the menu. So try to find a better solution: either never highlight other item than the current item in the menu, or not allow selecting these specific ones.

smadbe (approved):
👍

PR #159, opened 2020-10-28
Fix right panel broken when clicking an item in chapter children list

GabrielVidal1:
The right panel is no longer broken but the left menu is when clicking on an item in chapter children list.
The left menu can be fixed by reloading the page though


smadbe (approved):
Ok, looks ok. I assume you have tested it carefully.

PR #168, opened 2020-10-30
Implement groupRecentActivity service

mblockelet:
Second pull request for #124 : implements the RecentActivityService which calls groupRecentActivity, and load the data from item-log-view.


smadbe (changes requested):
Good. Only minor remarks.
Wasn't the chapter view supposed to be a different component (different route) from log view? Maybe it missed during the previous review that looks like a mistake.
Also more related with prev PR,  would it be possible to have the spacing between the icon and progress in the progress tab?

PR #174, opened 2020-11-02
Group edition page

### LINKED ISSUE #164: Group edition, step 1: empty edit page
## Motivations
Be able to edit groups (similarly to items)

## Subtasks
* on group page, when user is allowed to manage "group and memberships". Display the edit button icon in the header. When clicking on it, move to edit page. 
* on the empty edit page, clicking on "cancel" returns on the group page

GabrielVidal1:
Fixes #164 
Add an empty edit page for groups.

### DIFF
--- a/src/app/modules/group/pages/group-by-id/group-by-id.component.ts
+++ b/src/app/modules/group/pages/group-by-id/group-by-id.component.ts
@@ -1,0 +1,72 @@
+import { Component, OnDestroy } from '@angular/core';
+import { ActivatedRoute, Router } from '@angular/router';
+import { Subscription } from 'rxjs';
+import { filter, map } from 'rxjs/operators';
+import { FetchError, Fetching, isReady, Ready } from 'src/app/shared/helpers/state';
+import { CurrentContentService, EditAction, GroupInfo, isGroupInfo } from 'src/app/shared/services/current-content.service';
+import { Group } from '../../http-services/get-group-by-id.service';
+import { GroupDataSource } from '../../services/group-datasource.service';
+
+const GroupBreadcrumbCat = 'Groups';
+
+/**
+ * GroupByIdComponent is just a container for detail or edit page but manages the fetching on id change and (un)setting the current content.
+ */
+@Component({
+  selector: 'alg-group-by-id',
+  templateUrl: './group-by-id.component.html',
+  styleUrls: [ './group-by-id.component.scss' ],
+  providers: [ GroupDataSource ],
+})
+export class GroupByIdComponent implements OnDestroy {
+
+  private subscriptions: Subscription[] = []; // subscriptions to be freed up on destroy
+
+  constructor(
+    private router: Router,
+    private activatedRoute: ActivatedRoute,
+    private currentContent: CurrentContentService,
+    private groupDataSource: GroupDataSource
+  ) {
+
+    // on route change: refetch item if needed
+    this.activatedRoute.paramMap.subscribe(params => {
+      const id = params.get('id');
+      this.currentContent.current.next({
+        id: id, type: 'group', breadcrumbs: { category: GroupBreadcrumbCat, path: [], currentPageIdx: -1 }
+      } as GroupInfo);
+      if (id) this.groupDataSource.fetchGroup(id);
+    });
+
+    // on state change, update current content page info (for breadcrumb)
+    this.subscriptions.push(
+      this.groupDataSource.state$.pipe(
+      filter<Ready<Group>|Fetching|FetchError,Ready<Group>>(isReady),
+      map(state => ({
+        type: 'group',
+        id: state.data.id,
+        breadcrumbs: {
+          category: GroupBreadcrumbCat,
+          path: [{ title: state.data.name, navigateTo: [ 'groups', 'details', state.data.id ] }],
+          currentPageIdx: 0,
+        },
+        title: state.data.name,
+      }))
+    ).subscribe(p => this.currentContent.current.next(p)),
+
+    this.currentContent.editAction$.pipe(
+      filter(action => [ EditAction.StartEditing, EditAction.StopEditing ].includes(action))
+    ).subscribe(action => {
+      const currentInfo = this.currentContent.current.value;
+      if (isGroupInfo(currentInfo)) {
+        void this.router.navigate([ 'groups', 'by-id', currentInfo.id, action === EditAction.StartEditing ? 'edit' : 'details' ]);
+      }
+    })
+    );
+  }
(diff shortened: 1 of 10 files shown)


src/app/modules/group/pages/group-by-id/group-by-id.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    // on state change, update current content page info (for breadcrumb)
+    this.subscriptions.push(
+      this.groupDataSource.state$.pipe(
+      filter<Ready<Group>|Fetching|FetchError,Ready<Group>>(isReady),
+      map(state => ({
+        type: 'group',
+        id: state.data.id,
+        breadcrumbs: {
+          category: GroupBreadcrumbCat,
+          path: [{ title: state.data.name, navigateTo: [ 'groups', 'details', state.data.id ] }],
```
AyeRlS, line 50:
```suggestion
          path: [{ title: state.data.name, navigateTo: [ 'groups', 'by-id', state.data.id, 'details' ] }],
```

src/app/shared/services/current-content.service.ts
```diff
@@ -33,7 +33,12 @@ export interface ItemDetails {
   validated?: boolean,
 }
 
-export interface GroupInfo extends ContentInfo { type: 'group' }
+type GroupId = string;
+
+export interface GroupInfo extends ContentInfo {
+  id: GroupId,
+  type: 'group'
+}
```
AyeRlS, line 41: Why did you specify the type of GroupId separately instead of doind something like `id: string` ?

GabrielVidal1: To be consistent with the id of the `ItemInfo` interface (which is of the type `ItemId`, which is a `string)`

src/app/modules/group/pages/group-by-id/group-by-id.component.ts
GabrielVidal1: fixed

src/app/shared/services/current-content.service.ts
smadbe: Actually, I do that quite often. 
The reasons are that in type theory, a groupId and an attemptId are completely different types, they are not compatible between each other, there is no place you can use one at the place of the other one, the same as between a number and an array of strings. The "string" here is just the container for the actual type (as an array of bytes is a common container for number and the array of strings). So without pushing the concept to hard typing every variable, I use it for a main usage: when the variable name is "id"or used as "id" in its usage.  This way we do not use it by mistake in an arg which is a attemptId, a itemId, ...
It is not generalized to everywhere in the app for the moment to prevent importing tons of ids everywhere and it is only useful at some places, but it could be.

src/app/modules/group/pages/group-by-id/group-by-id.component.ts
```diff
@@ -0,0 +1,72 @@
...
+  private subscriptions: Subscription[] = []; // subscriptions to be freed up on destroy
+
+  constructor(
+    private router: Router,
+    private activatedRoute: ActivatedRoute,
+    private currentContent: CurrentContentService,
+    private groupDataSource: GroupDataSource
+  ) {
+
+    // on route change: refetch item if needed
```
smadbe, line 32:
```suggestion
    // on route change: refetch group if needed
```

src/app/modules/group/pages/group-by-id/group-by-id.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    ).subscribe(p => this.currentContent.current.next(p)),
+
+    this.currentContent.editAction$.pipe(
+      filter(action => [ EditAction.StartEditing, EditAction.StopEditing ].includes(action))
+    ).subscribe(action => {
+      const currentInfo = this.currentContent.current.value;
+      if (isGroupInfo(currentInfo)) {
+        void this.router.navigate([ 'groups', 'by-id', currentInfo.id, action === EditAction.StartEditing ? 'edit' : 'details' ]);
+      }
+    })
```
smadbe, line 64: looks like there is an indentation issue

src/app/modules/group/pages/group-edit/group-edit.component.html
```diff
@@ -0,0 +1,9 @@
+<ng-container *ngIf="state$ | async as state">
+
+  <p-progressSpinner *ngIf="state.tag === 'fetching'"></p-progressSpinner>
+
+  <p *ngIf="state.tag === 'error'">Error while loading the group</p>
+
+  <p>Group loaded successfully!</p>
```
smadbe, line 7: you do not to check the state to print successfully

src/app/modules/group/pages/group-overview/group-overview.component.ts
```diff
@@ -4,7 +4,8 @@ import { GroupDataSource } from '../../services/group-datasource.service';
 @Component({
   selector: 'alg-group-overview',
   templateUrl: './group-overview.component.html',
-  styleUrls: [ './group-overview.component.scss' ]
+  styleUrls: [ './group-overview.component.scss' ],
+  providers: [ GroupDataSource ],
```
smadbe, line 8: That will create a new instance of the datasource for this component while the goal is the share it.

smadbe (changes requested):
👍 Few details to be changed

src/app/shared/services/current-content.service.ts
AyeRlS: Okay, thanks !

src/app/modules/group/pages/group-overview/group-overview.component.ts
GabrielVidal1: fixed, I kept the provider in this component and removed the one in the `GroupOverviewComponent`

src/app/modules/group/pages/group-edit/group-edit.component.html
GabrielVidal1: fixed

src/app/modules/group/pages/group-by-id/group-by-id.component.ts
GabrielVidal1: fixed

src/app/modules/group/pages/group-by-id/group-by-id.component.ts
GabrielVidal1: fixed

smadbe (approved):
👍

smadbe:
Resolve conflicts AND wait the CI to pass before merging.

PR #182, opened 2020-11-04
Textarea reactive

AyeRlS:
Changing the text area component to use it within a reactive form

### DIFF
--- a/src/app/modules/group/components/group-invite-users/group-invite-users.component.html
+++ b/src/app/modules/group/components/group-invite-users/group-invite-users.component.html
@@ -1,7 +1,7 @@
 <alg-section-paragraph icon="fa fa-user-plus" label="Invite users" [collapsible]="true">
   <span class="content-title">Add users name list, separated by commas.</span>
   <div class="invite-users">
-    <alg-textarea [parentForm]="inviteForm" [inputName]="inputName" icon="show" placeholder="login_1, login_2..." (textChange)="onTextChange($event)">
+    <alg-textarea [parentForm]="inviteForm" [inputName]="inputName" icon="show" placeholder="login_1, login_2..." (textChange)="onTextChange($event)" [disabled]="state === 'loading'">
     </alg-textarea>
 
     <alg-message *ngIf="state === 'too_many'" type="error" detail="You cannot invite more than 100 users at once"
--- a/src/app/modules/group/components/group-invite-users/group-invite-users.component.html
+++ b/src/app/modules/group/components/group-invite-users/group-invite-users.component.html
@@ -1,7 +1,7 @@
 <alg-section-paragraph icon="fa fa-user-plus" label="Invite users" [collapsible]="true">
   <span class="content-title">Add users name list, separated by commas.</span>
   <div class="invite-users">
-    <alg-textarea icon="show" placeholder="login_1, login_2..." (textChange)="onTextChange($event)" [disabled]="state === 'loading'">
+    <alg-textarea [parentForm]="inviteForm" inputName="invitations" icon="show" placeholder="login_1, login_2..." (textChange)="onTextChange($event)">
     </alg-textarea>
 
     <alg-message *ngIf="state === 'too_many'" type="error" detail="You cannot invite more than 100 users at once"
--- a/src/app/modules/group/components/group-invite-users/group-invite-users.component.html
+++ b/src/app/modules/group/components/group-invite-users/group-invite-users.component.html
@@ -1,7 +1,7 @@
 <alg-section-paragraph icon="fa fa-user-plus" label="Invite users" [collapsible]="true">
   <span class="content-title">Add users name list, separated by commas.</span>
   <div class="invite-users">
-    <alg-textarea [parentForm]="inviteForm" inputName="invitations" icon="show" placeholder="login_1, login_2..." (textChange)="onTextChange($event)">
+    <alg-textarea [parentForm]="inviteForm" [inputName]="inputName" icon="show" placeholder="login_1, login_2..." (textChange)="onTextChange($event)">
     </alg-textarea>
 
     <alg-message *ngIf="state === 'too_many'" type="error" detail="You cannot invite more than 100 users at once"
--- a/src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
+++ b/src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
@@ -5,6 +5,7 @@ import { ERROR_MESSAGE } from 'src/app/shared/constants/api';
 import { TOAST_LENGTH } from 'src/app/shared/constants/global';
 import { CreateGroupInvitationsService, InvitationResult } from '../../http-services/create-group-invitations.service';
 import { Group } from '../../http-services/get-group-by-id.service';
+import { FormBuilder } from '@angular/forms';
 
 interface Message
 {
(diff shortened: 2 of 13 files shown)


src/app/modules/group/components/group-invite-users/group-invite-users.component.html
```diff
@@ -1,7 +1,7 @@
 <alg-section-paragraph icon="fa fa-user-plus" label="Invite users" [collapsible]="true">
   <span class="content-title">Add users name list, separated by commas.</span>
   <div class="invite-users">
-    <alg-textarea icon="show" placeholder="login_1, login_2..." (textChange)="onTextChange($event)" [disabled]="state === 'loading'">
+    <alg-textarea [parentForm]="inviteForm" inputName="invitations" icon="show" placeholder="login_1, login_2..." (textChange)="onTextChange($event)">
```
smadbe, line 4: why has the "disabled" disappeared?

src/app/modules/shared-components/components/input-error/input-error.component.html
```diff
@@ -0,0 +1,5 @@
+<div *ngIf="control && control.errors" style="color: #a94442">
```
smadbe, line 1: Good idea to make a component for that. Why "input" if you want to reuse it not only for 'input' but also in other types of form elements??

smadbe: The style thing should be in the CSS file.

smadbe, line 1: `control && control.errors` can be `control?.errors`

src/app/modules/shared-components/components/input-error/input-error.component.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  control?: AbstractControl;
+
+  constructor() {
+  }
+
+  ngOnChanges(): void {
+    const elm = this.form?.get(this.inputName);
+    if (!elm) return;
+    this.control = elm;
```
smadbe, line 21: Ok so you can just write that I think.
```suggestion
    this.control = this.form?.get(this.inputName);
```

src/app/modules/shared-components/components/textarea/textarea.component.ts
```diff
@@ -18,12 +21,16 @@ export class TextareaComponent implements OnInit {
 
   ngOnInit(): void {}
 
+  getValue(): string {
+    return this.parentForm?.get(this.inputName)?.value as string;
+  }
+
   setValue(text: string): void {
```
smadbe, line 28: getValue/setValue and change detection should be requested directly to the parentForm I assume.

smadbe (changes requested):

src/app/modules/group/components/group-invite-users/group-invite-users.component.html
smadbe: Ok, maybe handled below. I'll see when the other changes will be fixed.

AyeRlS: > It looks like you're using the disabled attribute with a reactive form directive. If you set disabled to true when you set up this control in your component class, the disabled attribute will actually be set in the DOM for you. We recommend using this approach to avoid 'changed after checked' errors.
>
> Example: 
> form = new FormGroup({
   >  first: new FormControl({value: 'Nancy', disabled: true}, Validators.required),
   >  last: new FormControl('Drew', Validators.required)
>  });

The way disabled was handled worked but threw this warning in the console when used with a reactive form. So I implemented the change the warning suggested:

`inviteForm = this.formBuilder.group({ invitations: { value: '', disabled: this.state == 'loading' } });`

src/app/modules/shared-components/components/input-error/input-error.component.html
AyeRlS: This component can be used with a component that allow input, but I can see why it could be confusing, what do you think would be a better name ?
Also, the css has been moved.

src/app/modules/shared-components/components/input-error/input-error.component.ts
AyeRlS: I've done some changes to make the fact that elm can be null clearer

src/app/modules/shared-components/components/input-error/input-error.component.html
AyeRlS: This doesn't seem do be working

src/app/modules/group/components/group-invite-users/group-invite-users.component.html
smadbe: `this.state == 'loading'` is evaluated at object initialization, it is not a function.

smadbe: So probably it is a good idea, but the way it is done will not work IMO

src/app/modules/shared-components/components/textarea/textarea.component.ts
AyeRlS: They have been removed

src/app/modules/group/components/group-invite-users/group-invite-users.component.html
AyeRlS: I've changed the way it's done !

GabrielVidal1 (approved):

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -23,15 +23,16 @@ export class GroupInviteUsersComponent {
   @Input() group?: Group
   @Output() refreshRequired = new EventEmitter<void>();
 
-  @ViewChild(TextareaComponent) private textArea?: TextareaComponent;
-
   state: 'empty'|'too_many'|'loading'|'ready' = 'empty';
+  inviteForm = this.formBuilder.group({ invitations: '' });
+  inputName = 'invitations'
```
smadbe, line 28: Mark it as a constant
```suggestion
  readonly inputName = 'invitations'
```

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -95,13 +96,13 @@ export class GroupInviteUsersComponent {
   }
 
   onInviteClicked(): void {
-    if (!this.textArea || !this.group) return;
+    if (!this.group) return;
```
smadbe, line 99: Aren't you supposed to check that the form is valid somehow?

src/app/modules/item/pages/item-edit-content/item-edit-content.component.html
```diff
@@ -1,3 +1,3 @@
 <alg-section icon="fa fa-book" label="Description">
-  <alg-textarea [value]="value" (textChange)="updateForm($event)"></alg-textarea>
+  <alg-textarea [parentForm]="parentForm" inputName="description"></alg-textarea>
```
smadbe, line 2: There are still change to item-edit, you didn't really start a PR from master 😔
How am I supposed to review that?

smadbe (changes requested):

src/app/modules/shared-components/components/input-error/input-error.component.html
smadbe: FormError ?

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
smadbe: And where is the validation BTW? Isn't it supposed to be "moved" to reactive form?

PR #190, opened 2020-11-06
Parent skills

GabrielVidal1:
Add the parent skill list in the item content tab when the current item is a skill.

### DIFF
--- a/src/app/modules/item/components/parent-skills/parent-skills.component.html
+++ b/src/app/modules/item/components/parent-skills/parent-skills.component.html
@@ -1,0 +1,33 @@
+<p-progressSpinner *ngIf="state === 'loading'"></p-progressSpinner>
+
+<p *ngIf="state === 'error'">Error while loading the skill parents</p>
+
+<ng-container *ngIf="state === 'ready'">
+  <alg-section
+  label="Parent skills"
+  icon="fa fa-link"
+  >
+    <ul class="data-list-container sub-skill">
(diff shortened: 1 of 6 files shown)


src/app/modules/item/components/parent-skills/parent-skills.component.html
```diff
@@ -0,0 +1,33 @@
+<p-progressSpinner *ngIf="state === 'loading'"></p-progressSpinner>
+
+<p *ngIf="state === 'error'">Error while loading the skill parents</p>
+
+<ng-container *ngIf="state === 'ready'">
+  <alg-section
```
smadbe, line 6: You could have put the ngIf on the alg-section directly.

src/app/modules/item/http-services/get-item-parents.service.ts
```diff
@@ -0,0 +1,76 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class GetItemParentsService {
+
+  constructor(private http: HttpClient) { }
+
+  get(id: string, attemptId: string): Observable<ItemParent[]> {
+    // eslint-disable-next-line no-console
+    console.log('call');
```
smadbe, line 52: !!!
... but why the hell have you put this `eslint-disable`?  It is there especially so that you do not forget to remove the console.log at the end.

smadbe (approved):
👍 
Just fix the console log before merging.

PR #191, opened 2020-11-06
Textarea

AyeRlS:
Changing the textarea component to be used with reactive forms


src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -115,7 +122,7 @@ export class GroupInviteUsersComponent {
         this.displayResponse(res);
 
         // Clear the textarea
-        if (this.textArea) this.textArea.setValue('');
+        this.inviteForm.get(this.inputName)?.patchValue('');
```
smadbe, line 125: Use the var discussed above.

`setValue` looks safer than patchValue when the goal is not to update the whole form (as we do in edit). You can also decide to emit the event (in both fct) but probably here it is not what we want.

src/app/modules/shared-components/components/textarea/textarea.component.ts
```diff
@@ -1,29 +1,27 @@
...
 
-  setValue(text: string): void {
-    this.value = text;
-    this.onChange();
-  }
-
-  onChange(): void {
-    this.textChange.emit(this.value);
+  ngOnChanges(): void {
+    if (this.disabled) this.parentForm?.disable();
```
smadbe, line 24: I don't think the "disabled" property should be kept is the way to do it is to used the formgroup. At the end, it means that is the parent which ask his child to disable itself, which itself change a state of the parent...

smadbe (changes requested):
Good 👍   Just check these quick remarks.
I haven't checked the tests yet, I'll give you my feedback in a few minutes.

smadbe (changes requested):
For the test, it is just before `this.inviteForm.get(this.inputName)` has no `get` as `this.formBuilder.group({ invitations: '' });` and you set that the group function return an empty object.
Anyway, there is nothing in this test, just remove the file.

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -18,22 +19,41 @@ interface Message
...
   messages: Message[] = [];
+  subscription?: Subscription;
 
   constructor(
     private createGroupInvitationsService: CreateGroupInvitationsService,
     private messageService: MessageService,
-  ) {}
+    private formBuilder: FormBuilder,
+  ) {
+    this.subscription = this.inviteForm.get(this.inputName)?.valueChanges.subscribe((change: string) => this.checkInput(change));
```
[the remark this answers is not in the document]
AyeRlS: Done

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -18,22 +19,41 @@ interface Message
...
 })
-export class GroupInviteUsersComponent {
+export class GroupInviteUsersComponent implements OnDestroy {
 
   @Input() group?: Group
   @Output() refreshRequired = new EventEmitter<void>();
 
-  @ViewChild(TextareaComponent) private textArea?: TextareaComponent;
-
+  readonly inputName = 'invitations'
```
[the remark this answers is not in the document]
AyeRlS: Done

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -18,22 +19,41 @@ interface Message
...
+    private formBuilder: FormBuilder,
+  ) {
+    this.subscription = this.inviteForm.get(this.inputName)?.valueChanges.subscribe((change: string) => this.checkInput(change));
+  }
+
+  ngOnDestroy(): void {
+    this.subscription?.unsubscribe();
+  }
+
+  checkInput(input: string): void {
```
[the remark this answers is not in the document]
AyeRlS: Changed

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -18,22 +19,41 @@ interface Message
...
+    this.state = 'ready';
```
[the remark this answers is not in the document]
AyeRlS: Fixed

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -81,27 +101,14 @@ export class GroupInviteUsersComponent {
...
   onInviteClicked(): void {
-    if (!this.textArea || !this.group) return;
+    if (!this.group) return;
 
     // clear the messages
     this.messages = [];
 
     // remove empty logins and duplicates
-    const logins = this.textArea.value.split(',')
+    const logins = (this.inviteForm.get(this.inputName)?.value as string).split(',')
```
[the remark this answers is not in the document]
AyeRlS: Fixed

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -81,27 +101,14 @@ export class GroupInviteUsersComponent {
...
-    if (logins.length === 0) {
-      this.state = 'empty';
-    } else if (logins.length >= 100) {
-      this.state = 'too_many';
-    }
-  }
-
   onInviteClicked(): void {
-    if (!this.textArea || !this.group) return;
+    if (!this.group) return;
```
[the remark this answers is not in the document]
AyeRlS: Done

Should this kind of failsafe be implemented every times? The button calling this function is only enabled when state == ready, so a regular user isn't supposed to find itself calling this function if the state isn't ready.

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
AyeRlS: Isn't patchValue what we want here ? https://www.tektutorialshub.com/angular/setvalue-patchvalue-in-angular/#setvalue-vs-patchvalue.

src/app/modules/shared-components/components/textarea/textarea.component.ts
AyeRlS: The way 'disabled' is done is different now.

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -39,17 +39,19 @@ export class GroupInviteUsersComponent implements OnInit, OnDestroy {
...
   }
 
   ngOnDestroy(): void {
     this.subscription?.unsubscribe();
   }
 
-  checkInput(input: string): void {
+  loginListChanged(newValue: string): void {
+    if (this.state === 'loading')
+      return;
```
smadbe, line 60: When no brackets -> on one line

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -107,13 +107,17 @@ export class GroupInviteUsersComponent implements OnInit, OnDestroy {
 
   /* events */
   onInviteClicked(): void {
-    if (!this.group) return;
+    if (!this.group || this.state != 'ready') return;
```
smadbe, line 119: `!==` for string comparison

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -113,11 +113,11 @@ export class GroupInviteUsersComponent implements OnInit, OnDestroy {
     this.messages = [];
 
     // remove empty logins and duplicates
-    const input = this.inviteForm.get(this.inputName)?.value as string;
-    if (!input)
+    const control = this.inviteForm.get(this.inputName);
+    if (!control)
```
smadbe, line 126: no brackets->one line

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -26,7 +26,7 @@ export class GroupInviteUsersComponent implements OnInit, OnDestroy {
 
   readonly inputName = 'logins'
   state: 'empty'|'too_many'|'loading'|'ready' = 'empty';
-  inviteForm = this.formBuilder.group({ invitations: '' });
+  inviteForm = this.formBuilder.group({ logins: '' });
```
smadbe, line 30: That shows that repeating the constant "logins" here was not a good idea :-(

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
```diff
@@ -46,17 +48,24 @@ export class GroupInviteUsersComponent implements OnInit, OnDestroy {
     this.subscription?.unsubscribe();
   }
 
+  setState(newState: GroupInviteState): void {
+    if (this.state == newState) return;
+    if (this.state == 'loading') this.inviteForm.enable(); // enable the form only if the previous state was disabled
+    if (newState == 'loading') this.inviteForm.disable();
```
smadbe, line 54: === for string comparison

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
smadbe: IMO, `patchValue` is more when we work a big object with many attributes and we want to replace one. Here we want to reset one specific known attribute, we prefer if it properly crash if it does not work (instead of ignoring as patchValue would do).

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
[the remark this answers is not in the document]
smadbe: Several reasons:
- Here, it is not completely a responsibility of this class. It is a public method, an external class (or the template) may call this function at a wrong moment. 
- Sometimes, the fact a function is called even we are not supposed to is not a problem as the action is not harmful.
- By the important software engineering principle of "fail-fast", it is important to catch unexpected behavior as soon as possible. Otherwise we may get behavior very complicated to debug if this happens.
A nice class design tries to make unconsistent cases impossible to happen (e.g., it would be impossible to call action in case the state is loading). That is not always possible to achieve this, if so, we have to add guards.

smadbe (changes requested):
Details remaining

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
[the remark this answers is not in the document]
AyeRlS: Ok, thanks !

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
AyeRlS: Okay, I've changed it

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
AyeRlS: Fixed

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
AyeRlS: Fixed

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
AyeRlS: Fixed

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
AyeRlS: I removed the constant

src/app/modules/group/components/group-invite-users/group-invite-users.component.ts
AyeRlS: Fixed

smadbe (approved):
Looks good 👍

PR #192, opened 2020-11-06
Default item

smadbe:
* Load a default (from config) item by default as home page.
* Load this item by default in the left menu if the arrival page is not an item-related page.
* Deselect the currently selected item in the menu when not an item-related page


src/app/shared/helpers/item-route.ts
```diff
@@ -40,6 +53,17 @@ export function incompleteItemUrl(itemId: string): RouterCommands {
   return [ 'items', 'by-id', itemId ];
 }
 
+/*
+ * Url (as string) of the details page for the given item route
+ */
+export function linkToRoute(item: ItemRoute): string {
```
GabrielVidal1, line 59: shouldn't the name be `routeToLink` ? because the function returns a link from a route

GabrielVidal1 (changes requested):

AyeRlS (approved):

src/app/shared/helpers/item-route.ts
smadbe: Ah I did not mean "convert a link to a route" but "(get) link which leads to a route" ;-)
I will find a better name...

GabrielVidal1 (approved):

smadbe:
Naming changed. All the url* functions in this file should be rename to something proper as url is sometimes a string, sometimes an array for the router.

PR #193, opened 2020-11-09
Item edit description

### LINKED ISSUE #125: Item edition. 3. Update description
# Description

# Details
* Create a content tab (not with routing, just the content directly)
* Add the description field as a text field (as shown in the design) with is saved with the header save button is pressed.
* When empty (after trim), send "null" to the backend.

AyeRlS:
Fixes #125
Adding the item description input to the item edit page.

### DIFF
--- a/src/app/modules/item/item-routing.module.ts
+++ b/src/app/modules/item/item-routing.module.ts
@@ -7,7 +7,6 @@ import { ItemContentComponent } from './pages/item-content/item-content.componen
 import { ItemProgressComponent } from './pages/item-progress/item-progress.component';
 import { ItemCurrentSituationComponent } from './pages/item-current-situation/item-current-situation.component';
 import { ItemLogViewComponent } from './pages/item-log-view/item-log-view.component';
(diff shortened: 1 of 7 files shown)


src/app/modules/item/item-routing.module.ts
```diff
@@ -43,6 +44,12 @@ import { ItemLogViewComponent } from './pages/item-log-view/item-log-view.compon
         {
           path: 'edit',
           component: ItemEditComponent,
+          children: [
+            {
+              path: '',
+              component: ItemEditContentComponent
+            }
+          ]
```
smadbe, line 52: Your code does not use routing to choose tab so there is no routing to be configured to your component.

smadbe (changes requested):
Just one thing.

src/app/modules/item/item-routing.module.ts
AyeRlS: Fixed

smadbe (approved):

PR #194, opened 2020-11-09
Group user list

GabrielVidal1:
Added the group user list to the group composition tab.
A few things I'm not confident about:
* When the first and last name are null, what should be displayed in the 'Name' column?
* Same for the 'Member Since' column (the `memberSince` attribute can be null)
* Also, should the date format be something specific? or should that be something that depend on the language of the current user?

### DIFF
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -1,0 +1,48 @@
+<p-progressSpinner *ngIf="state === 'loading'"></p-progressSpinner>
+
+<p *ngIf="state === 'error'">Error while loading the group members</p>
+
+<p *ngIf="state === 'empty'">This group is empty, check below, the different ways to add users or sub-groups.</p>
+
+<alg-grid
+*ngIf="state === 'ready'"
+[data]="members"
+[showGear]="false"
+>
+  <ng-template #headerTemplate let-columns>
+    <tr>
+      <th>
+        <div class="header-container">
+          Name
+        </div>
+      </th>
+      <th>
+        <div class="header-container">
+          Login
+        </div>
+      </th>
+      <th>
+        <div class="header-container">
+          Member Since
+        </div>
+      </th>
+    </tr>
+  </ng-template>
+
+  <ng-template
+    #bodyTemplate
+    let-member
+  >
+    <tr class="p-selectable-row">
+      <td>
+        {{ member.user.firstName }} {{member.user.lastName}}
+      </td>
+      <td>
+        {{ member.user.login }}
+      </td>
+      <td>
+        {{ member.memberSinceText }}
+      </td>
+    </tr>
+  </ng-template>
+</alg-grid>
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -41,7 +41,7 @@
         {{ member.user.login }}
       </td>
       <td>
-        {{ member.memberSinceText }}
+        {{ member.memberSince | date }}
       </td>
     </tr>
   </ng-template>
--- a/src/app/modules/group/components/user-list/user-list.component.ts
+++ b/src/app/modules/group/components/user-list/user-list.component.ts
@@ -1,4 +1,5 @@
-import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Component, Input, OnChanges, OnDestroy, SimpleChanges } from '@angular/core';
+import { Subscription } from 'rxjs';
 import { Group } from '../../http-services/get-group-by-id.service';
 import { GetGroupMembersService, Member } from '../../http-services/get-group-members.service';
 
@@ -11,12 +12,14 @@ interface MemberAddition {
   templateUrl: './user-list.component.html',
   styleUrls: [ './user-list.component.scss' ]
 })
-export class UserListComponent implements OnChanges {
+export class UserListComponent implements OnChanges, OnDestroy {
 
   @Input() group? : Group;
   state: 'loading' | 'error' | 'empty' | 'ready' = 'loading';
 
   members: (Member&MemberAddition)[] = [];
+
+  private subscription?: Subscription;
 
   constructor(private getGroupMembersService: GetGroupMembersService) { }
 
@@ -27,8 +30,8 @@ export class UserListComponent implements OnChanges {
   private reloadData(): void {
     if (this.group) {
       this.state = 'loading';
-
-      this.getGroupMembersService.getGroupMembers(this.group.id)
+      this.subscription?.unsubscribe();
+      this.subscription = this.getGroupMembersService.getGroupMembers(this.group.id)
         .subscribe(
           members => {
             const dateFormatter = new Intl.DateTimeFormat('fr');
@@ -48,5 +51,9 @@ export class UserListComponent implements OnChanges {
     } else {
       this.state = 'error';
     }
+  }
+
+  ngOnDestroy(): void {
+    this.subscription?.unsubscribe();
   }
 }
--- a/src/app/modules/group/components/user-list/user-list.component.ts
+++ b/src/app/modules/group/components/user-list/user-list.component.ts
@@ -1,0 +1,52 @@
+import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Group } from '../../http-services/get-group-by-id.service';
+import { GetGroupMembersService, Member } from '../../http-services/get-group-members.service';
+
+interface MemberAddition {
+  memberSinceText: string,
+}
+
+@Component({
+  selector: 'alg-user-list',
+  templateUrl: './user-list.component.html',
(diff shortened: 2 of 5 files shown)


src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -0,0 +1,52 @@
...
+      this.state = 'loading';
+
+      this.getGroupMembersService.getGroupMembers(this.group.id)
+        .subscribe(
+          members => {
+            const dateFormatter = new Intl.DateTimeFormat('fr');
+
+            this.members = members.map(member => ({
+              ...member,
+              memberSinceText: member.memberSince === null ? '' : dateFormatter.format(member.memberSince)
```
smadbe, line 38: As it is just formatting, probably using the date pipe is a cleaner way to format date. (you'll save the addition as well)
Leave the default locale for now, we'll use the user's one when we will support internationalization.

src/app/modules/group/components/user-list/user-list.component.html
```diff
@@ -0,0 +1,48 @@
...
+    #bodyTemplate
+    let-member
+  >
+    <tr class="p-selectable-row">
+      <td>
+        {{ member.user.firstName }} {{member.user.lastName}}
+      </td>
+      <td>
+        {{ member.user.login }}
+      </td>
```
smadbe, line 42: It is correct that it is not beautiful when first/last name is null/empty. Maybe you could merge is a single "name" column and, as we did at other places, do "Gabriel Vidal (gvi)" or "gvi" if no last/first name. Maybe you could do your own pipe not have to redefine once more this rule (and replace other places where is it already done).

src/app/modules/group/components/user-list/user-list.component.html
```diff
@@ -0,0 +1,48 @@
...
+      <td>
+        {{ member.memberSinceText }}
```
smadbe, line 44: The case where it is null is quite rare. Having it empty in this case is ok.

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  ngOnChanges(_changes: SimpleChanges): void {
+    this.reloadData();
+  }
+
+  private reloadData(): void {
+    if (this.group) {
+      this.state = 'loading';
+
+      this.getGroupMembersService.getGroupMembers(this.group.id)
```
smadbe, line 31: As you did at the other places, probably you need to keep the subscription to cancel it if it changes to prevent request reordering causing the wrong content to be displayed.

src/app/modules/group/http-services/get-group-members.service.ts
```diff
@@ -0,0 +1,64 @@
...
+import { Observable } from 'rxjs';
+import { appConfig } from 'src/app/shared/helpers/config';
+import { map } from 'rxjs/operators';
+
+interface RawMember {
+  id: string,
+  member_since: string|null;
+  action: 'invitation_accepted'|'join_request_accepted'|'joined_by_code'|'added_directly';
+  user: {
+    login: string|null,
```
smadbe, line 12: login should not be  not nullable

smadbe (changes requested):
I answered your questions in the review.

src/app/modules/group/http-services/get-group-members.service.ts
```diff
@@ -0,0 +1,64 @@
...
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { appConfig } from 'src/app/shared/helpers/config';
+import { map } from 'rxjs/operators';
+
+interface RawMember {
+  id: string,
+  member_since: string|null;
+  action: 'invitation_accepted'|'join_request_accepted'|'joined_by_code'|'added_directly';
+  user: {
```
smadbe, line 11: Actually the `user` attribute is nullable. For now, you can filter out (in the service directly) the entries with null `user` as they are for members which are not users. (I thought this service was for user only but that's not the case)

src/app/modules/group/components/user-list/user-list.component.html
```diff
@@ -0,0 +1,48 @@
+<p-progressSpinner *ngIf="state === 'loading'"></p-progressSpinner>
+
+<p *ngIf="state === 'error'">Error while loading the group members</p>
+
+<p *ngIf="state === 'empty'">This group is empty, check below, the different ways to add users or sub-groups.</p>
+
+<alg-grid
+*ngIf="state === 'ready'"
+[data]="members"
+[showGear]="false"
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.html
GabrielVidal1: fixed, I created the `userDisplay` pipe for this. I still need to replace other place where this can be used.

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: fixed

src/app/modules/group/http-services/get-group-members.service.ts
GabrielVidal1: fixed

src/app/modules/group/http-services/get-group-members.service.ts
GabrielVidal1: fixed

src/app/shared/pipes/userDisplay.ts
```diff
@@ -0,0 +1,17 @@
...
+  lastName :string|null,
+}
+
+@Pipe({ name: 'userDisplay' })
+export class UserDisplayPipe implements PipeTransform {
+  transform<T extends UserInfos>(userInfos: T) : string {
+    if (userInfos.firstName === null || userInfos.lastName === null) {
+      return userInfos.login;
+    }
+    return `${userInfos.firstName} ${userInfos.lastName} (${userInfos.login})`;
```
smadbe, line 15: The case where only one of firstName/lastName is null is ignored. Probably we should be in the second scenario for that case, but not displaying "null" for the null one.

src/app/shared/pipes/userDisplay.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+interface UserInfos {
+  login: string,
+  firstName :string|null,
+  lastName :string|null,
+}
+
+@Pipe({ name: 'userDisplay' })
+export class UserDisplayPipe implements PipeTransform {
```
smadbe, line 10: I don't really like the "User Display" term... 
To be consistent with the other angular pipe, the better name would probably "user" and the class name "UserPipe".

src/app/shared/pipes/userDisplay.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+interface UserInfos {
+  login: string,
+  firstName :string|null,
+  lastName :string|null,
```
smadbe, line 6:
```suggestion
  login: string,
  firstName: string|null,
  lastName: string|null,
```

smadbe (changes requested):

src/app/modules/group/components/user-list/user-list.component.html
smadbe: do you plan to do it in this PR? It is maybe better.

GabrielVidal1: I created a separate PR (#197) for the user pipe, as some refactoring had to be done.

src/app/shared/pipes/userDisplay.ts
GabrielVidal1: fixed in PR #197

src/app/shared/pipes/userDisplay.ts
GabrielVidal1: fixed in PR #197

src/app/shared/pipes/userDisplay.ts
GabrielVidal1: fixed in PR #197

smadbe (approved):

smadbe (changes requested):
You still have to fix the columns in the member table.

GabrielVidal1:
I rebased from master to have the latest changes (the user pipe mainly) and used it in the member table (the name column to be exact)

smadbe (commented):
Is it normal there is no separation line between rows in the left column? 

Screenshot 2020-11-12 at 09 39 56

GabrielVidal1:
I fixed the missing separation between each row

smadbe:
You haven't pushed any changes 🤷‍♂️

smadbe (approved):

PR #195, opened 2020-11-09
Always fetching on refresh

smadbe:
I find this "optimization" quite confusing as, on return from editing, the former content was displayed while the refresh is on progress.
Now the loading spinner is displayed while refreshing, even the content is the same.

### DIFF
--- a/src/app/modules/group/services/group-datasource.service.ts
+++ b/src/app/modules/group/services/group-datasource.service.ts
@@ -28,18 +28,19 @@ export class GroupDataSource implements OnDestroy {
     private getGroupByIdService: GetGroupByIdService,
   ) {
     this.fetchOperation.pipe(
-      // switchMap does cancel the previous ongoing processing if a new one comes
-      switchMap(id => {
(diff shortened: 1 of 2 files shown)


GabrielVidal1 (approved):

src/app/modules/group/services/group-datasource.service.ts
```diff
@@ -28,18 +28,19 @@ export class GroupDataSource implements OnDestroy {
...
-          map(res => readyState(res)),
-          catchError(e => of(errorState(e)))
-        );
-
-        // if the fetched group is the same as the current one, do not change state to "loading" (silent refresh)
-        const currentState = this.state.value;
-        if (isReady(currentState) && currentState.data.id === id) return dataFetch;
-        else return concat(of(fetchingState()), dataFetch);
-      })
+      // on new fetch operation to be done: set "fetching" stae and fetch the data which will result in a ready or error state
```
AyeRlS, line 33:
```suggestion
      // on new fetch operation to be done: set "fetching" state and fetch the data which will result in a ready or error state
```

PR #204, opened 2020-11-13
Chapter grid component

### LINKED ISSUE #169: Item edition. step 4. Children reordering. a: Import component
## Motivations

## Subtasks
* Import the section-chapter-wrapper component, clean it so that it uses proper types etc

AyeRlS:
Fixes #169
Adding the chapter grid component


smadbe (commented):
As there are no data, I cannot say if it works.  There are still things that do not have to be there probably. I'll continue this PR myself, it will be easier.

smadbe (approved):
I approve my own changes ;-)

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -1,22 +1,52 @@
...
   styleUrls: [ './item-children-edit.component.scss' ]
 })
 export class ItemChildrenEditComponent implements OnChanges {
+  @Input() itemData?: ItemData;
 
-  @Input() data: ItemChild[] = [
-    { id: '100', title: 'First item', order: 1 },
-    { id: '50', title: 'Second item', order: 2 },
-  ];
+  state: 'loading' | 'error' | 'empty' | 'ready' = 'ready';
```
smadbe, line 15: Why hiding completely the component if it is empty ? The user still need to know there is no children to know he can add some.

smadbe (changes requested):
first comment for today. To be continued...

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
AyeRlS: I've added a message when the chapter don't have child

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -1,22 +1,51 @@
...
+      this.state = 'loading';
+      this.subscription?.unsubscribe();
+      this.subscription = this.getItemChildrenService
+        .get(this.itemData.item.id, this.itemData.currentResult.attemptId)
+        .subscribe(children => {
+          this.data = children.map((child, idx) => ({
+            id: child.id,
+            title: child.string.title,
+            order: idx
+          }));
```
smadbe, line 39: I prefer you do this "map" in a Rx pipe map function. (I know we do not do it everywhere in the existing code)
This way, if there is a problem (for instance there is no "string" in "child", it ends up as a proper error (in the block below) instead of crash the whole thing.

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -1,22 +1,51 @@
...
+          this.state = this.data.length ? 'ready' : 'empty';
```
smadbe, line 40: As you have data available in the template, I prefer you use data.length directly into the template rather than introducing and using this "empty" state.

src/app/modules/item/pages/item-edit-content/item-edit-content.component.ts
```diff
@@ -9,6 +10,10 @@ import { FormGroup } from '@angular/forms';
 export class ItemEditContentComponent {
   @Input() parentForm?: FormGroup;
 
-  constructor() {}
+  itemData$ = this.itemDataSource.itemData$;
```
smadbe, line 13: If we do not use routing, we should not use the datasource. It should be passed by input.

smadbe (changes requested):
👍  Only minor changes.

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts

src/app/modules/item/pages/item-edit-content/item-edit-content.component.ts
AyeRlS: Done

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -31,18 +32,19 @@ export class ItemChildrenEditComponent implements OnChanges {
       this.subscription?.unsubscribe();
       this.subscription = this.getItemChildrenService
         .get(this.itemData.item.id, this.itemData.currentResult.attemptId)
-        .subscribe(children => {
-          this.data = children.map((child, idx) => ({
+        .pipe(
+          map(children => children.map((child, idx) => ({
             id: child.id,
             title: child.string.title,
             order: idx
```
smadbe, line 39: Ah I didn't see it earlier: the order has to be the order returned by the service!

src/app/modules/item/pages/item-edit/item-edit.component.html
```diff
@@ -22,7 +22,12 @@
       </nav>
 
       <div class="bg-white" >
-        <alg-item-edit-content *ngIf="rla1.isActive" [parentForm]="itemForm"></alg-item-edit-content>
+        <alg-item-edit-content
+          *ngIf="rla1.isActive"
+          [parentForm]="itemForm"
+          [itemData$]="itemData$"
+        >
```
smadbe, line 29: Typically we do not pass observable but just change the value when it changes (which may be considered as valid as well, but at least that's not what we do). Check how it is done in `item-details.component.html`.

smadbe (changes requested):

src/app/modules/item/pages/item-edit/item-edit.component.html
AyeRlS: Done

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
AyeRlS: Fixed

src/app/modules/item/pages/item-edit-content/item-edit-content.component.ts
AyeRlS: Fixed

src/app/modules/item/pages/item-edit-content/item-edit-content.component.html
```diff
@@ -1,3 +1,9 @@
-<alg-section icon="fa fa-book" label="Description">
-  <alg-textarea [parentForm]="parentForm" inputName="description"></alg-textarea>
-</alg-section>
+<ng-container *ngIf="itemData && itemData.currentResult">
+
+  <alg-section icon="fa fa-book" label="Description">
+    <alg-textarea [parentForm]="parentForm" inputName="description"></alg-textarea>
+  </alg-section>
+
+  <alg-item-children-edit [itemData]="itemData" *ngIf="itemData.item.type === 'Chapter'"></alg-item-children-edit>
```
smadbe, line 7: for "Chapter" and "Skill"

smadbe (changes requested):

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -33,10 +33,10 @@ export class ItemChildrenEditComponent implements OnChanges {
...
         .pipe(
-          map(children => children.map((child, idx) => ({
+          map(children => children.map(child => ({
             id: child.id,
             title: child.string.title,
-            order: idx
+            order: child.order
           })))
         ).subscribe(children => {
           this.data = children;
```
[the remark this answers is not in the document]
AyeRlS: Fixed

src/app/modules/item/pages/item-edit-content/item-edit-content.component.html
AyeRlS: Fixed

smadbe (approved):
👍

PR #206, opened 2020-11-13
Adding subtitle input to item edit component

AyeRlS:


smadbe:
The bug should now be fixed in the backend.

smadbe (changes requested):
The subtitle is not supposed to look like the title (in size). And as they have no caption, they should (both) have a placeholder when empty. The design branch is doing it right.

smadbe (approved):
👍

PR #207, opened 2020-11-16
Group user list sort

### LINKED ISSUE #186: Group composition. 2. sort on user list
## Motivations
List composition of a group like in https://france-ioi.github.io/algorea-designs/01.Groups_02.Composition

## Subtasks
* Add sorting on user list based on member_since and user login

GabrielVidal1:
Fixes #186 
Add sorting on the columns of the user list in the group composition tab.

### DIFF
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -2,10 +2,7 @@
 
 <p *ngIf="state === 'error'">Error while loading the group members</p>
 
-<p *ngIf="state === 'empty'">This group is empty, check below, the different ways to add users or sub-groups.</p>
-
 <alg-grid
-  *ngIf="state === 'ready'"
   [data]="members"
   [showGear]="false"
   (sort)="onCustomSort($event)"
@@ -39,5 +36,9 @@
         {{ member.memberSince | date }}
       </td>
     </tr>
+  </ng-template>
+
+  <ng-template #emptymessageTemplate>
+    <p>This group is empty, check below, the different ways to add users or sub-groups.</p>
   </ng-template>
 </alg-grid>
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -2,7 +2,7 @@
 
 <p *ngIf="state === 'error'">Error while loading the group members</p>
 
-<alg-grid
+<alg-grid #grid
   [data]="members"
   [showGear]="false"
   (sort)="onCustomSort($event)"
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -8,17 +8,20 @@
   *ngIf="state === 'ready'"
   [data]="members"
   [showGear]="false"
+  (sort)="onCustomSort($event)"
 >
   <ng-template #headerTemplate let-columns>
     <tr>
-      <th>
+      <th pSortableColumn="user.login">
         <div class="header-container">
           Name
+          <p-sortIcon field="user.login"></p-sortIcon>
         </div>
       </th>
-      <th>
+      <th pSortableColumn="member_since">
         <div class="header-container">
           Member Since
+          <p-sortIcon field="member_since"></p-sortIcon>
         </div>
       </th>
     </tr>
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -1,12 +1,12 @@
-<p-progressSpinner *ngIf="state.tag === 'fetching'  && members.length == 0"></p-progressSpinner>
+<p-progressSpinner *ngIf="state === 'fetching'  && members.length == 0"></p-progressSpinner>
 
-<p *ngIf="state.tag === 'error'">Error while loading the group members</p>
+<p *ngIf="state === 'error'">Error while loading the group members</p>
 
 <alg-grid #grid
   [data]="members"
   [showGear]="false"
   (sort)="onCustomSort($event)"
-  [loading]="state.tag === 'fetching' && members.length > 0"
+  [loading]="state === 'fetching' && members.length > 0"
 >
   <ng-template #headerTemplate let-columns>
     <tr>
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -1,12 +1,12 @@
-<p-progressSpinner *ngIf="state === 'loading' && members.length == 0"></p-progressSpinner>
+<p-progressSpinner *ngIf="state.tag === 'fetching'  && members.length == 0"></p-progressSpinner>
 
-<p *ngIf="state === 'error'">Error while loading the group members</p>
+<p *ngIf="state.tag === 'error'">Error while loading the group members</p>
 
 <alg-grid #grid
   [data]="members"
   [showGear]="false"
   (sort)="onCustomSort($event)"
-  [loading]="state === 'loading' && members.length > 0"
+  [loading]="state.tag === 'fetching' && members.length > 0"
 >
   <ng-template #headerTemplate let-columns>
     <tr>
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -1,10 +1,12 @@
+<p-progressSpinner *ngIf="state === 'loading' && members.length == 0"></p-progressSpinner>
+
 <p *ngIf="state === 'error'">Error while loading the group members</p>
 
 <alg-grid
   [data]="members"
   [showGear]="false"
   (sort)="onCustomSort($event)"
-  [loading]="state === 'loading'"
+  [loading]="state === 'loading' && members.length > 0"
 >
   <ng-template #headerTemplate let-columns>
     <tr>
--- a/src/app/modules/group/components/user-list/user-list.component.html
+++ b/src/app/modules/group/components/user-list/user-list.component.html
@@ -1,11 +1,10 @@
-<p-progressSpinner *ngIf="state === 'loading'"></p-progressSpinner>
-
 <p *ngIf="state === 'error'">Error while loading the group members</p>
 
 <alg-grid
   [data]="members"
   [showGear]="false"
   (sort)="onCustomSort($event)"
+  [loading]="state === 'loading'"
 >
   <ng-template #headerTemplate let-columns>
     <tr>
--- a/src/app/modules/group/components/user-list/user-list.component.ts
+++ b/src/app/modules/group/components/user-list/user-list.component.ts
@@ -3,7 +3,7 @@ import { SortEvent } from 'primeng/api';
 import { merge, of, Subject } from 'rxjs';
 import { map, switchMap } from 'rxjs/operators';
 import { GridComponent } from 'src/app/modules/shared-components/components/grid/grid.component';
-import { errorState, FetchError, Fetching, fetchingState, isReady, Ready, readyState } from 'src/app/shared/helpers/state';
+import { fetchingState, isReady, readyState } from 'src/app/shared/helpers/state';
 import { Group } from '../../http-services/get-group-by-id.service';
 import { GetGroupMembersService, Member } from '../../http-services/get-group-members.service';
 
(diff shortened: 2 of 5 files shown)


src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -8,43 +10,48 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
   templateUrl: './user-list.component.html',
   styleUrls: [ './user-list.component.scss' ]
 })
-export class UserListComponent implements OnChanges, OnDestroy {
+export class UserListComponent implements OnInit {
```
smadbe, line 13: So you do not handle the change of group anymore.

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -8,43 +10,48 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
...
 
-  ngOnChanges(_changes: SimpleChanges): void {
-    this.reloadData();
+  ngOnInit(): void {
+    this.dataFetching.pipe(
+      switchMap(mode => {
+        if (this.group) {
+          if (mode === 'Reload')
+            this.state = 'loading';
+
```
smadbe, line 31: "Loading" while it is currently fetching the loading would make sense as well, but not via hiding the whole table. I see that the primeng table has a `loading` property that could be used for that.

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -8,43 +10,48 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
...
+          return this.getGroupMembersService.getGroupMembers(this.group.id, this.currentSort);
+        } else {
+          throw new Error('group is null');
+        }
+      })
+    ).subscribe(
+      members => {
+        this.members = members;
+
+        if (this.members.length === 0) this.state = 'empty';
```
smadbe, line 41: I don't find very relevant to have an "empty" state in this class, you can just check that `members.length === 0` in the template. 
Also the primeng table support the case where it is empty I think... yes there is this "emptyMessageTemplate" that could be used for that, would be better than reimplementing our own logic.

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -8,43 +10,48 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
...
-    } else {
-      this.state = 'error';
-    }
-  }
+  onCustomSort(event: SortEvent): void {
+    const sortMeta = event.multiSortMeta?.map(meta => (meta.order === -1 ? `-${meta.field}` : meta.field));
 
-  ngOnDestroy(): void {
-    this.subscription?.unsubscribe();
+    if (sortMeta && JSON.stringify(sortMeta) !== JSON.stringify(this.currentSort)) {
```
smadbe, line 52: (Having less state is always better (less inconsistency, ...))
I don't think storing `currentSort` is useful in anyway, it just has to be passed to the dataFetching, but not to be stored.
While quickly checking the impact, we can see it means that if `group` change, the sort is not kept, but that seems to be a better idea anyway. Note there is a "reset" method to be used on p-table especially for this case.

src/app/modules/group/http-services/get-group-members.service.ts
```diff
@@ -21,8 +21,8 @@ export interface Member {
   id: string,
   memberSince: Date|null;
   action: 'invitation_accepted'|'join_request_accepted'|'joined_by_code'|'added_directly';
-  user: {
-    login: string|null,
+  user: null|{
```
smadbe, line 24: Wasn't this done in the previous PR?

smadbe (changes requested):
Here is a first bunch of comments. Some of the comments are not really on your specific implementation but would also apply to other places, but it is not too late to start doing better (and fix the existing in another PR).
Globally, I think you're approach, with a subject, work.

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -8,43 +10,48 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
...
+export class UserListComponent implements OnInit {
 
   @Input() group? : Group;
   state: 'loading' | 'error' | 'empty' | 'ready' = 'loading';
+  currentSort: string[] = [];
 
   members: Member[] = [];
 
-  private subscription?: Subscription;
+  private dataFetching = new BehaviorSubject<'Reload'|'Sort'>('Reload');
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -8,43 +10,48 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
...
   constructor(private getGroupMembersService: GetGroupMembersService) { }
 
-  ngOnChanges(_changes: SimpleChanges): void {
-    this.reloadData();
+  ngOnInit(): void {
+    this.dataFetching.pipe(
+      switchMap(mode => {
+        if (this.group) {
+          if (mode === 'Reload')
+            this.state = 'loading';
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: The function `onCustomSort` is called when the data is refreshed (I don't know why) but if I get rid of the `currentSort` variable, the data will be reloaded continuously, so I kept it to not enter the if statement if the data is already sorted.

src/app/modules/group/http-services/get-group-members.service.ts
GabrielVidal1: When I rebase the previous PR, some commits were not included :/
I did another PR with the missing commits #209

src/app/modules/group/components/user-list/user-list.component.ts
smadbe: From what I see from ngprime code, when there is a change in the content, the reapply the sort... which makes sense if it is a static sort. I do not see any other easy solution...

src/app/modules/shared-components/components/grid/grid.component.ts
```diff
@@ -67,6 +67,7 @@ export class GridComponent implements OnInit, OnChanges {
   @Input() dataKey?: string;
   @Input() frozenWidth?: string;
   @Input() showGear = true;
+  @Input() loading = false;
```
smadbe, line 70: 👍  nice to do these changes in separated commits! 👍

src/app/modules/group/components/user-list/user-list.component.html
```diff
@@ -1,29 +1,23 @@
...
-<p *ngIf="state === 'empty'">This group is empty, check below, the different ways to add users or sub-groups.</p>
-
 <alg-grid
-*ngIf="state === 'ready'"
-[data]="members"
-[showGear]="false"
+  [data]="members"
+  [showGear]="false"
+  (sort)="onCustomSort($event)"
+  [loading]="state === 'loading'"
```
smadbe, line 7: The idea was to keep the spinner the first time but this table loading when only the sort is changing. It is a bit cheating but maybe you could rely on `state === 'loading' && members.length > 0`

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -8,43 +10,48 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
...
+        this.state = 'ready';
+      },
+      _err => {
+        this.state = 'error';
+      });
+  }
 
   ngOnChanges(_changes: SimpleChanges): void {
-    this.reloadData();
+    this.dataFetching.next([]);
```
smadbe, line 46: You should use the `reset` method on the table, see https://www.primefaces.org/primeng/showcase/#/table

smadbe (changes requested):

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.html
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -8,43 +10,48 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
...
+  ngOnInit(): void {
+    this.dataFetching.pipe(
+      switchMap(mode => {
+        if (this.group) {
+          if (mode === 'Reload')
+            this.state = 'loading';
+
+          return this.getGroupMembersService.getGroupMembers(this.group.id, this.currentSort);
+        } else {
+          throw new Error('group is null');
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -29,13 +29,16 @@ export class UserListComponent implements OnChanges {
 
           return this.getGroupMembersService.getGroupMembers(this.group.id, sort);
         } else {
-          throw new Error('group is null');
+          this.state = 'error';
```
smadbe, line 35: It is worse 😁. Actually, I missed it in the previous review, but having side effects (modifying this) in the switchMap is a very bad idea. There is a very common pattern we used often where we emit several state update in the switch map...

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -19,6 +20,8 @@ export class UserListComponent implements OnChanges {
 
   members: Member[] = [];
 
+  @ViewChild('grid', { static: true }) private grid?: GridComponent;
```
smadbe, line 23: is there a good reason to make it static? If not, it should not be.

smadbe (changes requested):

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: Should I revert this change and keep the `throw new Error('group is null');` instead?

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: fixed

GabrielVidal1:
I changed the way the data is fetched, now I use the state interface to return the data, and I pass the inputs to switchMap instead of using the component's `this`

src/app/modules/group/components/user-list/user-list.component.ts
```diff
@@ -14,37 +15,31 @@ import { GetGroupMembersService, Member } from '../../http-services/get-group-me
 export class UserListComponent implements OnChanges, OnDestroy {
 
   @Input() group? : Group;
-  state: 'loading' | 'error' | 'ready' = 'loading';
+
+  state: Ready<Member[]> | Fetching | FetchError = fetchingState();
 
   currentSort: string[] = [];
 
   members: Member[] = [];
```
smadbe: I do not really like storing twice the same data in the instance method. A simple solution I see is to store as "state" on line 28, the state.tag.

smadbe (changes requested):

src/app/modules/group/components/user-list/user-list.component.ts
GabrielVidal1: I kept it so the table is still visible during the sorting, otherwise the whole table is replace by the progress spinner.
I don't see another solution, I have to keep the data stored somewhere during the fetching

smadbe (approved):
👍  for the last fix. 
I do not recheck everything, I assume that is ok.

PR #210, opened 2020-11-17
Group observation

smadbe:
Add group observation in the menu. In practice, it just switch on/off this mode without changing anything to the content (just refreshing content), but that will be done in future work.
Easier to review commit per commit.


GabrielVidal1 (approved):

AyeRlS (approved):
Just one small thing, when the group name is long, the close arrow is not visible anymore and we can't go back unless we refresh.

smadbe:
> Just one small thing, when the group name is long, the close arrow is not visible anymore and we can't go back unless we refresh.

@AyeRlS Ah sorry I saw your remark after merging. Actually this [x] is just a stupid trick before a design does this properly.

PR #221, opened 2020-11-23
Item children reordering

### LINKED ISSUE #205: Item edition. step 5. Children reordering. c: reordering
## Motivations

## Subtasks
* Add reordering to ItemChildrenEditComponent
* let the user re-order the children
* on save (! handle the case when the list of children is not loaded yet or in error), save the children so only the `item_id` and `order` attributes.
* if possible only update children if there were a change

AyeRlS:
Fixes #205

Allow reordering of the children and implementing the service to save the new order


src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -50,4 +52,11 @@ export class ItemChildrenEditComponent implements OnChanges {
     }
   }
 
+  orderChanged(): void {
+    if (this.state == 'error') return;
+    this.parentForm?.patchValue({
+      children: this.data.map((child, idx) => ({ item_id: child.id, order: idx })),
+    });
```
smadbe, line 59: I do not have the impression that's the way to go to update your data to your parent form. Is it?

https://ritchiejacobs.be/angular-custom-form-component
(I haven't checked it in details yet)

smadbe (changes requested):

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
AyeRlS: The reason why I'm using that this.data variable is because of the p-table. 

I'm not the one changing the order of this.data, but primeng. And the p-table can't change the value inside a reactive form. So it seems like I have to use a variable and copy the content of that variable into the form.

Wouldn't creating a custom form component just move the problem?

smadbe: What I meant is I am not sure you are supposed to change `this.parentForm?.patchValue` from here. 
Either it is not a form component and the main form fetches the info from this component, or it is a form component and it behaves as a form component. Probably option (1) is easier and more relevant here, but if so, there is no `parentForm` to be known by this component.

src/app/modules/item/http-services/update-item.service.ts
```diff
@@ -0,0 +1,33 @@
+import { Injectable } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+import { Observable } from 'rxjs';
+import { assertSuccess, SimpleActionResponse } from '../../../shared/http-services/action-response';
+import { appConfig } from '../../../shared/helpers/config';
+import { map } from 'rxjs/operators';
+import { ItemChild } from '../components/item-children-edit/item-children';
```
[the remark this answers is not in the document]
AyeRlS: Done

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts

smadbe (changes requested):
I think that's better this way. Just some few more remarks, but it should be quite quick to apply.

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -12,11 +17,13 @@ import { map } from 'rxjs/operators';
 })
 export class ItemChildrenEditComponent implements OnChanges {
   @Input() itemData?: ItemData;
+  @Input() parentForm?: FormGroup;
```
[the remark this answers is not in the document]
AyeRlS: Fixed

src/app/modules/item/pages/item-edit-content/item-edit-content.component.ts
```diff
@@ -11,6 +12,8 @@ export class ItemEditContentComponent {
   @Input() parentForm?: FormGroup;
   @Input() itemData? : ItemData;
 
+  @Output() change = new EventEmitter<ChildData[]>();
```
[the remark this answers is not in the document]
AyeRlS: Fixed

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -78,6 +79,12 @@ export class ItemEditComponent implements OnDestroy {
     });
   }
 
+  updateForm(children: ChildData[]): void {
+    this.itemForm.patchValue({
+      children: children.map((child, idx) => ({ item_id: child.id, order: idx })),
+    });
+  }
+
```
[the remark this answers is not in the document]
AyeRlS: Done

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -79,22 +78,8 @@ export class ItemEditComponent implements OnDestroy {
...
-
-    if (children === null) return undefined;
-
-    const res: ItemChanges = {};
-    const childrenValue = children.value as ItemChild[];
-    if (childrenValue.length) res.children = childrenValue;
-
-    return res;
+  updateItemChanges(children: ChildData[]): void {
+    if (children.length) this.itemChanges.children = children.map((child, idx) => ({ item_id: child.id, order: idx }));
```
smadbe, line 82: Why `if (children.length) `?  Will not work as soon as you'll allow removing children.

smadbe (approved):

src/app/modules/item/pages/item-edit/item-edit.component.ts

Issue #217, opened 2020-11-18
Add new child to chapter


smadbe:
## Motivations
Allow to add a new item to a chapter ([design](https://france-ioi.github.io/algorea-designs/img/03.Activities_Edit_05.Edit%20chapter%20content_a.%20New%20content.png))

## Subtasks
* Use the "Add a content" component used in the design, but only with the leftmost input field 
* Put this component in shared as it will be used by groups as well (try not to be too specific to items in your code)
* This field has, as placeholder, with a darker font than in the design so that it is readable: "Enter a title to create a new child".
* When 3 letters have been typed, display the type selector (as in the design), with as option:
  - chapter, the usual chapter (folder) logo with, as description, "A new folder which can contain any activities."
  - task, with the "fa" "code" icon and, as description, "A new task which users can try to solve."
  - if the current item is a "skill: skill, fa-graduation-cap  icon, "A new sub-skill."
* When one of these icons are pressed:
  - if valid, the whole component, under the title is replaced by a loader, and the [itemCreate service](https://france-ioi.github.io/algorea-devdoc/api/#operation/itemCreate) is called with: 
    - item.type matching the pressed button
    - as `language_tag` , "en"
    - as `title` the given title
  - on service response:
    - on success: display green toast "The new item has been successfully created", remove the loader, clear the title field, and add the new item in the list of children
   - on failure, display red toast, remove the loader, keep the entered title as it was
* to be rechecked with @dle : shouldn't the "x" be on the right of the text field?

PR #226, opened 2020-11-25
Add child component

AyeRlS:
Adding a dedicated component to add new content to a chapter or skill as asked in #217 

Please note that clicking the icons and the service behind it hasn't been implemented yet. Also, the little arrow above this component hasn't been added yet.


smadbe (changes requested):
The close button 'x' is at the place I say it was not appropriate to place it to 🤷‍♂️

src/app/modules/shared-components/components/add-content/add-content.component.html
```diff
@@ -0,0 +1,30 @@
+<alg-sub-section icon="sign-in-alt" label="Add a content" [isClosed]="state == 'closed'" (close)="clearInput()">
+  <alg-input [parentForm]="newContentForm" name="title" placeholder="Enter a title to create a new child"></alg-input>
+
+  <div *ngIf="state == 'opened'" class="new-content-type">
```
smadbe, line 4: All string comparison must use `===`

src/app/modules/shared-components/components/add-content/add-content.component.html
```diff
@@ -0,0 +1,30 @@
+<alg-sub-section icon="sign-in-alt" label="Add a content">
+  <alg-input [parentForm]="newContentForm" name="title" placeholder="Enter a title to create a new child" [hasClearButton]="true"></alg-input>
```
smadbe, line 2: the clear button should be shown only if something has been typed!

src/app/modules/item/components/item-children-edit/item-children-edit.component.html
```diff
@@ -42,13 +42,16 @@
           </div>
         </ng-template>
       </p-table>
+
+      <span *ngIf="!data.length">No children for this chapter.</span>
+
+      <alg-add-content [allowSkills]="this.itemData?.item?.type === 'Skill'"></alg-add-content>
```
smadbe, line 48: * Don't use "this" in templates. 
* Item is not optional

src/app/modules/shared-components/components/input/input.component.scss
```diff
@@ -63,9 +63,25 @@
...
       font-style: italic;
-      color: lightgray;
+      color: rgb(118, 118, 121);
       font-size: 1.2rem;
     }
+
+
+  }
+
+  .clear-button {
```
smadbe, line 73: could you make it grey as the text and the left icon?

src/app/modules/shared-components/components/add-content/add-content.component.html
AyeRlS: Fixed

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  templateUrl: './add-content.component.html',
+  styleUrls: [ './add-content.component.scss' ],
+})
+export class AddContentComponent implements OnInit, OnDestroy {
+  @Input() allowSkills = false;
+
+  newContentForm: FormGroup = this.formBuilder.group({ title: '' });
+
+  state: 'opened' | 'closed' = 'closed';
+  subscription?: Subscription;
```
smadbe, line 16: private

src/app/modules/shared-components/components/input/input.component.ts
```diff
@@ -24,21 +23,28 @@ export class InputComponent implements OnInit, OnChanges {
   @Input() inputType = 'text';
   @Input() inputIcon = 'font'; // a font-awesome icon identifier
   @Input() buttonIcon?: string; // a font-awesome icon identifier for the input button
+  @Input() hasClearButton = false;
 
   @Output() click = new EventEmitter();
 
-  control: AbstractControl | null = null;
+  subscription?: Subscription;
```
smadbe, line 30: private

src/app/modules/shared-components/components/input/input.component.ts
```diff
@@ -24,21 +23,28 @@ export class InputComponent implements OnInit, OnChanges {
...
+  value = '';
```
smadbe, line 31: what do you use this value for?

src/app/modules/shared-components/components/input/input.component.html
```diff
@@ -5,6 +5,11 @@
     [ngClass]="{'has-button': buttonIcon}"
     [formControlName]="name"
   >
+  <span *ngIf="hasClearButton" class="clear-button">
+    <button (click)="clearInput()">
```
smadbe, line 9: Couldn't you just put a template variable on the input and do `#input.value=''` here?

smadbe (changes requested):

src/app/modules/item/components/item-children-edit/item-children-edit.component.html
AyeRlS: image
I really understand why but it won't let me compile without that '?'

AyeRlS: My bad, I forgot that '?'  doesn't work in the same way it does in a ts file. Can I use '!' here?

src/app/modules/shared-components/components/input/input.component.html
AyeRlS: Done

src/app/modules/shared-components/components/input/input.component.ts
AyeRlS: Use to verify that the length of the input and show the 'x' if there is something. But I've removed it to use a template variable

src/app/modules/shared-components/components/input/input.component.ts
AyeRlS: It has been removed

src/app/modules/shared-components/components/add-content/add-content.component.ts

src/app/modules/shared-components/components/input/input.component.scss
AyeRlS: Done

src/app/modules/shared-components/components/add-content/add-content.component.html
AyeRlS: It was supposed to be already done. Fixed

src/app/modules/item/components/item-children-edit/item-children-edit.component.html
AyeRlS: Using a '!' can cause problem if this component is used incorrectly so I've used another way

src/app/modules/shared-components/components/input/input.component.html
smadbe: My remark was for `(click)="clearInput()"`

smadbe:
Bug: I type 3 char, I press "x", the chapter/task button are still displayed

src/app/modules/shared-components/components/input/input.component.html
AyeRlS: After testing, the `#input.value=''` doesn't feat our needs, I've reverted to what was done previously.

As it is not the cleanest way to do it, in the future we should look into alternatives to see if there is a better way to do things, like using [this](https://angular.io/api/forms/DefaultValueAccessor)

smadbe (approved):

PR #234, opened 2020-11-26
Group descendant teams

### LINKED ISSUE #189: Group composition: 5. descendant teams
## Motivations
List composition of a group like in https://france-ioi.github.io/algorea-designs/01.Groups_02.Composition

## Subtasks
* Create a user/team filter under the children/descendant filter
* Use service: https://france-ioi.github.io/algorea-devdoc/api/#operation/groupTeamDescendantView
* Display column:
  - name 
  - parent group(s): comma separated list of parent group's name.
  - members: comma separated list of users' first and lastname
* allow sort by name

GabrielVidal1:
Fixes #189 
I'm just not sure what to display when the filter is set to 'Direct Children' and 'Teams'.
For now, an empty list is returned

### DIFF
--- a/src/app/modules/group/components/member-list/member-list.component.ts
+++ b/src/app/modules/group/components/member-list/member-list.component.ts
@@ -36,7 +36,7 @@ const descendantUsersColumns: Column[] = [
 const descendantTeamsColumns: Column[] = [
   { field: 'name', header: 'Name', sortable: true },
   { field: 'parentGroups', header: 'Parent group(s)' },
-  { field: 'parentGroups', header: 'Parent group(s)' },
+  { field: 'members', header: 'Member(s)' },
 ];
 
 interface Data {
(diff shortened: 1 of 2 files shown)


smadbe:
> I'm just not sure what to display when the filter is set to 'Direct Children' and 'Teams'.
> For now, an empty list is returned

Just the group children with type "Team". I don't understand what is the problem.

src/app/modules/group/components/member-list/member-list.component.ts
```diff
@@ -32,9 +33,15 @@ const descendantUsersColumns: Column[] = [
   { field: 'parentGroups', header: 'Parent group(s)' },
 ];
 
+const descendantTeamsColumns: Column[] = [
+  { field: 'name', header: 'Name', sortable: true },
+  { field: 'parentGroups', header: 'Parent group(s)' },
+  { field: 'parentGroups', header: 'Parent group(s)' },
```
smadbe, line 39: wrong field name?
I suppose that's because you haven't tested as you did not have teams. In such case, ask me to update the db (not always possible but may be very simple).
Here I have added teams in 164594055934439342 (which is a child of 947670356997220543) that can be managed by the usual user.

smadbe (changes requested):

src/app/modules/group/components/member-list/member-list.component.ts

smadbe (approved):

PR #239, opened 2020-11-27
Add child to array

AyeRlS:
Adding a new item to the item children edit component, but for the moment doesn't change the database

### DIFF
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
@@ -47,7 +47,7 @@
 
         <span *ngIf="!data.length">No children for this chapter.</span>
 
-        <alg-add-content [allowSkills]="itemData.item.type === 'Skill'" (childrenAdded)="addChild($event)"></alg-add-content>
+        <alg-add-content [allowSkills]="itemData.item.type === 'Skill'" (contentAdded)="addChild($event)"></alg-add-content>
 
       </div>
 
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
@@ -45,7 +45,7 @@
 
       <span *ngIf="!data.length">No children for this chapter.</span>
 
-      <alg-add-content [allowSkills]="this.itemData?.item?.type === 'Skill'"></alg-add-content>
+      <alg-add-content [allowSkills]="this.itemData?.item?.type === 'Skill'" (childrenAdded)="addChild($event)"></alg-add-content>
 
     </div>
 
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
@@ -26,6 +26,7 @@
                 <p-tableCheckbox [value]="rowData"></p-tableCheckbox>
               </td> -->
               <!-- <td class="type">text in blue...</td> -->
+              <td class="type">{{ rowData.type }}</td>
               <td class="main">{{ rowData.title }}
               </td>
             </tr>
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
@@ -46,7 +46,7 @@
 
         <span *ngIf="!data.length">No children for this chapter.</span>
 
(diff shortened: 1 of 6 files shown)


src/app/modules/item/components/item-children-edit/item-children-edit.component.html
```diff
@@ -46,7 +46,7 @@
 
         <span *ngIf="!data.length">No children for this chapter.</span>
 
-        <alg-add-content [allowSkills]="itemData.item.type === 'Skill'"></alg-add-content>
+      <alg-add-content [allowSkills]="itemData.item.type === 'Skill'" (childrenAdded)="addChild($event)"></alg-add-content>
```
smadbe, line 49: the indent deleted in front seems to be an error

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -9,6 +11,7 @@ import { Subscription } from 'rxjs';
 })
 export class AddContentComponent implements OnInit, OnDestroy {
   @Input() allowSkills = false;
+  @Output() childrenAdded = new EventEmitter<NewItem>();
```
smadbe, line 14: This is not necessarily children in other scenarios. So, "contentAdded" seems more appropriate.

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -4,6 +4,13 @@ import { ItemData } from '../../services/item-datasource.service';
 import { GetItemChildrenService } from '../../http-services/get-item-children.service';
 import { Subscription } from 'rxjs';
 import { map } from 'rxjs/operators';
+import { ItemType } from '../../../../shared/helpers/item-type';
+
+export interface NewItem {
```
smadbe, line 9: It is the duty of the emitter to define what he sends, not the listening component! 

What the creating component knows is just the title and the type, I would just return these two `{ title: string, type: ItemType }`. It is not the job of the add-content component to decide of the language a priori.

src/app/modules/item/components/item-children-edit/item-children.ts
```diff
@@ -1,6 +1,8 @@
+import { ItemType } from '../../../../shared/helpers/item-type';
 
 export interface ItemChild {
-  id: string,
+  id?: string,
   title: string | null,
   order: number,
+  type?: ItemType,
```
smadbe, line 7: It is not optional in item so it makes probably sense to have it non-option here and get it from the service (may be useful to display an icon is front anyway)

src/app/modules/item/components/item-children-edit/item-children.ts
```diff
@@ -1,6 +1,8 @@
+import { ItemType } from '../../../../shared/helpers/item-type';
 
 export interface ItemChild {
```
smadbe, line 3: If ItemChild is only used in this component, it could go (back) in the component maybe.

smadbe (changes requested):

src/app/modules/item/components/item-children-edit/item-children-edit.component.html
AyeRlS: Fixed

src/app/modules/shared-components/components/add-content/add-content.component.ts

src/app/modules/item/components/item-children-edit/item-children.ts
AyeRlS: Done

src/app/modules/item/components/item-children-edit/item-children.ts
AyeRlS: I was thinking about doing it in its own PR, But it's done, I've just moved it to this PR, tell me if you prefer that I display icons instead of what as been done in the design

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
AyeRlS: Done

src/app/modules/item/components/item-children-edit/item-children.ts
smadbe: Ah I didn't mean that the icon has to be done in this PR but that including the type as non optional would allow icons later. Ok I'll check what you have done.

smadbe: What you have done is ok.

smadbe (approved):
👍 looks good

PR #242, opened 2020-11-27
Child teams and sessions

### LINKED ISSUE #235: Group composition: Child Teams and Sessions
Implement the "teams" and "sessions" list for group children in group composition (use the filter from the service).

Use columns:
- for session: name, user count
- teams: name, member(s)

GabrielVidal1:
Fixes #235 
Add missing handling of 'Sessions' and 'Teams' filter for direct children of a group

### DIFF
--- a/src/app/modules/group/components/member-list/member-list.component.ts
+++ b/src/app/modules/group/components/member-list/member-list.component.ts
@@ -28,6 +28,11 @@ const groupsColumns: Column[] = [
   { field: 'userCount', header: 'User Count' },
 ];
 
+const nameUserCountColumns: Column[] = [
+  { field: 'name', header: 'Name', sortable: true },
+  { field: 'userCount', header: 'User Count' },
+];
+
 const descendantUsersColumns: Column[] = [
   { field: 'login', header: 'Name' },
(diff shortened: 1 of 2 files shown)


src/app/modules/group/components/member-list/member-list.component.ts
```diff
@@ -124,13 +130,13 @@ export class MemberListComponent implements OnChanges, OnDestroy {
               })),
             })));
         } else {
-          return of({
-            columns: descendantTeamsColumns,
-            rowData: [],
-          });
+          return this.getGroupChildrenService.getGroupChildren(groupId, sort, [ 'Team' ])
+            .pipe(map(children => ({
+              columns: groupsColumns,
```
smadbe, line 135: The columns are not correct. You may have guessed it (no need to put the type as it is fixed type) and asked me, it would have been easier than let me find the issue.

I've updated the issue.

smadbe (changes requested):

src/app/modules/group/components/member-list/member-list.component.ts

smadbe (approved):

PR #249, opened 2020-12-02
Remove child

### LINKED ISSUE #218: Chapter/Skills child removal
## Motivations
Allow to remove a child from a chapter (see [designs](https://france-ioi.github.io/algorea-designs/03.Activities_Edit_05.Edit-chapter-content) and design branch)

## Subtasks
* Allow to select one or several item in the child list
* Add the "select all" button, make it work properly
* Add "Remove from this chapter" button.
* When clicking the button, remove the selected item from the list (do not call any service)

AyeRlS:
Fixes #218

Removing selected child on save.

### DIFF
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
@@ -18,9 +18,9 @@
             <div class="summary-container">
               <p-tableHeaderCheckbox></p-tableHeaderCheckbox>
               <span class="select-all">Select all</span>
-              <div class="export-tools">
-                <span class="remove" (click)="deleteSelected()">
-                  <span>Remove</span>
+              <div class="export-tools" [ngClass]="{disabled: selectedRows.length === 0}">
+                <span class="remove" (click)="onRemove()">
+                  <span>Remove selected</span>
                   <i class="fa fa-trash-alt"></i>
                 </span>
               </div>
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.html
@@ -18,6 +18,12 @@
             <div class="summary-container">
               <p-tableHeaderCheckbox></p-tableHeaderCheckbox>
               <span class="select-all">Select all</span>
+              <div class="export-tools">
+                <span class="remove" (click)="deleteSelected()">
+                  <span>Remove</span>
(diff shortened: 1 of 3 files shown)


src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -60,6 +61,11 @@ export class ItemChildrenEditComponent implements OnChanges {
     this.data.push({ title: child.title, type: child.type });
   }
 
+  deleteSelected(): void {
```
smadbe, line 64: Call it "remove", that's is more correct than "delete". And "*Selected" is not very clear either.
So call it, for instance, `onRemove`

src/app/modules/item/components/item-children-edit/item-children-edit.component.html
```diff
@@ -3,16 +3,29 @@
...
               <td class="main">{{ rowData.title }}</td>
             </tr>
           </ng-template>
+          <ng-template pTemplate="summary">
+            <div class="summary-container">
+              <p-tableHeaderCheckbox></p-tableHeaderCheckbox>
+              <span class="select-all">Select all</span>
+              <div class="export-tools">
+                <span class="remove" (click)="deleteSelected()">
+                  <span>Remove</span>
```
smadbe, line 23: Call it "Remove selected" and grey it out (and so disable it) when nothing is selected (as in pending request panel)

smadbe (changes requested):
The selected row is not readable anymore :-/ We should try to fix that.
Can you do a proposal? Maybe just apply a light blue background instead?

src/app/modules/item/components/item-children-edit/item-children-edit.component.html
AyeRlS: Done

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts

AyeRlS:
As discussed, I have set the background color when selected to match unselected items, and I haven't found a way to disable that behavior.

src/app/modules/item/components/item-children-edit/item-children-edit.component.html
smadbe: It is not greyed-out, it is not displayed at all. I would prefer greyed-out.

smadbe (changes requested):
👍  except the greyed out

smadbe (approved):

PR #273, opened 2020-12-14
Add existing content

### LINKED ISSUE #237: Add existing content as item child - step 1: design
## Motivations
When editing a chapter (should be easy adaptable to groups), allow choosing among existing items to add as child. 

## Subtasks
* New input on the right of the previous create item input. (as in the design). (Add a "x" on the right of the input which clear the text as on the left one) 
* For both inputs: 
  - when the input is selected (cursor in it): clear the other input, and make it lighter (but don't disable it) 
  - if 1 or 2 letters are typed, add a red validation text that says that at least 3 char are required to create / search for items.
* When >=3 letters are typed, display the panel (with static data for now, use the search term as title to show the the search term is passed) with: 
  - a task/skill/chapter icon in front
  - a title 
  - a "+ Add" (instead of "v Insert") button on the right
* Having < 3char in the input field (or pressing the x button, or selection the left input) closes the panel

AyeRlS:
Fixes #237 

Added an field to add existing item as children to another item.

I'm not really convinced about the focus/blur behavior, for example: if you click one of the field and then click away, the info box saying "Enter at least 3 characters" will stay visible.


src/app/modules/shared-components/components/add-content/add-content.component.html
```diff
@@ -1,23 +1,47 @@
...
+      (focus)="onFocus('existingTitle')"
+      (blur)="onBlur()"
+      [ngClass]="{'input-light': focused === 'newTitle'}"
+      name="existingTitle"
+      placeholder="Search for existing content"
+    >
+    </alg-input>
+  </div>
+
+  <alg-message *ngIf="state === 'closed' && focused !== null" detail="Enter at least 3 characters"></alg-message>
```
smadbe, line 26: Checking a bit, I find the ngprime tooltip quite good for this usage, and only when the field is on focus (and <3char)

smadbe (changes requested):

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -12,33 +12,58 @@ export class AddContentComponent implements OnInit, OnDestroy {
   @Input() allowSkills = false;
   @Output() contentAdded = new EventEmitter<{ title: string, type: ItemType }>();
 
-  newContentForm: FormGroup = this.formBuilder.group({ title: '' });
+  newContentForm: FormGroup = this.formBuilder.group({ newTitle: '', existingTitle: '' });
 
   state: 'opened' | 'closed' = 'closed';
+  focused: 'existingTitle' | 'newTitle' | null = null;
```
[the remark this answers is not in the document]
AyeRlS: Those are the titles of the new/existing items. Maybe something like `existingItemTitle` would be clearer?

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -12,33 +12,58 @@ export class AddContentComponent implements OnInit, OnDestroy {
...
+      }
+    });
   }
 
   ngOnDestroy(): void {
     this.subscription?.unsubscribe();
   }
 
-  onClick(type: ItemType): void {
+  onFocus(name: 'existingTitle' | 'newTitle'): void {
```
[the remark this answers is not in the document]
AyeRlS: Fixed

src/app/modules/shared-components/components/add-content/add-content.component.html
AyeRlS: Done

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -12,33 +12,58 @@ export class AddContentComponent implements OnInit, OnDestroy {
   @Input() allowSkills = false;
   @Output() contentAdded = new EventEmitter<{ title: string, type: ItemType }>();
 
-  newContentForm: FormGroup = this.formBuilder.group({ title: '' });
+  newContentForm: FormGroup = this.formBuilder.group({ newTitle: '', existingTitle: '' });
 
   state: 'opened' | 'closed' = 'closed';
```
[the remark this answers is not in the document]

src/app/modules/shared-components/components/add-content/add-content.component.html
```diff
@@ -27,4 +51,16 @@
     </div>
   </div>
 
+  <p-table *ngIf="state === 'opened' && focused === 'existingTitle'" [value]="test">
+    <ng-template pTemplate="body" let-rowData>
+      <tr>
+        <td class="type">{{ rowData.type }}</td>
+        <td class="main">{{ currentValue }}</td>
+        <td>
+          <alg-button icon="fa fa-plus" label="Add"></alg-button>
```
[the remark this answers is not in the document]
AyeRlS: Done

smadbe (changes requested):
The 3 characters etc behavior seems to be broken on http://dev.algorea.org/branch/add-existing-content/

AyeRlS:
> 
> 
> The 3 characters etc behavior seems to be broken on http://dev.algorea.org/branch/add-existing-content/

Fixed

src/app/modules/shared-components/components/input/input.component.ts
```diff
@@ -23,11 +23,17 @@ export class InputComponent {
   @Input() buttonIcon?: string; // a font-awesome icon identifier for the input button
   @Input() hasClearButton = false;
 
+  @Input() pTooltip = '';
+  @Input() tooltipEvent = 'hover';
+  @Input() tooltipPosition = 'right';
+  @Input() tooltipDisabled = false;
```
smadbe, line 29: don't you have to disable it by default to prevent having it enabled everywhere?

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -37,10 +37,10 @@ export class AddContentComponent implements OnInit, OnDestroy {
 
   ngOnInit(): void {
     this.newContentForm.valueChanges.subscribe((change: {newTitle: string, existingTitle: string}) => {
-      if (change.newTitle.length < 3 && change.existingTitle.length < 3) this.state = 'closed';
+      if (change.newTitle.length < 3 && change.existingTitle.length < 3) this.expanded = false;
```
smadbe, line 40: you should compute the length of the trimmed string, otherwise typing 3 spaces launches a search and I don't think it will be accepted by the backend.

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -37,10 +37,10 @@ export class AddContentComponent implements OnInit, OnDestroy {
...
       else {
         this.currentValue = change.newTitle.length > 2 ? change.newTitle : change.existingTitle;
```
smadbe, line 42: use `>=3`, it will be easier to fix if this number changes (may be a constant btw)

src/app/modules/shared-components/components/add-content/add-content.component.ts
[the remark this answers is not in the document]
smadbe: "searchExisting" and "create" would be good maybe.

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -12,33 +12,58 @@ export class AddContentComponent implements OnInit, OnDestroy {
   @Input() allowSkills = false;
   @Output() contentAdded = new EventEmitter<{ title: string, type: ItemType }>();
 
-  newContentForm: FormGroup = this.formBuilder.group({ title: '' });
+  newContentForm: FormGroup = this.formBuilder.group({ newTitle: '', existingTitle: '' });
+
+  expanded = false;
+  focused: 'existingTitle' | 'newTitle' | null = null;
```
smadbe, line 18: You never put this variable back to null. In practice it means the input stays highlighted, even after a clear and doing something else.
BTW, why null and not undefined? Undefined is better in general in TS.

smadbe (changes requested):

src/app/modules/shared-components/components/input/input.component.ts
AyeRlS: The tooltip is enabled only if there is content in the `pTooltip `. `tooltipDisabled` is only doing something if the tooltip is shown. I would argue that both setting it to true or false by default can make sense, but in primeng its default value is false so I did the same

src/app/modules/shared-components/components/add-content/add-content.component.ts
AyeRlS: Fixed

src/app/modules/shared-components/components/add-content/add-content.component.ts

src/app/modules/shared-components/components/add-content/add-content.component.ts

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -54,13 +54,13 @@ export class AddContentComponent implements OnInit, OnDestroy {
...
-    this.newContentForm.get('existingTitle')?.setValue('');
+    if (this.focused == 'newTitle') return;
+    this.reset('newTitle');
   }
 
   onExistingFocus(): void {
-    this.focused = 'existingTitle';
-    this.newContentForm.get('newTitle')?.setValue('');
+    if (this.focused == 'existingTitle') return;
+    this.reset('existingTitle');
```
smadbe: === !!!

smadbe (changes requested):
If I fill an input then click on "x", the input stays focused even if doing something else.

src/app/modules/shared-components/components/add-content/add-content.component.ts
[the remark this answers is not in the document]

smadbe (approved):
👍 Look good 👍  Great!

PR #274, opened 2020-12-15
I18n

### LINKED ISSUE #21: Add i18n
From Ivan:
```
I am going to use `ngx-translate` module for internationalization.
We can translate the text inside the html files using the translate pipe or directive.
It supports the translation inside ts files using `TranslateService`.
The translation files can be stored at assets/i18n/{lang} or other locations we prefer in json format.
eg. en.json, fr.json, ...
```

smadbe:
Fixes #21 . 
Enable internationalization (i18n) for the project. Add some examples of translation and some explanation in the readme.
(to be reviewed commit per commit)

### DIFF
--- a/src/locale/messages.fr.xlf
+++ b/src/locale/messages.fr.xlf
@@ -26,7 +26,7 @@
           <context context-type="linenumber">12</context>
         </context-group>
       </trans-unit><trans-unit id="871322351859596725" datatype="html">
-        <source>Direct Children Only</source><target state="translated">Enfants uniquement</target>
+        <source>Direct Children Only</source><target state="translated">Enfants directs</target>
         <context-group purpose="location">
           <context context-type="sourcefile">src/app/modules/group/components/group-composition-filter/group-composition-filter.component.ts</context>
           <context context-type="linenumber">33</context>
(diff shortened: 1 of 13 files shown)


GabrielVidal1 (approved):

src/locale/messages.fr.xlf
```diff
@@ -25,6 +25,20 @@
           <context context-type="sourcefile">src/app/core/components/group-nav/group-nav.component.html</context>
           <context context-type="linenumber">12</context>
         </context-group>
+      </trans-unit><trans-unit id="871322351859596725" datatype="html">
+        <source>Direct Children Only</source><target state="translated">Enfants uniquement</target>
```
AyeRlS, line 29: I feel like we are losing the meaning of "direct" in "Direct Children Only".

Also, it seems strange that in English every word is capitalized and not in French.

smadbe: Capitalization of each word for titles is something which exists in English style and not in French. Anyway, here it is all uppercased ;-).
I find "Enfants directs uniquement" very long and boring in this menu, but maybe "Enfants directs" would be better than "Enfants uniquement".

AyeRlS (approved):

PR #277, opened 2020-12-16
Progress section

GabrielVidal1:
Add the progress-section and boolean-section components for the permissions edition modal.

The last commit is only to test the component ([here](http://dev.algorea.org/branch/progress-sections/#/items/by-id/4702;parentAttempId=0;path=/details) just below the item children) and will be removed before the merge.

### DIFF
--- a/src/app/modules/shared-components/components/progress-section/progress-section.component.ts
+++ b/src/app/modules/shared-components/components/progress-section/progress-section.component.ts
@@ -1,0 +1,55 @@
+import { Component, Input, OnChanges, SimpleChanges, Output, EventEmitter, ContentChild, TemplateRef, OnInit } from '@angular/core';
+
+export interface ProgressSectionValue<T> {
+  label: string,
+  comment: string,
+  value: T,
+  disabled?: boolean,
+}
+
+@Component({
+  selector: 'alg-progress-section',
+  templateUrl: './progress-section.component.html',
+  styleUrls: [ './progress-section.component.scss' ]
+})
+export class ProgressSectionComponent<T> implements OnChanges, OnInit {
+
+  @Input() title = '';
+  @Input() icon = '';
+
+  @Input() defaultValue?: T;
+  @Input() value?: T;
+
+  @Input() values: ProgressSectionValue<T>[] = [];
+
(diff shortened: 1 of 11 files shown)


src/app/modules/item/pages/item-content/item-content.component.ts
```diff
@@ -10,4 +10,35 @@ export class ItemContentComponent {
 
   @Input() itemData?: ItemData;
 
+  values = [{
```
smadbe, line 13: I would have found better to have all this info in the template instead than in a variable (which will be very long at the end) in the ts with the code logic. It would probably have to be done with ng-template and parsing the children to extract the metadata to reuse them... probably not very clean / type-safe. 
Except if you have an idea of better solution.
(at least, these vars should be placed in a separate file (in the final perm component))

smadbe (approved):
👍  very good. One minor change required.

src/app/modules/shared-components/components/progress-section/progress-section.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  @Input() title = '';
+  @Input() description = '';
+  @Input() icon = '';
+
+  @Input() defaultValue?: T;
+
+  @Input() values: ProgressSectionValue<T>[] = [];
+
+  @Input() collapsed = true;
+  @Input() disabled?: boolean;
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/item/pages/item-content/item-content.component.ts
GabrielVidal1: I couldn't find a better way to pass the different values to the component than this array, but I now use templates in some parts of the component (for example the body, header or the description) so that this component can be easily repurposed.

This greatly simplified the boolean-section component, which is now part of this PR (before, it inherited the progress-section component, but now it just uses it with different templates).
They can both be tested [here](http://dev.algorea.org/branch/progress-sections/#/items/by-id/4702;parentAttempId=0;path=/details) below the item children.

smadbe (approved):
Sounds good! 👍

PR #283, opened 2020-12-21
Add existing service

### LINKED ISSUE #238: Add existing content as item child - step 2: link services
## Motivations
When editing a chapter (should be easy adaptable to groups), allow choosing among existing items to add as child. 

## Subtasks
* Implement delayed search: when >3 char are typed and are not changed for 0.5sec (and only at that moment), launch the search to the search panel.
* Call the search service, with a limit of 4 results
* Display the loading spinner while searching is ongoing, error when occurs (without blocking everything)
* Make sure the previous searches are always cancelled when new requests are done
* When "+ Add" button is pressed, add the item at the end of the current children list.

AyeRlS:
Fixes #238
-
Step 2 of Add existing content as item child :
- Replace the static data with the actual data
- Add the selected item to the children of the current item.

### DIFF
--- a/src/app/modules/item/http-services/search-item.service.ts
+++ b/src/app/modules/item/http-services/search-item.service.ts
@@ -24,9 +24,8 @@ export class SearchItemService {
     excludedTypes?: ItemType[],
     limit = 4,
   ): Observable<ItemFound[]> {
-    let params = new HttpParams().set('search', searchString);
+    let params = new HttpParams().set('search', searchString).set('limit', limit.toString());
 
-    if (limit) params = params.set('limit', limit.toString());
     if (includedTypes) params = params.set('types_include', includedTypes.join(','));
     if (excludedTypes) params = params.set('types_exclude', excludedTypes.join(','));
 
--- a/src/app/modules/item/http-services/search-item.service.ts
+++ b/src/app/modules/item/http-services/search-item.service.ts
@@ -1,0 +1,38 @@
+import { Injectable } from '@angular/core';
+import { HttpClient, HttpParams } from '@angular/common/http';
+import { Observable } from 'rxjs';
+import { appConfig } from '../../../shared/helpers/config';
+import { ItemType } from '../../../shared/helpers/item-type';
+
+export interface ItemFound {
+  id: string,
+  title: string,
+  type: ItemType
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SearchItemService {
+
+  constructor(private http: HttpClient) {
+  }
+
+  search(
+    searchString: string,
+    includedTypes?: ItemType[],
+    excludedTypes?: ItemType[],
+    limit = 4,
+  ): Observable<ItemFound[]> {
+    let params = new HttpParams().set('search', searchString);
+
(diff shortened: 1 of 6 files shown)


src/app/modules/item/http-services/search-item.service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  search(
+    searchString: string,
+    includedTypes?: ItemType[],
+    excludedTypes?: ItemType[],
+    limit = 4,
+  ): Observable<ItemFound[]> {
+    let params = new HttpParams().set('search', searchString);
+
+    if (limit) params = params.set('limit', limit.toString());
```
smadbe, line 29: `limit` is always set. With `if (limit) ` you are only filtering "0" but there is no reason not to put the limit if =0.

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -2,6 +2,7 @@ import { Component, EventEmitter, Input, OnDestroy, OnInit, Output } from '@angu
 import { FormBuilder, FormGroup } from '@angular/forms';
 import { Subscription } from 'rxjs';
 import { ItemType } from '../../../../shared/helpers/item-type';
+import { ChildData } from '../../../item/components/item-children-edit/item-children-edit.component';
```
smadbe, line 7: You cannot make a component depends on a parent component.

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -25,33 +22,47 @@ export class AddContentComponent implements OnInit, OnDestroy {
...
+    this.subscriptions.push(
+      this.newContentForm.valueChanges.subscribe((changes: typeof defaultFormValues) => {
+        this.trimmedInputsValue = {
+          create: changes.create.trim(),
+          searchExisting: changes.searchExisting.trim(),
+        };
+      })
+    );
+
+    const existingTitleControl: Observable<string> | undefined = this.newContentForm.get('searchExisting')?.valueChanges;
```
smadbe, line 48: "newContentForm" was not supposed to create new content at first? (vs addContent)

smadbe (changes requested):
Looks good. Here are some comments. I'll think, I'll recheck the code in more details afterwards.

An item cannot be added twice as a child, so you should prevent that somehow. Either by greying out the "add" button in this case, or by failing properly on add (the list should not be reseted in this case).

src/app/modules/shared-components/components/add-content/add-content.component.ts
AyeRlS: I've renamed the form.

src/app/modules/item/http-services/search-item.service.ts
AyeRlS: Fixed.

src/app/modules/shared-components/components/add-content/add-content.component.ts
AyeRlS: Fixed

smadbe (approved):
I rechecked into more details. Looks good 👍

PR #288, opened 2021-01-05
Disabling children reordering on save

### LINKED ISSUE #282: Item children not frozen while loading
Bug: While saving an edited item, all inputs are frozen but the children. Children should also not be editable while saving.

AyeRlS:
Fixes #282 

It is just a simple css fix to disable pointer events


smadbe (changes requested):
Probably you should set opacity to 0.5 as well, otherwise it might seem buggy.

smadbe (changes requested):
On my computer, the list completely disappear. Didn't we say that the opacity is a number between 0 and 1?

smadbe (approved):

PR #289, opened 2021-01-05
Permissions edit dialog

### LINKED ISSUE #270: Permission panel through progress grid. Step 1: Importing the panel
## Motivations
Allow modifying  the permission via the progress matrix as show in the design branch.

## Subtasks
* On click on a cell, display the small summary box with:
  - the time spent
  - the score ring
  - the "access" button - only shown if the user has `can_grant_group_access` on the group and `can_grant_view` !== 'none' of the item
* the "access" button opens a modal as shown below:
  - For the title, indicate "<item title>: access given to <group/user name>" with the actual data.
  - for the rest, don't bind to the actual data for now
  - use the yellow color like here, but the icons and details of the "design" branch look ok (ask if you are unsure)
03 Activities_02 Participation_a Users situation (access collapsed)

GabrielVidal1:
Fixes #270

The last commit is just to grant the permission edition access by default for testing purposes (so that the current user can open the permission edition dialog box to test it [here](http://localhost:4200/#/items/by-id/4102;attempId=0;path=4702/details/progress)), it will not be merged. 
A few issues:
* The `timeSpent` value I get is absurdly large (or even negative) for some results. I think this is a backend problem.
* I followed the design in the design branch, so the progress sections are blue instead of yellow


src/app/shared/pipes/duration.ts
```diff
@@ -7,3 +7,18 @@ export class DurationToMinPipe implements PipeTransform {
...
+  transform(duration: number): string {
+    if (duration < 60) {
+      return `${Math.floor(duration)} min`;
+    } else if (duration < 60 * 24) {
+      return `${Math.floor(duration / 60)} hours`;
+    } else if (duration < 60 * 24 * 30) {
+      return `${Math.floor(duration / 60 / 24)} days`;
+    } else {
+      return `${Math.floor(duration / 60 / 24 / 30)} months`;
+    }
```
smadbe, line 22: The duration is in seconds (cfr the spec), not in minutes. That explains why you get results such as "1610 months", it is actually 26 months, which is quite realistic for a task which has been started 2y ago and not completed yet.
For the negative value,s I'll try to see if it can/should be fixed in the db, but anyway, you have to handle it properly. For instance, for the first category, you could just do `${Math.max(Math.floor(duration),0)} min`;

smadbe: Actually, for duration, probably you should re-use what has already been done with the "Duration" class (and the "toMin" pipe).

smadbe (commented):
An initial remark already.

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+@Component({
+  selector: 'alg-permissions-edit-dialog',
+  templateUrl: './permissions-edit-dialog.component.html',
+  styleUrls: [ './permissions-edit-dialog.component.scss' ]
+})
+export class PermissionsEditDialogComponent {
+
+  readonly canViewValues = [
```
smadbe, line 26: In my opinion, all these texts should be in the html, not here. Actually, the array could just be copy-pasted in the html.

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
```diff
@@ -0,0 +1,146 @@
...
+      value: 'all_with_grant',
+      label: 'All and grant',
+      comment: 'User(s) can also give "Can edit" access to others'
+    }
+  ];
+
+  constructor(
+    public dialogRef: MatDialogRef<PermissionsEditDialogComponent>,
+    public dialog: MatDialog,
+    @Inject(MAT_DIALOG_DATA) public data: PermissionsEditData,
```
smadbe, line 138: I do not have the full view of its usage, but this seems to be a way of sharing data between components. I don't think that's a great idea. If there is an actual reason to do it this way, we should discuss it first. Please check what it is used for.

src/app/shared/pipes/duration.ts
GabrielVidal1: fixed

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.html
```diff
@@ -0,0 +1,86 @@
...
+        (change)="this.data.permissions.can_edit = $event"
+        [values]="canEditValues"
+      ></alg-progress-section>
+      <alg-boolean-section
+        title="Can attach official sessions"
+        icon="fa fa-paperclip"
+        [defaultValue]="this.data.permissions.can_make_session_official"
+        (change)="this.data.permissions.can_make_session_official = $event"
+      >
+        <ng-template #label>User(s) may attach official sessions to this item, that will be visible to everyone in the content tab of the item</ng-template>
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
GabrielVidal1: I use the Primeng dialog component, so this wasn't needed anymore.

src/app/modules/item/components/item-chapter-view/group-situation-chapter-view/user-progress/user-progress.component.ts
```diff
@@ -1,6 +1,8 @@
 import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { MatDialog } from '@angular/material/dialog';
```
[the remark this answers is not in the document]
GabrielVidal1: fixed, I changed the dialog and now use the Primeng one.
I managed to do the same thing by changing a few things. The only major difference is the css that is not the same for the title of the dialog box (I couldn't figure out how to make it match the design)

src/app/modules/item/components/item-chapter-view/group-situation-chapter-view/user-progress/user-progress.component.ts
```diff
@@ -20,11 +19,21 @@ export class UserProgressComponent implements OnChanges {
 
   state: 'success'|'in-progress'|'no-score'|'not-started' = 'no-score';
 
+  @ViewChild(PermissionsEditDialogComponent) dialog!: PermissionsEditDialogComponent;
```
smadbe, line 24: I don't see why you would need to store your child more than at another place. You can pass arg without it. And you can use an input-output if you need to share "visible" (but I don't even see why it would be useful).

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.html
```diff
@@ -1,17 +1,22 @@
...
+  [modal]="true"
+  [draggable]="false"
+  [closable]="false"
+>
+  <p-header>
+    <div class="dialog-header">
+      <span class="dialog-handler">
+        <i class="fa fa-lock"></i>
+      </span>
+      <span class="dialog-title"> Access given to {{ data.title }}</span>
```
smadbe, line 13: Why calling that "data"? Everything is data ;-) If it is not fetched data or anonymous data, I think that is better to name for what it is.

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
```diff
@@ -20,13 +21,20 @@ export interface PermissionsEditData {
...
   @Input() data?: PermissionsEditData;
   @Output() close = new EventEmitter<Permissions>();
+  @Input() targetType: TypeFilter = 'Users';
 
   public visible = false;
 
+  targetTypeString = '';
+
+  ngOnInit(): void {
+    this.targetTypeString = this.targetType.slice(0, -1).toLowerCase();
```
smadbe, line 35: Wow that's is very hacky. If the filter is changed for any reason, the text here will break with nobody noticing it. It should be based on a solid state, a constant/enum or whatever which cannot be changed without breaking somewhere with the types.

smadbe (changes requested):

src/app/shared/helpers/duration.ts
```diff
@@ -28,10 +30,32 @@ export class Duration {
...
+    } else if (this.ms < HOURS) {
+      return `${Math.floor(this.ms / MINUTES)} min`;
+    } else if (this.ms < DAYS) {
+      return `${Math.floor(this.ms / HOURS)} hours`;
+    } else if (this.ms < MONTHS) {
+      return `${Math.floor(this.ms / DAYS)} days`;
+    } else if (this.ms < MONTHS * 12) {
+      return `${Math.floor(this.ms / MONTHS)} months`;
+    } else {
+      return `${Math.floor(this.ms / MONTHS / 12)} years`;
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.html
```diff
@@ -19,7 +19,33 @@
         icon="fa fa-eye"
         [defaultValue]="this.data.permissions.can_view"
         (change)="this.data.permissions.can_view = $event"
-        [values]="canViewValues"
+        [values]="[
```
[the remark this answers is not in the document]
GabrielVidal1: I'm thinking of putting these arrays in constants in a separate file, so I can use the $localize thing

src/app/modules/item/components/item-chapter-view/group-situation-chapter-view/user-progress/user-progress.component.html
```diff
@@ -44,3 +44,7 @@
     <span class="indicator"></span>
   </mat-menu>
 </ng-container>
+<alg-permissions-edit-dialog
+  [data]="{ title: title ? title : '', permissions: permissions}"
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/item/components/item-chapter-view/group-situation-chapter-view/user-progress/user-progress.component.ts
GabrielVidal1: fixed, I use a simple Input now

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.html
GabrielVidal1: fixed

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
GabrielVidal1: fixed

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.html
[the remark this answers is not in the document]
smadbe: So re-put it in the ts but using constants defined in another ts? Yes, that looks like a valid solution. I would have preferred a 100% html solution, but I cannot figure out a good one.

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
```diff
@@ -32,10 +28,20 @@ export class PermissionsEditDialogComponent implements OnInit {
...
+      case 'Users':
+        this.targetTypeString = 'user';
+        break;
+      case 'Groups':
+        this.targetTypeString = 'group';
+        break;
+      case 'Teams':
+        this.targetTypeString = 'team';
+        break;
+    }
```
smadbe, line 40: Don't forget to make these to be translatable as well.

src/app/shared/helpers/duration.ts
```diff
@@ -4,6 +4,7 @@ export const MINUTES = 60 * SECONDS;
 export const HOURS = 60 * MINUTES;
 export const DAYS = 24 * HOURS;
 export const MONTHS = 30 * DAYS;
+export const YEARS = 12 * MONTHS;
```
smadbe, line 7: probably that's more correct to say a year is 365 days than saying that it is 12 months of 30 days.

smadbe (changes requested):

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
GabrielVidal1: fixed

src/app/shared/helpers/duration.ts
GabrielVidal1: fixed

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
```diff
@@ -37,13 +37,13 @@ export class PermissionsEditDialogComponent implements OnInit {
...
+        this.targetTypeString = $localize`The user`;
         break;
       case 'Groups':
-        this.targetTypeString = 'group';
+        this.targetTypeString = $localize`The group`;
         break;
       case 'Teams':
-        this.targetTypeString = 'team';
+        this.targetTypeString = $localize`The team`;
         break;
```
smadbe: I would find cleaner to have this in the text ts as well. And so just passing the targetType to the function.

smadbe (approved):
small change remaining (no more review required)

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
GabrielVidal1: fixed

PR #290, opened 2021-01-05
Fix language deployment

smadbe:
Improve handling of the languages:
* allow refresh
* allow login (does not work on branches)
* include the lang picker
* root page autoredirects to English instead of the white page for selecting En/Fr


GabrielVidal1 (changes requested):
For the lang picker, I think the selection is not very intuitive, the currently selected language should be at the top of the list when clicked, or it should be in a different font (or not selectable)

smadbe:
> For the lang picker, I think the selection is not very intuitive, the currently selected language should be at the top of the list when clicked, or it should be in a different font (or not selectable)

Yes, I know. That's out of scope of this PR. Actually this design of the language picker has been validated earlier by Mathias (but probably would need a redesign). Here I just need a way for dev to switch at least.

GabrielVidal1 (approved):

PR #293, opened 2021-01-06
Fixing unnecessary request to update string service

### LINKED ISSUE #286: Sending request to change item string when nothing as been changed
When only the item has been changed, and not the item string, not request should be sent to change the item string

AyeRlS:
Fixes #286

### DIFF
--- a/src/app/modules/item/pages/item-edit/item-edit.component.ts
+++ b/src/app/modules/item/pages/item-edit/item-edit.component.ts
@@ -132,26 +132,33 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
 
   // Item string changes
   private getItemStringChanges(): ItemStringChanges | undefined {
-    const title = this.itemForm.get('title');
-    const subtitle = this.itemForm.get('subtitle');
-    const description = this.itemForm.get('description');
+    const titleControl = this.itemForm.get('title');
+    const subtitleControl = this.itemForm.get('subtitle');
+    const descriptionControl = this.itemForm.get('description');
+    const initialValues = this.initialFormData?.string;
 
-    if (title === null || description === null || subtitle === null) return undefined;
-    if (!this.dirtyControl(title, description, subtitle)) return {};
+    if (titleControl === null || subtitleControl === null || descriptionControl === null || !initialValues) return undefined;
 
-    return {
-      title: (title.value as string).trim(),
-      subtitle: (subtitle.value as string).trim() || null,
-      description: (description.value as string).trim() || null,
-    };
+    const res: ItemStringChanges = {};
+
+    const title = titleControl.value as string;
+    if (title !== initialValues.title) res.title = title.trim();
+
+    const subtitle = subtitleControl.value as string;
+    if (subtitle !== initialValues.subtitle) res.subtitle = subtitle.trim();
+
+    const description = descriptionControl.value as string;
+    if (description !== initialValues.description) res.description = description.trim();
+
+    return res;
   }
 
   private updateString(): Observable<void> {
     if (!this.initialFormData) return throwError(new Error('Missing ID form'));
     const itemStringChanges = this.getItemStringChanges();
-    if (!itemStringChanges) return throwError(new Error('Invalid form'));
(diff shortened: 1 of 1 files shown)


src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -127,6 +127,7 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
 
   // Item string changes
   private getItemStringChanges(): ItemStringChanges | undefined {
+    if (this.itemForm.pristine || !this.itemForm.touched) return {};
```
smadbe, line 130: why both pristine and not touched?

smadbe (changes requested):
Have you tested it? I still see the request. :-(

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: For some reason, on the first loading, the itemForm is marked as dirty. So I had to add the !this.itemForm.touched to handle this case.

smadbe: If marked dirty, why wouldn't the save/cancel display? They are based on dirty. And if the code goes through "markAsPristine" as is not pristine afterwards, something looks suspicious.

smadbe (changes requested):
And you should the same for the other request. If there is nothing to update in children, it is useless to send a PUT with `{}` as data to be changed..

smadbe:
> Have you tested it? I still see the request. :-(

I cannot reproduce my issue, strange... So you can ignore my remark.

src/app/modules/item/pages/item-edit/item-edit.component.ts
AyeRlS: So in the end, the way the save/cancel display work was conflicting with what I was trying to do. So I've changed the way to check to see if the fields have been changed

smadbe:
It still does the empty update while there is no update to children. Haven't you tried to fix that?
(test: http://dev.algorea.org/branch/update-string-when-needed/fr/#/items/by-id/972755911692173036;attempId=0;path=4702,4102,1980584647557587953/edit - I add and remove a space in the description and save, both strings (let's say that's ok) and children requests are sent)

AyeRlS:
I didn't understood there was also a problem with the other service, I'm going to fix that too.

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -120,30 +120,40 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
...
+
+    const title = titleControl.value as string;
+    if (title !== initialValues.title) res.title = title.trim();
+
+    const subtitle = subtitleControl.value as string;
+    if (subtitle !== initialValues.subtitle) res.subtitle = subtitle.trim();
+
+    const description = descriptionControl.value as string;
+    if (description !== initialValues.description) res.description = description.trim();
+
```
smadbe, line 148: Don't you loose the fact that when string is empty for subtitle and description, you were pushing "null" ?

smadbe: 👍  (It was a check you wrote a few weeks ago after many iterations. It should not be me who makes you notice that you do not handle this anymore.)

smadbe (changes requested):
I open a item with an empty description, I add a space, remove it, I still have a request sent... 😔

AyeRlS:
All I'm seeing are OPTIONS and GET requests, what item are you testing this on?

smadbe:
http://dev.algorea.org/branch/update-string-when-needed/fr/#/items/by-id/972755911692173036;attempId=0;path=4702,4102,1980584647557587953/edit 
Screenshot 2021-01-12 at 15 37 56

AyeRlS:
The item I was using for my test was the problem, it should be fixed now.

smadbe (approved):
Looks ok. I hope you have tested it carefully.

PR #296, opened 2021-01-07
Join group through item page

### LINKED ISSUE #252: Join group through item page (initial version)
## Motivations
Allow the user to join a group that corresponds to an activity. (cfr [designs](https://france-ioi.github.io/algorea-designs/03.Activities_00.Header_b2.Access-to-activity) or "Activity with access code" in [design branch](http://dev.algorea.org/branch/design))

## Subtasks
On the item page:
* if `prompt_to_join_group_by_code=true` and the current user has can_view=info on the item, display the "Access to activity" panel (see "Activity with access code" in [design branch](http://dev.algorea.org/branch/design), [pick design from there](https://github.com/France-ioi/AlgoreaFrontend/tree/design))
* only show the action button if at least one character has been entered
* use the [groupsCodeCheck](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupsCodeCheck) service to check if the service is valid. 
  - if invalid, display the error in a toast
  - if valid, display a simple confirm/cancel confirmation dialog (as in `pending-chages-guard`)
* on  confirmation, call the [groupsJoinByCode](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupsJoinByCode) service 
  -  on failure, display a toast with the usual unexpected error message
  - on success, display a success toast
* disable the action button while waiting for the responses to

GabrielVidal1:
Fixes #252 

When the user successfully joins the group, should the item page reload? because this prompt will still be visible on the page for now


src/app/modules/item/components/item-header/access-code-view/access-code-view.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  invalidCodeReasonToString(reason: InvalidCodeReason): string {
+    switch (reason) {
+      case 'already_member':
+        return $localize`You are already a member of this group`;
+      case 'conflicting_team_participation':
+        return $localize`There is conflicting group participation`;
+      case 'frozen_membership':
+        return $localize`Your membership is frozen`;
```
GabrielVidal1, line 58: I am not sure of the meaning of `conflicting_team_participation` and `frozen_membership` so I don't know if the message is really accurate, same goes for the french translation of these two

smadbe: You can use:
- conflicting_team_participation :  "You cannot join this team as it would conflict with another team you belong to."
- frozen_membership: "This group does not allow any membership change."
- team_conditions_not_met: "You cannot join this team as it would break the entry conditions of content it is participating to."

src/app/modules/item/components/item-header/item-header.component.ts
```diff
@@ -1,11 +1,19 @@
...
-export class ItemHeaderComponent {
+export class ItemHeaderComponent implements OnChanges {
   @Input() itemData?: ItemData;
+
+  showAccessCodeField = false;
+
+  ngOnChanges(_changes: SimpleChanges): void {
+    if (!this.itemData) return;
+    this.showAccessCodeField = this.itemData.item.prompt_to_join_group_by_code &&
+      this.itemData.item.permissions.can_view !== 'none';
```
smadbe, line 17: The specs says "the current user has can_view=info on the item", that is not the same!

src/app/modules/item/http-services/join-by-code.service.ts
```diff
@@ -0,0 +1,81 @@
...
+  rootActivityId: string|null,
+  rootSkillId: string|null,
+}
+
+export type InvalidCodeReason = 'no_group'|'frozen_membership'|'already_member'|'conflicting_team_participation'|'team_conditions_not_met';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class JoinByCodeService {
```
smadbe, line 43: These services may be used in group pages as well, so it should be in "shared" module I think.

src/app/modules/item/http-services/join-by-code.service.ts
```diff
@@ -0,0 +1,81 @@
...
+export interface GroupInfos {
+  managers: {
+    id: string,
+    login: string,
+    firstName: string|null,
+    lastName: string|null,
+  }[],
+  name: string,
+  requireLockMembershipApprovalUntil: string|null,
+  requirePersonalInfoAccessApproval: string|null,
```
smadbe, line 32: The goal of the non raw interface is also to convert raw values to a less primitive value. These should be transformed to date (or just removed if not used for now).

smadbe (changes requested):
Actually, I forgot one major requirement in the description:
If the code is valid but the group does not have this item as root_activity_id or root_skill_id (depending of the type of this item), ask confirmation to the user "The code does not correspond to the group attached to this page. Are you sure you want to join the group "..."?"

> When the user successfully joins the group, should the item page reload? because this prompt will still be visible on the page for now

As the perm may (and will probably) change after joining the group, yes, you should reload the content.

src/app/modules/item/components/item-header/item-header.component.ts
GabrielVidal1: fixed

src/app/modules/item/http-services/join-by-code.service.ts
GabrielVidal1: fixed

src/app/modules/item/http-services/join-by-code.service.ts
GabrielVidal1: fixed

src/app/modules/item/components/item-header/access-code-view/access-code-view.component.ts
GabrielVidal1: fixed

src/app/modules/item/components/item-header/access-code-view/access-code-view.component.ts
```diff
@@ -0,0 +1,83 @@
...
+        acceptIcon: 'fa fa-check',
+        rejectLabel: $localize`Cancel`,
+        accept: () => {
+          this.joinGroup(groupCode);
+        }
+      });
+    });
+  }
+
+  joinGroup(code: string): void {
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/item/components/item-header/access-code-view/access-code-view.component.html
```diff
@@ -5,6 +5,8 @@
       type="text"
       i18n-placeholder placeholder="Group code"
       (keyup)="0"
+      [(ngModel)]="code"
+      [disabled]="state == 'loading'"
```
smadbe, line 9: ===

src/app/modules/item/components/item-header/access-code-view/access-code-view.component.ts
```diff
@@ -13,6 +13,7 @@ import { ItemData } from '../../../services/item-datasource.service';
 export class AccessCodeViewComponent {
 
   @Input() itemData?: ItemData;
+  @Output() reload = new EventEmitter<void>();
```
smadbe, line 16: A output called  "reload" for an AccessCodeViewComponent is not really clear. (which may be used in different context)

smadbe (changes requested):
Few details.

src/app/modules/item/components/item-header/access-code-view/access-code-view.component.html
GabrielVidal1: fixed

src/app/modules/item/components/item-header/access-code-view/access-code-view.component.ts
GabrielVidal1: fixed

src/app/modules/item/components/item-header/access-code-view/access-code-view.component.ts
```diff
@@ -13,7 +13,7 @@ import { ItemData } from '../../../services/item-datasource.service';
 export class AccessCodeViewComponent {
 
   @Input() itemData?: ItemData;
-  @Output() reload = new EventEmitter<void>();
+  @Output() reloadItemPage = new EventEmitter<void>();
```
smadbe, line 16: In yesterday's remark, I added "which may be used in different context". So not sure it should be tight to an item page (possibly with some modifications of the component). Here what we just want to know is that there were a change and so that the parent component can decide what to do. So we could use for instance: "codeAccepted" or "membershipChanged" or ...

smadbe (approved):

PR #299, opened 2021-01-11
Permissions fetching and displaying

### LINKED ISSUE #271: Permission panel through progress grid. Step 2: read-only granted permissions
## Motivations
Display the actual granted permissions.
03 Activities_02 Participation_a Users situation (access)

## Subtasks
Base yourself on the future service specs (https://github.com/France-ioi/AlgoreaBackend/issues/643), for instance you may get: 
``` 
{
  "can_view": {
    "granted_only_group": "info",
    "computed": "content",
    "granted_group_membership": "none",
    "granted_self": "none",
    "granted_other": "none"
  },
  "can_enter_from": {
    "granted_only_group": "9999-12-31 23:59:59",
    ...
  }
}
```
* Use the "granted_only_group" of each to display the current permissions. 
* see https://france-ioi.github.io/algorea-devdoc/design/access-rights/items/ and https://france-ioi.github.io/algorea-devdoc/api/#operation/updatePermissions for possible values
* do not implement save yet, do not check extra constraints
* do not implement the extra lock icons etc

GabrielVidal1:
Fixes #271

Because the service is not ready in the backend, this hasn't been properly tested.

### DIFF
--- a/src/app/modules/item/components/item-chapter-view/group-situation-chapter-view/user-progress/user-progress.component.ts
+++ b/src/app/modules/item/components/item-chapter-view/group-situation-chapter-view/user-progress/user-progress.component.ts
@@ -1,6 +1,8 @@
-import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Component, Input, OnChanges, OnDestroy, SimpleChanges } from '@angular/core';
+import { Subscription } from 'rxjs';
 import { Group } from 'src/app/modules/group/http-services/get-group-by-id.service';
 import { Item } from 'src/app/modules/item/http-services/get-item-by-id.service';
+import { GetGroupPermissionsService } from 'src/app/shared/http-services/get-group-permissions.service';
 import { Permissions } from '../../../permissions-edit-dialog/permissions-edit-dialog.component';
 import { TypeFilter } from '../composition-filter/composition-filter.component';
 import { Progress } from '../group-situation-chapter-view.component';
@@ -10,10 +12,11 @@ import { Progress } from '../group-situation-chapter-view.component';
   templateUrl: './user-progress.component.html',
   styleUrls: [ './user-progress.component.scss' ]
 })
-export class UserProgressComponent implements OnChanges {
+export class UserProgressComponent implements OnChanges, OnDestroy {
 
   @Input() title?: string;
 
+  @Input() targetGroupId?: string;
   @Input() watchedGroup?: Group;
   @Input() item?: Item;
   @Input() userProgress?: Progress;
@@ -34,7 +37,15 @@ export class UserProgressComponent implements OnChanges {
     is_owner: true,
   };
 
-  visibleDialog = false;
+  dialog: 'loading'|'opened'|'closed' = 'closed';
+
+  private subscription?: Subscription;
+
+  constructor(private getGroupPermissionsService: GetGroupPermissionsService) {}
+
+  ngOnDestroy(): void {
+    this.subscription?.unsubscribe();
+  }
 
   ngOnChanges(_changes: SimpleChanges): void {
     if (!this.userProgress || !this.watchedGroup || !this.item) return;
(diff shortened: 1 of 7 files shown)


src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.html
```diff
@@ -29,8 +29,8 @@
       <alg-boolean-section
         i18n-title title="Can Enter"
         icon="fa fa-door-open"
-        [defaultValue]="this.permissions.can_enter_from"
-        (change)="this.permissions.can_enter_from = $event"
+        [defaultValue]="this.permissions.can_enter_from !== null"
+        (change)="this.permissions.can_enter_from = null"
       >
         <ng-template #label i18n>{{targetTypeString}} may enter this item (a contest or time-limited chapter)</ng-template>
       </alg-boolean-section>
```
GabrielVidal1, line 36: Should I handle the Can Enter section with the date picker like the design? I didn't for now because the date picker component is not part of the project now and I think it should be added in a different PR
image

src/app/shared/http-services/get-group-permissions.service.ts
```diff
@@ -0,0 +1,40 @@
...
+  can_watch: Permissions<'none'|'result'|'answer'|'answer_with_grant'>,
+  can_edit: Permissions<'none'|'children'|'all'|'all_with_grant'>,
+  can_enter_from: {
+    granted_only_group: string,
+  },
+  can_enter_until: {
+    granted_only_group: string,
+  },
+  can_make_session_official: boolean,
+  is_owner: boolean,
```
GabrielVidal1, line 26: I feel like I forgot some fields here but don't know exactly what

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.html
smadbe: I think we discussed it at the previous PR already: do not handle "can enter" for now.

src/app/modules/item/components/item-chapter-view/group-situation-chapter-view/user-progress/user-progress.component.ts
```diff
@@ -1,6 +1,8 @@
-import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { Component, Input, OnChanges, OnDestroy, SimpleChanges } from '@angular/core';
```
smadbe, line 1: I am sorry but I miss some important design "flaws" in the previous PRs.

1) This component is stored inside the first component. This is probably something done by the design, but that makes the project become a mess. There is no hierarchy between component. Oh here there is even an addition layer of hierarchy.
And the way it was organized was supposed to be: the actual content with data logic etc should be in "/pages" while independent component just for display should be "/components". So here, the part with the matrix should be in pages while the cell component should be in "component".

2) This should be the page (so the component of the matrix somehow) handling what "info" popup and permission dialog is open etc. There is only one open at a time anyway... It is like implementing is in the button component.  The user progress should be a design only thing.

3) (related with 2) I missed it but possibly that explain why this matrix is so slow, you have a matrix on possibly 15 columns, 1500 rows, so 22500 components, each storing their item, group, permissions, a state, possibly a rx observable, ... 

4) Each of these possibly 22500 components is filled with "find", which iterate over every element of a very long list to find its value. You could create the datastructure first and fill it while reading the data once. Using a O(n^2) algorithm instead of O(n) may have a big impact when that's related with thousands of cells.

Maybe all this may be fixed different PR when this PR is merged(?)

smadbe (changes requested):
If I test it, I get a request on a buggy url `/api/groups/672913018859223173/permissions/670968966872011405/39530140456452546)` which fails of course, and make the "access" button disappear 🤔

PR #301, opened 2021-01-13
Add subgroup

### LINKED ISSUE #294: Add a subgroup. 1. add new subgroup
## Motivations
As it is done in items and as it is shown on the design branch (on a group page > composition > add subgroup), let a manager add subgroups to a group.

## Subtasks
* Only if the user can manager group membership of the current group:
* Add the "add subgroup" blue block under the group composition table
* Use the same UI as in items: new content vs existing content.
* When entering a title for new content, the type choices are: 'Class', 'Club', 'Friends', 'Other'
* On click the group is created (https://france-ioi.github.io/algorea-devdoc/api/#operation/groupCreate) and added as subgroup (https://france-ioi.github.io/algorea-devdoc/api/#operation/groupAddChild).
* The usual errors are handled

GabrielVidal1:
Fixes #294 

I just have a few css problems to fix but the logic can be reviewed for now.


src/app/modules/group/components/add-sub-group/add-sub-group.component.ts
```diff
@@ -0,0 +1,82 @@
...
+      title: $localize`Friends`,
+      description: $localize`Friends`,
+    },
+    {
+      type: 'Other',
+      icon: 'fa fa-book',
+      title: $localize`Other`,
+      description: $localize`Other`,
+    },
+  ];
```
GabrielVidal1, line 56: I'm not sure what description and icon to have for each of the new group category

AyeRlS (approved):

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -18,26 +23,25 @@ const defaultFormValues = { create: '', searchExisting: '' };
   templateUrl: './add-content.component.html',
   styleUrls: [ './add-content.component.scss' ],
 })
-export class AddContentComponent implements OnInit, OnDestroy {
-  @Input() allowSkills = false;
-  @Output() contentAdded = new EventEmitter<AddedItem>();
+export class AddContentComponent<Type, SearchedValue extends AddedContent<Type>> implements OnInit, OnDestroy {
+
+  @Input() newContentTypes: NewContentType<Type>[] = [];
+  @Input() foundValues: SearchedValue[] = [];
```
smadbe, line 29: Not clear what it is exactly for. To clarify, you should call it "resultsFromSearch".

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -18,26 +23,25 @@ const defaultFormValues = { create: '', searchExisting: '' };
   templateUrl: './add-content.component.html',
   styleUrls: [ './add-content.component.scss' ],
 })
-export class AddContentComponent implements OnInit, OnDestroy {
-  @Input() allowSkills = false;
-  @Output() contentAdded = new EventEmitter<AddedItem>();
+export class AddContentComponent<Type, SearchedValue extends AddedContent<Type>> implements OnInit, OnDestroy {
+
+  @Input() newContentTypes: NewContentType<Type>[] = [];
```
smadbe, line 28: newContentTypes -> allowedTypesForNewContent

src/app/modules/group/http-services/group-creation.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+export class GroupCreationService {
+
+  constructor(private http: HttpClient) { }
+
+  create(name: string, type: 'Class'|'Team'|'Club'|'Friends'|'Other'|'Session'): Observable<GroupCreateResponse> {
+    const params = new HttpParams().set('name', name).set('type', type);
```
smadbe, line 24: Looks equivalent to `new HttpParams({ fromObject: { name: name, type: type } });` that I find cleaner. (to be checked)

src/app/modules/group/http-services/search-group.service.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+@Injectable({
+  providedIn: 'root'
+})
+export class SearchGroupService {
+
+  constructor(private http: HttpClient) { }
+
+  search(
```
smadbe, line 26: You overuse the generic types here. If you need two types with different allowed type, just define two types and maybe use "extends" if you do not want to repeat id, name, desc.

src/app/modules/group/http-services/group-creation.service.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+@Injectable({
+  providedIn: 'root'
+})
+export class GroupCreationService {
+
+  constructor(private http: HttpClient) { }
+
+  create(name: string, type: 'Class'|'Team'|'Club'|'Friends'|'Other'|'Session'): Observable<GroupCreateResponse> {
```
smadbe, line 23: This service should return directly an id and succeed only on success. Check create item.

src/locale/messages.fr.xlf
```diff
@@ -934,6 +934,12 @@
           <context context-type="sourcefile">src/app/modules/group/pages/managed-groups/managed-groups.component.html</context>
           <context context-type="linenumber">3</context>
         </context-group>
+      </trans-unit><trans-unit id="8471940861423120175" datatype="html">
+        <source>Group successfully added as child group</source><target state="new">Group successfully added as child group</target>
```
smadbe, line 938: Please do all the translations you introduced.

smadbe (changes requested):

src/app/modules/group/http-services/group-creation.service.ts
GabrielVidal1: fixed

src/app/modules/shared-components/components/add-content/add-content.component.ts
GabrielVidal1: fixed

src/app/modules/shared-components/components/add-content/add-content.component.ts
GabrielVidal1: fixed

src/app/modules/group/http-services/group-creation.service.ts
GabrielVidal1: fixed

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -18,26 +23,25 @@ const defaultFormValues = { create: '', searchExisting: '' };
   templateUrl: './add-content.component.html',
   styleUrls: [ './add-content.component.scss' ],
 })
-export class AddContentComponent implements OnInit, OnDestroy {
-  @Input() allowSkills = false;
-  @Output() contentAdded = new EventEmitter<AddedItem>();
+export class AddContentComponent<Type, SearchedValue extends AddedContent<Type>> implements OnInit, OnDestroy {
```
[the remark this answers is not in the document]
GabrielVidal1: fixed

src/app/modules/item/components/add-item/add-item.component.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  private dataFetching = new Subject<string>();
+
+  constructor(
+    private searchItemService: SearchItemService
+  ) {
+    this.dataFetching.pipe(
+      tap(_ => this.state = 'loading'),
+      debounce(() => timer(300)),
+      switchMap(value => this.searchItemService.search(value, [ 'Chapter', 'Course', 'Task' ])),
```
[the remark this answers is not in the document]
GabrielVidal1: fixed, I moved the call of the search service in the add-content component. That allowed me to remove the 'state' input (now the parent component pass its own search function to the add-content component).
I just not 100% confident about the way I removed the `tap(_ => this.state = 'loading')` and changed the state to fetching inside the `switchMap`, but it works as expected.

src/locale/messages.fr.xlf
GabrielVidal1: I merged master into this branch and completed the missing translations

src/app/modules/group/http-services/search-group.service.ts
GabrielVidal1: fixed

src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -58,8 +59,15 @@ export class AddContentComponent<Type, SearchedValue extends AddedContent<Type>>
       existingTitleControl.pipe(
         map(value => value.trim()),
         filter(value => this.checkLength(value)),
-      ).subscribe(value => {
-        this.search.emit(value);
+        switchMap(value => merge(
+          of(fetchingState()),
+          of(value).pipe(delay(300), switchMap(value =>
+            (this.searchFunction ? this.searchFunction(value).pipe(map(readyState)) : throwError(new Error('no search function'))),
```
smadbe, line 65: this.searchFunction not set is something which is unexpected, so don't pollute the code with that, just throw at the beginning of "init" if not set and put it in another variable (which will not be optional) that you can use here.

smadbe (changes requested):

src/app/modules/item/components/add-item/add-item.component.ts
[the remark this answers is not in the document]
smadbe: Very good refactoring! 👍

smadbe:
It seems there is html/css issue
Screenshot 2021-01-15 at 18 19 07

smadbe:
Ah and the creation does not work with the title I put... but it seems you pass the name in URL. The POST payload should be the body (see the other POSTs).

smadbe:
And I would expect the table to look like in items.
Screenshot 2021-01-15 at 18 22 41
Screenshot 2021-01-15 at 18 23 15

src/app/modules/shared-components/components/add-content/add-content.component.ts
GabrielVidal1: fixed

smadbe:
Have you tested it? If add subgroups in the "Pixal" group, but nothing seems to be added.

GabrielVidal1:
I fixed the problem, the addGroup service was not called

smadbe:
* BTW, in this list, when there is an error, it is not recovered (even after changing tabs). (bug)
Screenshot 2021-01-20 at 10 27 01
* Don't forget to fix conflicts (there should not be any merge on master soon)

smadbe (changes requested):
* The behavior is still strange when pressing "add" where the panel collapse but nothing happens before the green toast (so, especially is the query is slow). 
* I don't think there are reasons to refresh the full page in this case. Just the list below (only if on the "subgroups" part) should be reloaded... maybe we should force the switch to the "subgroups" tab actually... (here it is the opposite, as it reloads all the page, it always goes on "users" tab, which will never contain the new subgroup of course)

GabrielVidal1:
> * BTW, in this list, when there is an error, it is not recovered (even after changing tabs). (bug)

I couldn't reproduce it... can you give me the group which gives this error? 

> * Don't forget to fix conflicts (there should not be any merge on master soon)

I'll rebase on master when all the changes have been made.

> * The behavior is still strange when pressing "add" where the panel collapse but nothing happens before the green toast (so, especially is the query is slow).

I could change it and make the component have a loading icon when the new group is being added, but that means moving the component's logic from the group-composition component to the add-subgroup component and I'm not sure it's the right way to do it (It's better to keep the logic and UI separated)

> * I don't think there are reasons to refresh the full page in this case. Just the list below (only if on the "subgroups" part) should be reloaded... maybe we should force the switch to the "subgroups" tab actually... (here it is the opposite, as it reloads all the page, it always goes on "users" tab, which will never contain the new subgroup of course)

I fixed it, the filter of the member list switches to subgroups when the new sub-group is added.

smadbe:
> > * BTW, in this list, when there is an error, it is not recovered (even after changing tabs). (bug)
> 
> I couldn't reproduce it... can you give me the group which gives this error?

It is not an error on a specific group, it is just that I got a network issue, so the fetching failed, but then it was not trying to fetch anymore. It looks like the usual case when an subject goes to "error" and so is completed and does not pass any more events.

> > * The behavior is still strange when pressing "add" where the panel collapse but nothing happens before the green toast (so, especially is the query is slow).
> 
> I could change it and make the component have a loading icon when the new group is being added, but that means moving the component's logic from the group-composition component to the add-subgroup component and I'm not sure it's the right way to do it (It's better to keep the logic and UI separated)

The group composition can act (disable, hide, put a loader on, set a state) on the child, no need to move the logic to the child for that. Couldn't you use the primeng-blockui on it with a loader as content? (in the parent or child template)
Hum actually if the "add" fails, the user is to redo the search as well... but that's not a big deal (anyway we'll not be able to have a proper solution for the creation+add if only the second request fails)

> > * I don't think there are reasons to refresh the full page in this case. Just the list below (only if on the "subgroups" part) should be reloaded... maybe we should force the switch to the "subgroups" tab actually... (here it is the opposite, as it reloads all the page, it always goes on "users" tab, which will never contain the new subgroup of course)
> 
> I fixed it, the filter of the member list switches to subgroups when the new sub-group is added.

Screenshot 2021-01-22 at 10 56 41
There are two unexpected white lines around the "add" button.

smadbe:
And another big bug I think:
If I search for "test", I add the "test sans user", I ended up with several "test sans user" group. I first thought it was the backend not checking for duplicate... but actually the frontend create a new group each time ! 😔

GabrielVidal1:
I think I managed to track and fix all the mentioned issues.
For the loading of the add-sub-group component when a group is being added, I tried to use the primeng-blockui component but it produced very weird UI bugs so I used a simple boolean passed from the group-composition page to the add-sub-group component.
Also, if I try to add an existing group to a group that is already a parent group, there is the default error of the backend 'The action cannot be executed. If the problem persists, contact us'.
I think this could be avoided by having 'Already added' instead of the 'Add' button in the list when we try to search for groups that are already children of the selected group.

src/app/modules/group/components/member-list/member-list.component.ts
```diff
@@ -180,4 +181,9 @@ export class MemberListComponent implements OnChanges, OnDestroy {
       this.dataFetching.next({ groupId: this.group.id, filter: this.currentFilter, sort: this.currentSort });
     }
   }
+
+  setFilter(filter: Filter): void {
+    if (this.compositionFilter) this.compositionFilter.setFilter(filter);
```
smadbe, line 186: I suppose that is equivalent to 
```
this.compositionFilter?.setFilter(filter);
```
which is a bit shorter/cleaner

src/app/modules/group/pages/group-composition/group-composition.component.ts
```diff
@@ -52,7 +52,7 @@ export class GroupCompositionComponent implements OnChanges {
       _ => {
         this.displaySuccess($localize`Group successfully added as child group`);
         if (this.memberList) {
-          this.memberList.onFilterChange({ directChildren: true, type: TypeFilter.Groups });
+          this.memberList.setFilter({ directChildren: true, type: TypeFilter.Groups });
```
smadbe, line 57: same remark: `this.memberList?.setFilter...`

smadbe (changes requested):
Thanks for the improvements.
Remaining strange behavior:
- when adding the existing "test sans user" (I made the usual user manager of it) in "pixal",  I get each time a success... but I assume that's more a service issue? But it contradicts your "Also, if I try to add an existing group to a group that is already a parent group, there is the default error of the backend" 🤔
- I still have no loader during the add/create request

src/app/modules/group/pages/group-composition/group-composition.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/member-list/member-list.component.ts

GabrielVidal1:
The loading should now be working.
However, adding a group that is already a subgroup doesn't trigger an error from the backend anymore, I can't find a way to reproduce the error 🤔

smadbe (approved):
👍  Thanks

PR #303, opened 2021-01-14
Permissions readonly

GabrielVidal1:
Fixes #271

Because the service is not ready in the backend, this hasn't been properly tested.
I just checked the backend, and the feature has been approved but is not merged yet. I'll test this when it's been merged.

PR #305, opened 2021-01-15
Permissions fetching/updating

GabrielVidal1:
This PR fixes the PR #303 that merged code that was not tested yet (because the service wasn't ready too).
The first 3 commits fix the fetching and displaying of the permissions and the last 2 add the saving feature.

~~After some testing, when I change some permissions and save them (using the service to update permissions) and I open dialog box again, the permissions will appear unchanged. I think this problem comes from the backend.~~

Edit: It work perfectly now that I get permissions from the 'granted' field instead of 'granted_via_group_membership' field


src/app/modules/shared-components/components/progress-section/boolean-section/boolean-section.component.ts
```diff
@@ -1,30 +1,24 @@
...
   @Input() title = '';
   @Input() icon = '';
 
-  @Input() defaultValue?: boolean;
+  @Input() value = false;
 
   @ContentChild('description') descriptionTemplate?: TemplateRef<any>;
   @ContentChild('label') labelTemplate?: TemplateRef<any>;
 
   @Output() change = new EventEmitter<boolean>();
```
smadbe, line 18: Shouldn't you do a two-way binding here? So you would need to call the output `valueChange`. And then you could use more easily this component.
This way, you would not have to do 
```
        [value]="permissions.can_view"
        (change)="this.permissions.can_view = $event"
```  
but just 
```
[(value)]="permissions.can_view"
```

(it may apply to other component as well... to be checked...)

smadbe (changes requested):

src/app/modules/shared-components/components/progress-section/boolean-section/boolean-section.component.ts
GabrielVidal1: fixed, I used two way binding for the progress-section and the boolean-section components

smadbe (approved):

PR #317, opened 2021-02-17
Group associated activity

### LINKED ISSUE #313: "Associated activity" in group edit (part 1)
## Motivations
Be able to set the associated activity of group. Activities from group you belong to are those you have as root in the "activities" tab of the left menu.

## Subtasks
* Use a white background on group edition page, create a section "Group information" for the title and description.
* Create an "Associate Activity" section in which:
   - if there is a `root_activity_id`, display its name (on click go to this item) and a "remove" button. 
  - if there is none, display temporarily "There is no such activity"
* As this section loads content, it has to display a loader/error as usual.

GabrielVidal1:
Fixes #313 

I just had to remove the responsiveness of the fiels in the group edition page to make the new Associated Activity section display properly, I can fix it later if necessary.

### DIFF
--- a/src/app/modules/group/components/associated-activity/associated-activity.component.ts
+++ b/src/app/modules/group/components/associated-activity/associated-activity.component.ts
@@ -36,16 +36,17 @@ export class AssociatedActivityComponent implements OnDestroy, ControlValueAcces
   ) {
     this.activityFetching.pipe(
       switchMap(activityId => {
-        if (activityId === null) return of(readyState(null));
+        if (activityId === null) return of(readyState({ id: null, name: null }));
         return merge(
           of(fetchingState()),
-          this.getItemByIdService.get(activityId).pipe(map(item => readyState(item.string.title))),
+          this.getItemByIdService.get(activityId).pipe(map(item => readyState({ id: activityId, name: item.string.title }))),
         );
       })
     ).subscribe(state => {
       this.state = state.tag;
       if (isReady(state)) {
-        this.rootActivityName = state.data;
+        this.rootActivityId = state.data.id;
+        this.rootActivityName = state.data.name;
         this.onChange(this.rootActivityId);
       }
     });
--- a/src/app/modules/group/components/associated-activity/associated-activity.component.ts
+++ b/src/app/modules/group/components/associated-activity/associated-activity.component.ts
@@ -7,6 +7,8 @@ import { map, switchMap } from 'rxjs/operators';
 import { GetItemByIdService } from 'src/app/modules/item/http-services/get-item-by-id.service';
 import { fetchingState, isReady, readyState } from 'src/app/shared/helpers/state';
 
+type ActivityId = string;
+
 @Component({
   selector: 'alg-associated-activity',
   templateUrl: './associated-activity.component.html',
@@ -21,14 +23,16 @@ import { fetchingState, isReady, readyState } from 'src/app/shared/helpers/state
 })
 export class AssociatedActivityComponent implements OnDestroy, ControlValueAccessor {
 
-  rootActivityId: string|null = null;
-  rootActivityName: string|null = null;
+  rootActivity: null|{
+    id: ActivityId,
+    name: string|null,
+  } = null;
 
   state: 'fetching'|'ready'|'error' = 'fetching';
 
-  private activityFetching = new Subject<string|null>();
+  private activityFetching = new Subject<ActivityId|null>();
 
-  private onChange: (value: string|null) => void = () => {};
+  private onChange: (value: ActivityId|null) => void = () => {};
 
   constructor(
     private getItemByIdService: GetItemByIdService,
@@ -36,17 +40,19 @@ export class AssociatedActivityComponent implements OnDestroy, ControlValueAcces
   ) {
     this.activityFetching.pipe(
       switchMap(activityId => {
-        if (activityId === null) return of(readyState({ id: null, name: null }));
+        if (activityId === null) return of(readyState(null));
         return merge(
           of(fetchingState()),
-          this.getItemByIdService.get(activityId).pipe(map(item => readyState({ id: activityId, name: item.string.title }))),
+          this.getItemByIdService.get(activityId).pipe(map(item => readyState({
+            id: activityId,
+            name: item.string.title,
+          }))),
         );
       })
     ).subscribe(state => {
       this.state = state.tag;
       if (isReady(state)) {
-        this.rootActivityId = state.data.id;
-        this.rootActivityName = state.data.name;
+        this.rootActivity = state.data;
       }
     });
   }
@@ -55,11 +61,11 @@ export class AssociatedActivityComponent implements OnDestroy, ControlValueAcces
     this.activityFetching.complete();
   }
 
-  writeValue(rootActivityId: string|null): void {
+  writeValue(rootActivityId: ActivityId|null): void {
     this.activityFetching.next(rootActivityId);
   }
 
-  registerOnChange(fn: (value: string|null) => void): void {
+  registerOnChange(fn: (value: ActivityId|null) => void): void {
     this.onChange = fn;
   }
 
@@ -73,10 +79,9 @@ export class AssociatedActivityComponent implements OnDestroy, ControlValueAcces
   }
 
   onRemove(): void {
-    if (this.rootActivityId === null) return;
+    if (this.rootActivity === null || this.state !== 'ready') return;
 
-    this.rootActivityId = null;
-    this.rootActivityName = null;
-    this.onChange(this.rootActivityId);
+    this.rootActivity = null;
+    this.onChange(null);
   }
 }
--- a/src/app/modules/group/components/associated-activity/associated-activity.component.ts
+++ b/src/app/modules/group/components/associated-activity/associated-activity.component.ts
@@ -31,14 +31,14 @@ export class AssociatedActivityComponent implements OnDestroy, ControlValueAcces
 
   state: 'fetching'|'ready'|'error' = 'fetching';
 
-  private activityFetching = new Subject<ActivityId|null>();
+  private activityData = new Subject<ActivityId|null>();
 
   private onChange: (value: ActivityId|null) => void = () => {};
 
   constructor(
     private getItemByIdService: GetItemByIdService,
   ) {
-    this.activityFetching.pipe(
+    this.activityData.pipe(
       switchMap(activityId => {
         if (activityId === null) return of(readyState(null));
         return merge(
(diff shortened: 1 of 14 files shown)


smadbe:
If I go to "http://dev.algorea.org/branch/group-associated-activity/en/#/groups/by-id/672913018859223173/edit" with the usual user.
Screenshot 2021-02-18 at 10 07 21
 
* I get a "Error while loading the root activity"
* The first line of the issue "Use a white background on group edition page" does not seem to be applied
* The styles of "group information" and "associated activity" are different

GabrielVidal1:
I fixed the UI problems, the error message was due to a test statement I left in the code, it's fixed now.

smadbe:
Screenshot 2021-02-19 at 10 05 03

- Why such a large font for "No Root Activity"?
- The title input field should not be on a white background anymore.

smadbe (changes requested):
I don't know if it is supposed to be fixed but the font size looks still larger than the other places for regular content. Please check carefully.

GabrielVidal1:
I fixed the font size discrepancies

smadbe:
Screenshot 2021-02-22 at 10 26 27
Still not valid :-( please please, test it completely before asking for review.

GabrielVidal1:
I don't exactly know what size the font should be for each part of the component (my only reference is the mockup in the [commented design, the activity selected page](https://france-ioi.github.io/algorea-designs/01.Groups_Edit_02.Create-(Activity-selected)))

To be sure, should the activity name the same size as the fields name in the group information section? (same question for the text of the remove button)
This is what it will look like (with the same font size for the fields name, the activity name and the remove button):
image

src/app/modules/group/components/associated-activity/associated-activity.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  ]
+})
+export class AssociatedActivityComponent implements OnDestroy, ControlValueAccessor {
+
+  rootActivityId: string|null = null;
+  rootActivityName: string|null = null;
+
+  state: 'fetching'|'ready'|'error' = 'fetching';
+
+  private activityFetching = new Subject<string|null>();
```
smadbe, line 29: To make it more readable, I would alias an `ActivityId` to string.

src/app/modules/group/components/associated-activity/associated-activity.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  templateUrl: './associated-activity.component.html',
+  styleUrls: [ './associated-activity.component.scss' ],
+  providers: [
+    {
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: forwardRef(() => AssociatedActivityComponent),
+      multi: true,
+    }
+  ]
+})
```
smadbe, line 21: Great I didn't know that there was this alternative for handling form component. Actually it may apply to many things done by Amaury.

src/app/modules/group/components/associated-activity/associated-activity.component.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  onClickActivity(): void {
+    if (this.rootActivityId === null) return;
+
+    void this.router.navigate([ 'items', 'by-id', this.rootActivityId ]);
+  }
+
+  onRemove(): void {
+    if (this.rootActivityId === null) return;
```
smadbe, line 76: Probably you should check the state as well. `this.rootActivityId` could have a value which is not reflecting at all the current situation is the state is not ready. I know the button should not be there if not in ready state, but with race condition of UI, I am not sure we are fully guaranteed of this, better safe than sorry.

src/app/modules/group/components/associated-activity/associated-activity.component.ts
```diff
@@ -0,0 +1,82 @@
...
+    this.onChange = fn;
+  }
+
+  registerOnTouched(_fn: () => void): void {
+  }
+
+  onClickActivity(): void {
+    if (this.rootActivityId === null) return;
+
+    void this.router.navigate([ 'items', 'by-id', this.rootActivityId ]);
```
smadbe, line 72: Whenever possible, I prefer we handle the routing in the template, and also remove item-specific path (that may change) from specific components.
So, instead of having the router, create a function in `item-route.ts` such as `incompleteItemStringUrl`(I see that some fct in ` ItemRouter` should probably be in `item-route.ts`) with an appropriate short spec on top of the function.

smadbe (changes requested):
Very good, great way to handle form subcomponents ! I still need to discuss this a bit with you.
Sorry again for having been unclear for the font sizes. When it is not clear, ask on slack.

src/app/modules/group/components/associated-activity/associated-activity.component.ts
smadbe, line 29: Isn't it a more appropriate name than "abcFetching" for that? 🤔

src/app/modules/group/components/associated-activity/associated-activity.component.ts
```diff
@@ -0,0 +1,82 @@
...
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: forwardRef(() => AssociatedActivityComponent),
+      multi: true,
+    }
+  ]
+})
+export class AssociatedActivityComponent implements OnDestroy, ControlValueAccessor {
+
+  rootActivityId: string|null = null;
+  rootActivityName: string|null = null;
```
smadbe, line 25: To express that these are values which are the result of the fetching in ready state (and so, not corresponding to what it is currently get fetched for instance), I would keep them in a single variable (interface) (before we do better).

GabrielVidal1: fixed

src/app/modules/group/components/associated-activity/associated-activity.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/associated-activity/associated-activity.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/associated-activity/associated-activity.component.ts
GabrielVidal1: fixed

GabrielVidal1: I renamed it `activityData`, but I'm not sure if this is explicit enough...

src/app/modules/group/components/associated-activity/associated-activity.component.ts
```diff
@@ -73,10 +79,9 @@ export class AssociatedActivityComponent implements OnDestroy, ControlValueAcces
   }
 
   onRemove(): void {
-    if (this.rootActivityId === null) return;
+    if (this.rootActivity === null || this.state !== 'ready') return;
```
smadbe, line 77: As it is unexpected, throw an error instead. I was not doing that before but I try to throw errors on unexpected cases so that we can have these unexpected cases  in logs.

src/app/modules/group/components/associated-activity/associated-activity.component.ts
```diff
@@ -73,10 +79,9 @@ export class AssociatedActivityComponent implements OnDestroy, ControlValueAcces
...
 
-    this.rootActivityId = null;
-    this.rootActivityName = null;
-    this.onChange(this.rootActivityId);
+    this.rootActivity = null;
+    this.onChange(null);
```
smadbe, line 80: on remove, aren't you supposed to publish "null" on your subject instead?

smadbe: and is it normal you call "onChange" only in onRemove?

src/app/shared/helpers/item-route.ts
```diff
@@ -33,6 +33,13 @@ export function appDefaultItemRoute(cat: ItemTypeCategory = 'activity'): ItemRou
   };
 }
 
+/**
+ * Url (as string) of the item route with attemptId or path (only item id)
```
smadbe, line 37: "with" -> "without"?

src/app/modules/group/components/associated-activity/associated-activity.component.ts
smadbe: I would expect "data" for the result, not the subject. A subject is all but data :-)
"activityChanges"?

smadbe (changes requested):
👍  Few more questions/remarks

src/app/modules/group/components/associated-activity/associated-activity.component.ts
GabrielVidal1: > on remove, aren't you supposed to publish "null" on your subject instead?

The subject is only here for the ui, only to fetch the activity name

> and is it normal you call "onChange" only in onRemove?

For now the root activity can only be removed from the group, so the only place where the onChange function is called is in the onRemove function

src/app/shared/helpers/item-route.ts
GabrielVidal1: fixed

src/app/modules/group/components/associated-activity/associated-activity.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/associated-activity/associated-activity.component.ts
GabrielVidal1: fixed

src/app/modules/group/components/associated-activity/associated-activity.component.ts
smadbe: > > on remove, aren't you supposed to publish "null" on your subject instead?
> 
> The subject is only here for the ui, only to fetch the activity name

But couldn't you have a race condition problem ? 
1. the item A is loaded (through activityChanged subject), loading done
1. you click on "remove"... it is very slow
1. the item B is loaded, (through activityChanged subject), loading done
2. The response of the remove (2) arrrives... the activity is cleared out

?

GabrielVidal1: I did consider this problem and thought because the only case that was handled for now was the remove (which doesn't do network calls, so it's pretty instantaneous) I could directly call the onChange function when the user clicks on the remove button.

I refactored the code, so the onChange is called in the subscribe of the subject now.

smadbe: why shouldn't the normal case trigger onChange?

smadbe (approved):

PR #318, opened 2021-02-24
Left nav rework

smadbe:
Quite big PR which transforms the left menu so that group are presented in the same menu and in the same way as the items.
The current version is functional but not final at all:
- some groups do not load properly, require changes in the backend
- the "path" support of groups is still partial, there is no breadcrumb as well
- some behavior of the left menu still have to be improved

However, I think that is worth merging it in master so that the PR does not become too big. 
No need to review every detail, just to check that you understand the whole idea and you do not see clear errors (note that some may be fixed in further PR instead of this one).


src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -0,0 +1,181 @@
...
+      </span>
+      <span class="node-item-content" [ngClass]="{ unlocked: !node.locked }" (click)="selectNode(node)">
+        <span class="node-select">
+          <i class="fa" [ngClass]="{ 'fa-folder': !node.expanded, 'fa-folder-open': node.expanded }" style="font-size: 1.2rem;"></i>
+        </span>
+        <span class="node-label">
+          <span class="label-container">
+            <span class="node-label-title" [ngClass]="{ locked: node.locked }">
+              {{ node.label }}
+              <span *ngIf="node.data.element.groupName">(via group "{{ node.data.element.groupName }}")</span>
```
GabrielVidal1, line 79: missing localization

src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -0,0 +1,181 @@
...
+            </span>
+            <span class="node-state" *ngIf="node.data.status === 'loading'">
+              <p-progressSpinner [style]="{ width: '3rem', height: '3rem', 'margin-top': '-0.5rem' }"></p-progressSpinner>
+            </span>
+            <span class="node-state" *ngIf="node.data.status === 'error'">
+              <i class="fa fa-exclamation-triangle" style="font-size: 1.2rem;"></i>
```
GabrielVidal1, line 52: maybe move style in scss file because there is quite a few repetition in this file (for the p-progressSpinner and icons)

GabrielVidal1 (changes requested):

src/app/core/components/left-nav-tree/left-nav-tree.component.html
smadbe: actually for the progress spinner, it is not a css style, it is a string property given to the `p-progressSpinner` component.

src/app/core/components/left-nav-tree/left-nav-tree.component.html
smadbe: fixed

src/app/core/components/left-nav-tree/left-nav-tree.component.html
smadbe: fixed

GabrielVidal1 (approved):

PR #320, opened 2021-03-01
Activity/skill route

smadbe:
Replace `/items/...` paths by `/skills/...` and `/activities/...`. This way the menu does not have to wait for the first fetch to select the element.


smadbe:
@GabrielVidal1  Ignore the 2 last commits in your review as there are related with another PR that you approved.

GabrielVidal1 (approved):

PR #339, opened 2021-03-25
Reorg top nav3

smadbe:
Various improvement to the elements in the nav bars: redesign the lang picker, login/logout and user info on the top right.

### DIFF
--- a/src/app/core/components/language-picker/language-picker.component.scss
+++ b/src/app/core/components/language-picker/language-picker.component.scss
@@ -1,0 +1,35 @@
+@import "src/colors.scss";
+
+:host ::ng-deep .p-dropdown {
+  background-color: transparent;
+  border: none;
+  width: 4.1rem;
+
+  &:hover {
+    background-color: $hover-mask;
+  }
+
+  .p-dropdown-label {
+    text-transform: uppercase;
+    color: white;
+  }
+
+  .p-dropdown-panel {
+    background-color: $nav-bk-color;
+
+    .p-highlight {
+      background-color: $base-color;
+    }
+  }
+
+  .p-dropdown-item {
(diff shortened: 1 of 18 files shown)


src/app/core/components/language-picker/language-picker.component.scss
```diff
@@ -0,0 +1,35 @@
...
+  width: 4.1rem;
+
+  &:hover {
+    background-color: $hover-mask;
+  }
+
+  .p-dropdown-label {
+    text-transform: uppercase;
+    color: white;
+  }
```
GabrielVidal1, line 15: Only one letter is visible in the dropdown menu label, the two letters could be displayed by preventing the text ellipsis

image

GabrielVidal1 (changes requested):
When connected as the usual user, clicking on 'Log out' does nothing and clicking on 'Profile' loads the profile page but throws an error (in the console)

src/app/core/components/language-picker/language-picker.component.scss
smadbe: I always have the 2 letters on my browser. Good idea, I'll prevent ellipses.

smadbe:
> When connected as the usual user, clicking on 'Log out' does nothing and clicking on 'Profile' loads the profile page but throws an error (in the console)

For the log out not doing anything, that's because you have only changed the token, you should change the user_type to `authenticated` as well. (we could make a better use of the user_type vs user.is_temp, but that's more a improvement).

For the console error, that's something related with the `groups/me` page with the recently-changed menu, not related with this PR actually. I'll fix that in a separated PR.

GabrielVidal1 (approved):

PR #345, opened 2021-03-30
Log view experiment

smadbe:
Adapt state and introduce a new simple Rx operator which maps service to a state.
Apply it to log page.
(`tag` could be removed from states if we use it like that (require to update templates where it was used))

This change:
- simplifies/cleans a lot the component (shorter, straight to the point of what it does)
- simplifies the template handling as using for instance "datasource.isReady" makes the template know that `datasource.data` is set (no need to recheck that in addition), with the right type (no need to use $any() and to have use it as an "any") and that `isLoading` is false for instance...


GabrielVidal1 (approved):

GabrielVidal1:
Very nice way of doing things, I can see a lot of places where this will simplify a lot of the code

PR #347, opened 2021-04-02
Log view experiment 2

smadbe:
Various improvements based on the state changes of the previous PR #345


src/app/shared/operators/state.ts
```diff
@@ -1,11 +1,33 @@
...
+  return pipe(
+    filter((s): s is Ready<T> => s.isReady),
+    map(s => s.data)
+  );
+}
+
+/**
+ * Rx operator which maps the ready data using the `dataMapper` function or keep the state unchanged if the state is not ready
+ */
+export function mapStateData<T, U>(dataMapper: (data: T) => U): OperatorFunction<FetchState<T>,FetchState<U>> {
```
GabrielVidal1, line 29: I think the name of this function can be clearer, something like mapReadyStateData

GabrielVidal1 (approved):

src/app/shared/operators/state.ts
smadbe: Yes I thought about it but "ready" is the only state with data and everything related to the fetch state is for obtaining this data.

PR #351, opened 2021-04-06
My groups in left menu

smadbe:
Add "my groups" at the root of the "groups" tab in the left nav. (page was not accessible anymore)


GabrielVidal1 (approved):
Just a question: What will happen to the two categories "Groups you manage" and "Groups you joined"? Will they both be in the "My groups" tree? Or will there be 2 subtrees under "My groups"?

smadbe:
> Just a question: What will happen to the two categories "Groups you manage" and "Groups you joined"? Will they both be in the "My groups" tree? Or will there be 2 subtrees under "My groups"?

They will be merged in a same "my groups" page.

PR #354, opened 2021-04-16
Items decoding

### LINKED ISSUE #353: Sub / parent skills greyed out
As a temp user, if I visit http://dev.algorea.org/en/#/skills/by-id/3000;attempId=0;path=/details or http://dev.algorea.org/en/#/skills/by-id/3001;attempId=0;path=3000/details, the sub-/parent-skills are grey out... so with link disabled, for no reason I think.

GabrielVidal1:
Fixes #353 

After the PR #350, some consistency was lost in the naming of fiels (snake_case or CamelCase) which introduced bugs in other parts of the code, for example the `ItemPermissionsInfo` interface.
This broke the `ParentSkillsComponent` and `SubSkillsComponent` so I added decoding to the `get-item-children` and `get-item-parents` services to fix it.


smadbe (approved):
Thx 👍 
It is a pity that it did not fail properly when doing the initial change.

PR #358, opened 2021-04-21
Fixes style for cancel changes button on settings page

### LINKED ISSUE #356: Invert the normal and hover style of "cancel changes" button
In forms, when there is a change, the bottom right button "Cancel changes": 
The hover and normal style have been inverted, it should be transparent by default and flashy on hover. Fix that.

Iloveall:
Fixes #356


smadbe (approved):
excellent 👍

PR #370, opened 2021-04-26
Group service upd

smadbe:
Update the group-by-id service as the backend has changed its specs (current_user_manager does not exist anymore).
I changed the service to the iots decoder. Probably I shouldn't as the change was a bit larger than expected.


src/app/modules/group/helpers/group-code.ts
```diff
@@ -23,7 +26,7 @@ export interface CodeAdditions {
 export function withCodeAdditions<T extends CodeInfo>(g: T): T & CodeAdditions {
   return Object.assign({}, g, {
```
GabrielVidal1, line 27: Maybe use the spread operator (as it's done later in this file) 
```
{...g,  
 ...}
```

GabrielVidal1 (changes requested):

GabrielVidal1 (approved):

PR #371, opened 2021-04-26
Feature/personal page

### LINKED ISSUE #364: Personal page
## Motivations
Create a personal page as in the [design](https://france-ioi.github.io/algorea-designs/01.Groups_06/Personal-data) or [design branch (yourself in the left meny)](http://dev.algorea.org/branch/design)

## Subtasks
* The page will be accessible through the top right menu > "profile" (on `/groups/me`)
* For now, only display:
  * the title (firstname, lastname).
  * the content of the "personal data" tab. (without the tabs) (could already be in a dedicated component)
  * the info in the design and that you can find in the service `https://france-ioi.github.io/algorea-devdoc/api/#operation/userData` 
  * no "modify" buttons for now.
* No need to fetch the data, it has already been loaded by `UserSessionService`.
* While adding new attributes to CurrentUserHttpService, you can convert this service to decoders as done in `GetItemByIdService`.

Iloveall:
Fixes #364

### DIFF
--- a/src/app/modules/group/pages/current-user/current-user.component.html
+++ b/src/app/modules/group/pages/current-user/current-user.component.html
@@ -1,6 +1,6 @@
 <ng-container *ngIf="currentUser$ | async as currentUser">
   <div class="alg-page-header">
-    <div class="alg-page-header-title">{{ currentUser.firstName }} {{ currentUser.lastName }}</div>
+    <div class="alg-page-header-title --large">{{ currentUser.firstName }} {{ currentUser.lastName }}</div>
   </div>
 
   <div class="bg-white">
--- a/src/app/modules/group/pages/current-user/current-user.component.html
+++ b/src/app/modules/group/pages/current-user/current-user.component.html
@@ -1,1 +1,25 @@
 <div>Yourself page works</div>
+
+<div class="bg-white">
+  <alg-section i18n-label label="Personal Information" icon="fa fa-users">
+    {{ currentUser$ | async | json }}
+  </alg-section>
+</div>
+
+<div class="bg-white">
+  <alg-section i18n-label label="School Information" icon="fa fa-graduation-cap">
+    {{ currentUser$ | async | json }}
+  </alg-section>
+</div>
+
+<div class="bg-white">
+  <alg-section i18n-label label="Contact Information" icon="fa fa-envelope">
+    {{ currentUser$ | async | json }}
+  </alg-section>
+</div>
+
+<div class="bg-white">
+  <alg-section i18n-label label="Parameters" icon="fa fa-cog">
+    {{ currentUser$ | async | json }}
+  </alg-section>
+</div>
--- a/src/app/modules/group/pages/current-user/current-user.component.html
+++ b/src/app/modules/group/pages/current-user/current-user.component.html
@@ -64,22 +64,6 @@
           <span class="alg-data-list-item-caption" i18n>Primary email</span>
           <span class="alg-data-list-item-value">{{ currentUser.email }}</span>
         </li>
-        <li class="alg-data-list-item">
-          <span class="alg-data-list-item-caption" i18n>Address</span>
-          <span class="alg-data-list-item-value">{{ currentUser.address }}</span>
-        </li>
-        <li class="alg-data-list-item">
-          <span class="alg-data-list-item-caption" i18n>City</span>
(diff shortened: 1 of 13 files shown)


src/app/shared/http-services/current-user.service.ts
```diff
@@ -34,12 +50,11 @@ export class CurrentUserHttpService {
       .get<RawUserProfile>(`${appConfig().apiUrl}/current-user`)
       .pipe(
         map(raw => ({
+          ...raw,
```
smadbe, line 53: Check get-item-by-id and get-group-by-id. The get should be to an `unknown` type and then just decode.
Stick to the attribute names used in the backend, so use `groupId` and `tempUser` everywhere (may require to change other places of the code).

src/app/modules/group/pages/current-user/current-user.component.html
```diff
@@ -1 +1,81 @@
-<div>Yourself page works</div>
+<ng-container *ngIf="currentUser$ | async as currentUser">
+  <div class="alg-page-header">
+    <div class="alg-page-header-title">{{ currentUser.firstName }} {{ currentUser.lastName }}</div>
+  </div>
```
smadbe, line 4: Could you try to have the page title a bit larger than section title? A bit like [here](https://france-ioi.github.io/algorea-designs/01.Groups_06/Personal-data). Of course, we'll have to check that it is ok on the other pages as well, but that should make it.

src/app/modules/group/pages/current-user/current-user.component.ts
```diff
@@ -9,14 +11,20 @@ const currentUserBreadcrumbCat = $localize`Yourself`;
...
 
   constructor(
     private currentContent: CurrentContentService,
+    private currentUserHttpService: CurrentUserHttpService
   ) {
     this.currentContent.current.next(contentInfo({ breadcrumbs: { category: currentUserBreadcrumbCat, path: [], currentPageIdx: -1 } }));
   }
 
+  ngOnInit(): void {
+    this.currentUser$ = this.currentUserHttpService.getProfileInfo();
```
smadbe, line 25: As I wrote in the spec "No need to fetch the data, it has already been loaded by UserSessionService."

smadbe (changes requested):

src/app/shared/http-services/current-user.service.ts
Iloveall: Fixed

src/app/modules/group/pages/current-user/current-user.component.html
Iloveall: Fixed

src/app/modules/group/pages/current-user/current-user.component.ts
Iloveall: Fixed

src/app/modules/group/pages/current-user/current-user.component.html
```diff
@@ -1,6 +1,6 @@
 <ng-container *ngIf="currentUser$ | async as currentUser">
   <div class="alg-page-header">
-    <div class="alg-page-header-title">{{ currentUser.firstName }} {{ currentUser.lastName }}</div>
+    <div class="alg-page-header-title --large">{{ currentUser.firstName }} {{ currentUser.lastName }}</div>
```
smadbe, line 3: I meant: shouldn't all titles be "large"?

src/app/shared/helpers/input-is-not-null-or-undefined.ts
```diff
@@ -0,0 +1,3 @@
+export function inputIsNotNullOrUndefined<T>(input: null | undefined | T): input is T {
```
smadbe, line 1: 👍  let's call it `isNotNullNorUndefined` (update file name as well)

smadbe (changes requested):

src/app/shared/helpers/input-is-not-null-or-undefined.ts
Iloveall: Fixed

src/app/modules/group/pages/current-user/current-user.component.html
Iloveall: Fixed

smadbe (approved):
👍

PR #386, opened 2021-05-04
Feature/join group by code

### LINKED ISSUE #251: Join a group by code
## Motivations
Be able to join a group by entering code given by a teacher.
As in https://france-ioi.github.io/algorea-designs/01.Groups_02.Composition_k.Belong-to-teams or design branch, but not only for a team.

## Subtasks
* On the "My groups" page, below the group you joined lists, add a "Join a group by code" panel with a text field to enter the code and the "v" button. 
* The button is only enabled when at least one char is entered.
* When the button is pressed, the field+button are greyed out during processing.
* Use [groupsCodeCheck](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupsCodeCheck) to validate the code and display the appropriate error if any (probably reuse part of "access-code-view" component, move common parts in the "shared" module)
* If the code is valid, ask for confirmation (as in "access-code-view" component,) and use [groupsJoinByCode](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupsJoinByCode) to join.
* If the code is invalid, display the toast (or form validation)
* On success, display the toast and navigate to the group. 
* On failure, display the usual error toast.

Iloveall:
Fixes #251


smadbe (changes requested):
Code looks good. 
Two behaviors which need to be fixed:
- The spec is "below the group you joined lists", so before "groups you manage"
- after a successful joining, the "group you joined" table needs to be refreshed.

smadbe (approved):

PR #393, opened 2021-05-10
alg-input-error: add support for min and max number validators

ThomasJuster:


smadbe (approved):
looks good even if it's difficult to test it without a functionality using it.

PR #394, opened 2021-05-10
item-edit: add team controls

### LINKED ISSUE #256: Item settings: Teams (part 4)
## Motivations
Allow modifying most of the item parameters through interface (take the [design branch](http://dev.algorea.org/branch/design/) as an example)

## Subtasks
In item edit, in an "Advanced parameters" tab: (all two columns as in "Settings" in the design branch)
* "Teams" section (all types but skills)
  - entry_participant_type: "Participation as a team (only) (not implemented)" switch true/false (set to "Team" if yes, "User if no)   [only display the following fields if true]
  - entry_frozen_teams: "Freeze team on entry", switch true/false
  - entry_max_team_size: "Maximum team size": number >0
  - entry_min_admitted_members_ratio: "Minimum admitted members": horizontal selector among "All" "Half" "One" "None"

ThomasJuster:
Fixes #256

### DIFF
--- a/src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.html
+++ b/src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.html
@@ -166,7 +166,7 @@
       <div class="form-item-view">
         <span class="form-item-label" i18n>Maximum team size</span>
         <div class="form-item-control">
-          <p-inputNumber formControlName="entry_max_team_size"></p-inputNumber>
+          <p-inputNumber class="alg-input-number" formControlName="entry_max_team_size"></p-inputNumber>
           <alg-input-error [form]="this.parentForm" inputName="entry_max_team_size"></alg-input-error>
         </div>
       </div>
--- a/src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.html
+++ b/src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.html
@@ -156,29 +156,31 @@
         <alg-switch formControlName="entry_participant_type" mode="white"></alg-switch>
       </div>
     </div>
-    <div class="form-item-view">
-      <span class="form-item-label" i18n>Freeze team on entry</span>
-      <div class="form-item-control">
-        <alg-switch formControlName="entry_frozen_teams" mode="white"></alg-switch>
+    <ng-container *ngIf="!!parentForm.get('entry_participant_type')?.value">
+      <div class="form-item-view">
+        <span class="form-item-label" i18n>Freeze team on entry</span>
+        <div class="form-item-control">
+          <alg-switch formControlName="entry_frozen_teams" mode="white"></alg-switch>
+        </div>
       </div>
-    </div>
-    <div class="form-item-view">
-      <span class="form-item-label" i18n>Maximum team size</span>
-      <div class="form-item-control">
-        <p-inputNumber formControlName="entry_max_team_size"></p-inputNumber>
-        <alg-input-error [form]="this.parentForm" inputName="entry_max_team_size"></alg-input-error>
+      <div class="form-item-view">
+        <span class="form-item-label" i18n>Maximum team size</span>
+        <div class="form-item-control">
+          <p-inputNumber formControlName="entry_max_team_size"></p-inputNumber>
+          <alg-input-error [form]="this.parentForm" inputName="entry_max_team_size"></alg-input-error>
+        </div>
       </div>
-    </div>
-    <div class="form-item-view">
-      <span class="form-item-label" i18n>Minimum admitted members</span>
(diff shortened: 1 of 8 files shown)


src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -136,7 +144,13 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
...
         enteringTimeMaxEnabled: this.itemForm.get('entering_time_max_enabled'),
         enteringTimeMax: this.itemForm.get('entering_time_max'),
-      } : {})
+      } : {}),
+      ...(this.enableTeamSection ? {
+        entryParticipantType: this.itemForm.get('entry_participant_type'),
+        entryFrozenTeams: this.itemForm.get('entry_frozen_teams'),
+        entryMaxTeamSize: this.itemForm.get('entry_max_team_size'),
+        entryMinAdmittedMembersRatio: this.itemForm.get('entry_min_admitted_members_ratio'),
+      } : {}),
```
ThomasJuster, line 153: There are a lot of repeated transformations from snake case to lower case when using objects from the api, is there a dedicated place to declare such model mappers/transformers ?

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -45,6 +45,10 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
     entering_time_min: [ null ],
     entering_time_max_enabled: [ false ],
     entering_time_max: [ null ],
+    entry_participant_type: [ false ],
+    entry_frozen_teams: [ false ],
+    entry_max_team_size: [ '', Validators.min(1) ],
+    entry_min_admitted_members_ratio: [ '' ],
```
ThomasJuster, line 51: The form controls are named using snake case (probably to follow the backend model object), while the case convention for variables appears to be camel case. For further form developments, what casing should I use ?

src/app/modules/item/pages/item-edit/item-edit.component.ts
smadbe: There is mapper for incoming data but probably this form was written before. 
here are so many things to be improved for these form (not involving many pages of the platform hopefully)... but we'll start from the non-type safety with is pretty dangerous and error prone, but the case might be fixed with the type fix.

src/app/modules/item/pages/item-edit/item-edit.component.ts
smadbe: In particular, here, I see the the GET has been converted, but not the form nor the UPDATE.

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.html
```diff
@@ -148,4 +148,37 @@
       </div>
     </ng-container>
   </alg-section>
+
+  <alg-section icon="fa fa-users" i18n-label label="Team" *ngIf="enableTeamSection">
```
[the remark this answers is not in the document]
ThomasJuster: I tried to remain consistent, but I find the management of enableParticipation to be awkward. To me either a getter is written to declare the logic at one place to be dry, either the check is inlined everywhere which in my opinion is not ideal because then if the condition changes you have to replicate the change everywhere it's inlined.
That was for the same reason - DRY - that I put it as as an arg to its child, I thought that the same check should be used everywhere instead of replicating the logic in the child component.

As for the performance issue due to the change detection system of Angular, that part I know way less. It's strange to me that it would cause more computation. The getter will be run every time it is called yes, but the inline statements will also be run everywhere they are written. Maybe I'm missing something, I don't know… Angular caches in some way the ngIf statements ?

smadbe (changes requested):
I disable "Participation as a team (only) (not implemented)", save. After the refresh, the switch is back "on".

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
```diff
@@ -43,6 +43,20 @@ export class ItemEditAdvancedParametersComponent implements OnInit {
 
   minEnteringTimeMaxDate = new Date();
 
+  minAdmittedMembersRatioOptions: DropdownOption[] = [{
+    label: $localize`All the members must be admitted`,
+    value: 'All'
+  }, {
+    label: $localize`Half the members must be admitted`,
```
smadbe, line 50: Half **of**  the members ...

smadbe (approved):
👍 
To merge after this last detail.

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
ThomasJuster: Fixed

PR #411, opened 2021-05-13
Improves observation button

### LINKED ISSUE #398: Observation: toggle button
## Motivations
Improve the "observation" mode on groups

## Subtasks
* When visiting a group page, if the current group is the currently watched group (i.e. if `session$` in the `user-session` service has a `watchedGroup` which has the same id as the group of the current page), highlight the observation icon (maybe by using `$secondary-color` background when highlighted and white background (but border) when not-highlighted ? Screenshot 2021-05-11 at 15 29 29 Screenshot 2021-05-11 at 15 26 10 (or anything better) + use the highlight style on hover? 
* When clicking on the eye when the group is already been observed, disable observation.

Iloveall:
Fixes #398


src/app/modules/shared-components/components/page-navigator/page-navigator.component.ts
```diff
@@ -10,7 +13,19 @@ export class PageNavigatorComponent {
...
   @Output() edit = new EventEmitter<void>();
   @Output() watch = new EventEmitter<void>();
+  @Output() stopWatch = new EventEmitter<void>();
+
+  isCurrentGroupWatched$ = this.userSessionService.session$.pipe(
+    filter(isNotNullOrUndefined),
+    map(data => data.watchedGroup),
+    filter(isNotNullOrUndefined),
+    map(watchedGroup => this.groupId === watchedGroup.id)
+  );
```
smadbe, line 27: That is not correct, this way you'll not go back to false you'll have disable watching (`data.watchedGroup` back to undefined). There should be probably only one map and nothing else as pipe.

smadbe (changes requested):

src/app/modules/shared-components/components/page-navigator/page-navigator.component.ts
Iloveall: Yes, i left it because component does re init after start watch or navigation. But it could be forgotten in future if we'll do some refactoring, so I'll fix it

Iloveall: Fixed

src/app/modules/shared-components/components/page-navigator/page-navigator.component.ts
```diff
@@ -21,9 +21,7 @@ export class PageNavigatorComponent {
 
   isCurrentGroupWatched$ = this.userSessionService.session$.pipe(
     filter(isNotNullOrUndefined),
-    map(data => data.watchedGroup),
-    filter(isNotNullOrUndefined),
-    map(watchedGroup => this.groupId === watchedGroup.id)
+    map(data => data?.watchedGroup?.id === this.groupId)
```
smadbe, line 24: If data cannot be null (with your previous line), there is no reason for the first `?` I assume.
And a check that both are not undefined would be cleaner.

So I would expect something like
```
map(data => this.groupId !== undefined && data.watchedGroup?.id === this.groupId)
```

smadbe (changes requested):

src/app/modules/shared-components/components/page-navigator/page-navigator.component.ts
Iloveall: Fixed

smadbe (approved):

PR #417, opened 2021-05-17
Improves change tab

### LINKED ISSUE #412: Bug on dark group menu
* I visit any chapter
* I click on the "groups" tab in the left nav
* I visit a children of my chapter
* The menu switches to "activities" tab but stay black !  

In the review I've not checked the trigger for switching theme but clearly `onSelectionChangedByIdx` is not enough, it may also change on lines 55-69. 
Probably you should write a function `changeTab(tabIdx: number)` which does the tab change, theme notif and focus, called at 5 places where it changes.

Iloveall:
Fixes #412

### DIFF
--- a/src/app/core/components/left-nav/left-nav.component.ts
+++ b/src/app/core/components/left-nav/left-nav.component.ts
@@ -81,15 +81,11 @@ export class LeftNavComponent implements OnInit, OnDestroy {
   changeTab(index: number): void {
     this.activeTabIndex = index;
     this.dataSources[index].focus();
-    this.switchThemeColor();
+    this.themeChange.emit(this.activeTabIndex === 2 ? 'dark' : null);
   }
 
   retryError(): void {
     this.dataSources[this.activeTabIndex].retry();
-  }
-
-  switchThemeColor(): void {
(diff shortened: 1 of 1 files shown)


src/app/core/components/left-nav/left-nav.component.ts
```diff
@@ -75,8 +75,12 @@ export class LeftNavComponent implements OnInit, OnDestroy {
...
   onSelectionChangedByIdx(e: { index: number }): void {
-    this.activeTabIndex = e.index;
-    this.dataSources[e.index].focus();
+    this.changeTab(e.index);
+  }
+
+  changeTab(index: number): void {
+    this.activeTabIndex = index;
+    this.dataSources[index].focus();
     this.switchThemeColor();
```
smadbe, line 84: `switchThemeColor` does not really switch theme color.
Put the `this.themeChange.emit(this.activeTabIndex === 2 ? 'dark' : null);` line directly here, it will be cleaner.

smadbe (approved):

src/app/core/components/left-nav/left-nav.component.ts
Iloveall: Fixed

PR #421, opened 2021-05-18
implement language mismatch dialog

### LINKED ISSUE #390: Handle current language not matching user's
## Motivations
When the user is logged in, the language picker is not displayed on the top right anymore but can be changed in the profile (#389). In our case, the different languages are implemented through independent websites on different path. So if the user arrives on the platform on the wrong language (for instance, through a search engine), we should help him choose the right language.
When a language does not match his own default language, let him choose between changing his default language and navigate to his default language.

## Subtasks
* Make the localeService follow (subscribe to) user session and each time the user changes, check his locale as follow.
* If current locale matches user's default language, do nothing.
* Otherwise publish a state (a subject) which can be subscribed by the app component, which will display a non-closable modal on the UI as follow:
* If the user's default language is among the platform languages, display in the modal "The current language (`platform_language`) does not correspond to your account's default language (`user_default_language`). Do you want to continue in this language?" with 2 buttons "Continue in `platform_language` and change my default language to `platform_language` for this platform."  and "Visit the platform in `user_default_language`."
* If the user's default language is NOT among the platform languages, display "The platform does not exist in your default language `user_default_language`". Add one button  "Continue in `platform_language` and change my default language to `platform_language` for this platform." 
* In both case, add a comment in small after the button "The language can always be modified through your profile page." 
* On click "Continue in...", call the service to [update the  profile ](https://france-ioi.github.io/algorea-devdoc/api/#operation/userDataUpdate) (with a spinner while processing) and once done, close the modal .
* On click on "Visit the platform in `user_default_language`.", redirect to the same page in the user's default language (use the localeservice for that)

ThomasJuster:
Fixes #390


src/app/core/services/localeService.ts
```diff
@@ -1,24 +1,56 @@
-import { Injectable } from '@angular/core';
+import { Injectable, OnDestroy } from '@angular/core';
+import { BehaviorSubject, Subscription } from 'rxjs';
 import { appConfig } from 'src/app/shared/helpers/config';
+import { UserSessionService } from 'src/app/shared/services/user-session.service';
+
+export interface LanguageMismatch {
+  platformLanguage: string | undefined,
```
smadbe, line 7: A platform not having a current language does not really make sense. I realize that comes from currentTag which is optional. We should "fail-fast" and do something with this info directly instead of propagating this "should not happen" everywhere. 
Actually it is also maybe the case of `languages`, I do not remember the point of allowing it not to be defined (including in config).

src/app/core/components/language-mismatch/language-mismatch.component.html
```diff
@@ -0,0 +1,61 @@
+<p-dialog
+  *ngrxLet="languageMismatch$; let languageMismatch"
+  i18n-header header="Language mismatch"
+  [visible]="!!languageMismatch"
```
smadbe, line 4: I may be wrong but I would have made the component  loaded only if `!!languageMismatch`, so using
```
*ngIf="languageMismatch$|async as languageMismatch" 
[visible]="true"
```
Doesn't it work?

src/app/core/components/language-mismatch/language-mismatch.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(
+    private localeService: LocaleService,
+    private currentUserService: CurrentUserHttpService,
+  ) { }
+
+  onUpdateUserLanguage(language: string | undefined): void {
+    if (!language) return;
+    this.currentUserService.update({ default_language: language })
+      .pipe(mapToFetchState())
```
smadbe, line 25: `mapToFetchState` is made for fetching, not for action... probably does not fit the need here as it does not have the same state cycle as an action.

Wow wow who have written `  update(changes: object): Observable<any> `? I missed it during the review. 😣 There should not be any "any" nor "object" in the project.
Probably I could solve it separately.

src/app/core/components/language-mismatch/language-mismatch.component.ts
```diff
@@ -0,0 +1,40 @@
...
+      .subscribe({
+        next: state => {
+          this.updateUserState = state;
+        },
+        complete: () => {
+          this.languageMismatch$.next(undefined);
```
smadbe, line 31: Why separating next and complete for a service? It is a bad smell :-)

smadbe (changes requested):
I have not reviewed it completely but I have already many important remarks, so I suggest you already fix these.
Maybe we could discuss first, as they are many thing which are not directly related with this issue, we could fix each of them separately before coming back to this PR.

src/app/core/services/localeService.ts
```diff
@@ -1,24 +1,56 @@
...
+      const hasLanguageMismatch = !!this.currentTag && session.user.defaultLanguage !== this.currentTag;
+      if (hasLanguageMismatch) {
+        this.languageMismatch$.next({
+          platformLanguage: this.currentTag,
+          userDefaultLanguage,
+          userDefaultLanguageIsSupported: this.languages?.some(({ tag }) => tag === userDefaultLanguage) ?? false,
+        });
+      }
+    });
+  }
```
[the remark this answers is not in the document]
ThomasJuster: I can't do that for now because the user session is not refreshed causing the dialog to stay. Currently I can close the dialog by explicitly calling `languageMismatch$.next()`, which a smell too I guess.

Shouldn't I make the refresh task first and then come back to this one ?

src/app/core/services/localeService.ts
ThomasJuster: I agree, however the scope is starting to grow somehow, do you mind if I make intermediary PRs ?

PR #423, opened 2021-05-18
Auth forced token

smadbe:
* To ease development, add a way to force usage of a token (even when auth type is cookies).
* improve tests / code related to auth

### DIFF
--- a/src/app/shared/auth/auth-info.spec.ts
+++ b/src/app/shared/auth/auth-info.spec.ts
@@ -1,5 +1,13 @@
 import { SECONDS } from '../helpers/duration';
-import { clearTokenFromStorage, tokenAuthFromServiceResp, tokenAuthFromStorage } from './auth-info';
+import {
+  clearForcedTokenFromStorage,
+  clearTokenFromStorage,
+  forcedTokenAuthFromStorage,
+  hasForcedToken,
+  setForcedTokenInStorage,
+  tokenAuthFromServiceResp,
+  tokenAuthFromStorage
+} from './auth-info';
 
 describe('auth-info', () => {
 
@@ -9,6 +17,7 @@ describe('auth-info', () => {
 
   beforeEach(function() {
     clearTokenFromStorage();
+    clearForcedTokenFromStorage();
   });
 
   it('should save & load successfully a token', () => {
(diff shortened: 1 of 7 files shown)


src/app/shared/auth/auth-info.spec.ts
```diff
@@ -1,72 +1,86 @@
...
+    tokenAuthFromServiceResp(expectedToken, -1 /* expired 1sec ago */);
+    expect(tokenAuthFromStorage).toThrowError();
     expect(sessionStorage.getItem('access_token')).toBeNull();
     expect(sessionStorage.getItem('access_token_exp')).toBeNull();
   });
 
   it('should fail (and clear) when loading an invalid token', () => {
-    tokenAuthFromServiceResp('atemptok', -1 /* expired 1sec ago */);
+    tokenAuthFromServiceResp(expectedToken, -1 /* expired 1sec ago */);
     sessionStorage.setItem('access_token_exp', 'not a number');
```
ThomasJuster, line 60: Without this line, the test would still produce the same result as the test just upper, it might cause maintenance problems later.
If I'm getting it right, the idea would be to fail because access_token_exp is not a number without considering the expiration. If so, changing the expired value to a passing one will do.

src/app/shared/auth/auth-info.spec.ts
```diff
@@ -1,72 +1,86 @@
...
   });
 
   it('should fail (and clear) when loading a expired token', () => {
-    tokenAuthFromServiceResp('atemptok', -1 /* expired 1sec ago */);
-    const loaded = tokenAuthFromStorage();
-    expect(loaded).toBeUndefined();
+    tokenAuthFromServiceResp(expectedToken, -1 /* expired 1sec ago */);
+    expect(tokenAuthFromStorage).toThrowError();
     expect(sessionStorage.getItem('access_token')).toBeNull();
     expect(sessionStorage.getItem('access_token_exp')).toBeNull();
```
ThomasJuster, line 55: I didn't spot this one before, I think it's implementation detail and it makes the test more fragile. What do you think ?
I spotted others in the PR in case you think it should be changed.

ThomasJuster (changes requested):

src/app/shared/auth/auth-info.spec.ts
smadbe: 👍  I think that was error, indeed

ThomasJuster (approved):

PR #426, opened 2021-05-18
Refresh user session on profile update

### LINKED ISSUE #422: Refresh user session when profile data is updated
## Motivations

The user session should be refreshed  when updating user data.

## Subtasks

* In the current-user service, add a refresh subject to trigger on user update
* In the user-session service, observe the current user refresh subject to refetch user info

ThomasJuster:
Fixes #422

### DIFF
--- a/src/app/core/components/top-right-controls/top-right-controls.component.html
+++ b/src/app/core/components/top-right-controls/top-right-controls.component.html
@@ -1,9 +1,16 @@
 <div class="controls-zone" [ngClass]="drawLeftBorder ? 'with-left-border' : 'no-left-border'" *ngrxLet="session$ as session">
-  <alg-language-picker *ngIf="!session || (session && session.user.tempUser)"></alg-language-picker>
-  <alg-top-right-menu *ngIf="session && !session.user.tempUser; else notConnected"></alg-top-right-menu>
-  <ng-template #notConnected>
+
+  <ng-container
+    *ngIf="session && !session.user.tempUser; then authenticated else notAuthenticated"
+  ></ng-container>
+
+  <ng-template #authenticated>
+    <alg-top-right-menu></alg-top-right-menu>
+  </ng-template>
+
+  <ng-template #notAuthenticated>
+    <alg-language-picker></alg-language-picker>
     <p-button
-      *ngIf="!session || (session && session.user.tempUser)"
       i18n-label label="Log in / Sign in"
       icon="pi pi-power-off"
       iconPos="right"
@@ -11,5 +18,4 @@
       (click)="login()"
     ></p-button>
   </ng-template>
-
 </div>
--- a/src/app/core/components/top-right-controls/top-right-controls.component.ts
+++ b/src/app/core/components/top-right-controls/top-right-controls.component.ts
@@ -20,8 +20,4 @@ export class TopRightControlsComponent{
     this.sessionService.login();
   }
 
-  logout(): void {
-    this.sessionService.logout();
-  }
-
 }
--- a/src/app/core/components/top-right-menu/top-right-menu.component.ts
+++ b/src/app/core/components/top-right-menu/top-right-menu.component.ts
@@ -1,4 +1,6 @@
 import { Component } from '@angular/core';
+import { AuthService } from 'src/app/shared/auth/auth.service';
+import { appConfig } from 'src/app/shared/helpers/config';
 import { UserSessionService } from 'src/app/shared/services/user-session.service';
 
 @Component({
@@ -10,13 +12,22 @@ export class TopRightMenuComponent {
 
   readonly menuItems = [
     { label: 'Profile', icon: 'pi pi-user', routerLink: [ 'groups', 'me' ] },
+    ...(!appConfig.production
+      ? [{ label: 'Invalidate token', icon: 'pi pi-refresh', command: (): void => this.invalidateToken() }]
+      : []
+    ),
     { label: 'Log out', icon: 'pi pi-power-off', command: ():void => this.sessionService.logout() },
   ]
 
   sessions$ = this.sessionService.session$;
 
   constructor(
-    private sessionService: UserSessionService
+    private sessionService: UserSessionService,
+    private authService: AuthService,
   ) { }
 
+  private invalidateToken(): void {
+    const authStatus = this.authService.status$.value;
+    if (authStatus.authenticated) this.authService.invalidToken(authStatus);
+  }
 }
--- a/src/app/core/components/top-right-menu/top-right-menu.component.ts
+++ b/src/app/core/components/top-right-menu/top-right-menu.component.ts
@@ -1,4 +1,5 @@
 import { Component } from '@angular/core';
+import { MenuItem } from 'primeng/api';
 import { distinctUntilChanged, map } from 'rxjs/operators';
 import { AuthService } from 'src/app/shared/auth/auth.service';
 import { appConfig } from 'src/app/shared/helpers/config';
@@ -13,10 +14,7 @@ export class TopRightMenuComponent {
 
   readonly menuItems = [
     { label: 'Profile', icon: 'pi pi-user', routerLink: [ 'groups', 'me' ] },
-    ...(!appConfig.production
-      ? [{ label: 'Invalidate token', icon: 'pi pi-refresh', command: (): void => this.invalidateToken() }]
-      : []
-    ),
+    ...this.getDevelopmentMenuItems(),
     { label: 'Log out', icon: 'pi pi-power-off', command: ():void => this.sessionService.logout() },
   ]
 
@@ -31,4 +29,12 @@ export class TopRightMenuComponent {
     const authStatus = this.authService.status$.value;
     if (authStatus.authenticated) this.authService.invalidToken(authStatus);
   }
+
+  private getDevelopmentMenuItems(): MenuItem[] {
+    if (appConfig.production) return [];
+    return [
+      { label: 'Invalidate token', icon: 'pi pi-refresh', command: (): void => this.invalidateToken() },
+    ];
+  }
+
 }
--- a/src/app/core/components/top-right-menu/top-right-menu.component.ts
+++ b/src/app/core/components/top-right-menu/top-right-menu.component.ts
@@ -1,4 +1,5 @@
 import { Component } from '@angular/core';
+import { distinctUntilChanged, map } from 'rxjs/operators';
 import { AuthService } from 'src/app/shared/auth/auth.service';
 import { appConfig } from 'src/app/shared/helpers/config';
 import { UserSessionService } from 'src/app/shared/services/user-session.service';
@@ -19,7 +20,7 @@ export class TopRightMenuComponent {
     { label: 'Log out', icon: 'pi pi-power-off', command: ():void => this.sessionService.logout() },
   ]
 
-  sessions$ = this.sessionService.session$;
+  userLogin$ = this.sessionService.session$.pipe(map(session => session?.user.login), distinctUntilChanged())
 
   constructor(
     private sessionService: UserSessionService,
--- a/src/app/modules/item/pages/item-progress/item-progress.component.html
+++ b/src/app/modules/item/pages/item-progress/item-progress.component.html
@@ -1,5 +1,5 @@
-<ng-container *ngIf="watchedGroup$ | async as watchedGroup">
-  <alg-section icon="fa fa-list" i18n-label label="Group situation">
+<ng-container *ngIf="session$ | async as session">
+  <alg-section icon="fa fa-list" i18n-label label="{{session.watchedGroup ? 'Group' : 'User'}} situation">
     <alg-item-current-situation [itemData]="itemData"></alg-item-current-situation>
   </alg-section>
 </ng-container>
--- a/src/app/modules/item/pages/item-progress/item-progress.component.html
+++ b/src/app/modules/item/pages/item-progress/item-progress.component.html
@@ -1,5 +1,5 @@
-<ng-container *ngIf="session$ | async as session">
-  <alg-section icon="fa fa-list" i18n-label label="{{session.watchedGroup ? 'Group' : 'User'}} situation">
+<ng-container *ngIf="watchedGroup$ | async as watchedGroup">
+  <alg-section icon="fa fa-list" i18n-label label="Group situation">
     <alg-item-current-situation [itemData]="itemData"></alg-item-current-situation>
   </alg-section>
 </ng-container>
--- a/src/app/modules/item/services/item-datasource.service.ts
+++ b/src/app/modules/item/services/item-datasource.service.ts
@@ -52,7 +52,7 @@ export class ItemDataSource implements OnDestroy {
 
     ).subscribe(state => this.state.next(state));
 
-    this.subscription = this.userSessionService.session$.subscribe(_s => this.refreshItem());
+    this.subscription = this.userSessionService.userChanged$.subscribe(_s => this.refreshItem());
   }
 
   fetchItem(item: ItemRoute): void {
--- a/src/app/shared/services/user-session.service.ts
+++ b/src/app/shared/services/user-session.service.ts
@@ -78,7 +78,7 @@ export class UserSessionService implements OnDestroy {
   }
 
   updateCurrentUser(changes: UpdateUserBody): Observable<void> {
-    const update$ = this.currentUserService.update(changes).pipe(shareReplay());
+    const update$ = this.currentUserService.update(changes).pipe(shareReplay(1));
     update$.subscribe(() => this.userProfileUpdated$.next());
     return update$;
   }
--- a/src/app/shared/services/user-session.service.ts
+++ b/src/app/shared/services/user-session.service.ts
@@ -18,6 +18,7 @@ export interface UserSession {
 export class UserSessionService implements OnDestroy {
 
   session$ = new BehaviorSubject<UserSession|undefined>(undefined)
+  watchedGroup$ = this.session$.pipe(map(session => session?.watchedGroup), distinctUntilChanged())
 
   /** currently-connected user profile, temporary or not, excluding transient (undefined) states */
   user$ = this.session$.pipe(
--- a/src/app/shared/services/user-session.service.ts
+++ b/src/app/shared/services/user-session.service.ts
@@ -1,7 +1,7 @@
 import { Injectable, OnDestroy } from '@angular/core';
 import { EMPTY, BehaviorSubject, Subscription, Observable, Subject, of } from 'rxjs';
 import { AuthService } from '../auth/auth.service';
-import { switchMap, catchError, distinctUntilChanged, map, filter, mapTo, skip, delayWhen, share } from 'rxjs/operators';
+import { switchMap, catchError, distinctUntilChanged, map, filter, mapTo, skip, shareReplay } from 'rxjs/operators';
 import { CurrentUserHttpService, UpdateUserBody, UserProfile } from '../http-services/current-user.service';
 import { Group } from 'src/app/modules/group/http-services/get-group-by-id.service';
 import { isNotUndefined } from '../helpers/null-undefined-predicates';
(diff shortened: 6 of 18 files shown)


src/app/shared/services/user-session.service.ts
```diff
@@ -65,6 +66,14 @@ export class UserSessionService implements OnDestroy {
     return !session || session.user.tempUser;
   }
 
+  updateCurrentUser(changes: UpdateUserBody): Observable<void> {
+    return this.currentUserService.update(changes).pipe(
+      tap(() => {
+        this.refresh$.next();
+      }),
```
ThomasJuster, line 73: I'm not sure this is the best way of doing it

src/app/shared/services/user-session.service.ts
```diff
@@ -29,18 +30,18 @@ export class UserSessionService implements OnDestroy {
...
+      this.authService.status$.pipe(filter(auth => auth.authenticated)),
+      this.refresh$,
+    ).pipe(
+      switchMap(() => this.currentUserService.getProfileInfo().pipe(
+        catchError(_e => EMPTY)
+      )),
+      distinctUntilChanged((p1, p2) => p1 === p2 || (p1.groupId === p2.groupId))
     ).subscribe(profile => {
-      this.session$.next(profile ? { user: profile } : undefined);
+      this.session$.next({ user: profile });
```
smadbe, line 44: You completely change the behavior of session$ and so of the full app with that change as you remove the undefined from sessions...
It might be valid... or not... but you haven't changed all other parts of the app which depend on that so that's not correct.

src/app/shared/services/user-session.service.ts
```diff
@@ -29,18 +30,18 @@ export class UserSessionService implements OnDestroy {
...
-      }),
-      distinctUntilChanged((p1, p2) => p1 === p2 || (!!p1 && !!p2 && p1.groupId === p2.groupId))
+    this.subscription = merge(
+      this.authService.status$.pipe(filter(auth => auth.authenticated)),
+      this.refresh$,
+    ).pipe(
+      switchMap(() => this.currentUserService.getProfileInfo().pipe(
+        catchError(_e => EMPTY)
+      )),
+      distinctUntilChanged((p1, p2) => p1 === p2 || (p1.groupId === p2.groupId))
```
smadbe, line 42: Have you tested it ? Because thanks to the `distinctUntilChanged` there is no change in profile if it is the same user.

src/app/shared/services/user-session.service.ts
```diff
@@ -17,6 +17,7 @@ export interface UserSession {
 export class UserSessionService implements OnDestroy {
 
   session$ = new BehaviorSubject<UserSession|undefined>(undefined)
+  refresh$ = new Subject()
```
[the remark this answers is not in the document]
smadbe: To prevent any leak, you should complete the subject on destroy.
(and session$ as well, it has been forgotten)

src/app/shared/services/user-session.service.ts
```diff
@@ -29,18 +30,18 @@ export class UserSessionService implements OnDestroy {
...
-        if (!auth.authenticated) return of<UserProfile|undefined>(undefined);
-        return this.http.getProfileInfo().pipe(
-          catchError(_e => EMPTY)
-        );
-      }),
-      distinctUntilChanged((p1, p2) => p1 === p2 || (!!p1 && !!p2 && p1.groupId === p2.groupId))
+    this.subscription = merge(
+      this.authService.status$.pipe(filter(auth => auth.authenticated)),
+      this.refresh$,
+    ).pipe(
```
smadbe, line 38: Instead of the merge, I would have written:
```
this.authService.status$.pipe(
  repeatLatestWhen(this.refresh$),
  switchMap(...
```
(even if I admit the merge works here as you do not need the auth to get the profile)

src/app/shared/services/user-session.service.ts
```diff
@@ -29,18 +30,18 @@ export class UserSessionService implements OnDestroy {
...
+      switchMap(() => this.currentUserService.getProfileInfo().pipe(
+        catchError(_e => EMPTY)
```
smadbe, line 40: I know that was there already but it is strange.  What the best thing to do when it fails?

smadbe (changes requested):

src/app/shared/services/user-session.service.ts
ThomasJuster: Retry a few times, and if still not working warn through the feedback service. What would you say ?

src/app/shared/services/user-session.service.ts
ThomasJuster: I have tested it, indeed currently the profile won't change which is not what I want BUT
If I change the distinct until changed condition here, I'll have to inspect every subscription to `userSession.session$`, I intended to open another PR for that.

Side note: The check `p1 === p2` is useless since `{} !== {}` because javascript :shrug:

src/app/shared/services/user-session.service.ts
ThomasJuster: Right, and it probably breaks the logout behavior mostly

src/app/shared/services/user-session.service.ts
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: The feedback service UI is not initialized when the user session service constructor is executed, so it won't work as is.
I could still retry a few times, it's not very convincing to me right now.

src/app/shared/services/user-session.service.ts
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
smadbe: How would it help to warn the user? The problem is that afterwards, we would have, for the platform, the user authenticated but no user profile, an inconsistency which we would have to tackle at many places (and to do what?).
Probably we would have to fallback to the same error as the autherror. (it might just be because the user lost internet connection)

src/app/shared/services/user-session.service.ts
smadbe: If this misbehaving was intended, you should have mentioned in the PR description so that I know I have to ignore that issue.

smadbe: I probably know better than you each usage. Actually, most should be straightforward but 1 or 2 may be more critical.

src/app/shared/services/user-session.service.ts
ThomasJuster: Actually it worked, and it didn't change the behavior of session. `session$` relies on `authService.status$`, `session$` is undefined when `authService.status$` is `notAuthenticated()` and it can be undefined at 2 moments:
1. At start, because `auth.status$` is a Behavior subject with default value `notAuthenticated()`
2. At logout, but it won't change anything because a redirection is performed, thus leaving the app, and redirecting to the same app causing a re-initialisation, which creates a temporary user.

Beyond that, `auth.status$` will be of type `CookieAuthenticated | TokenAuthenticated`, which both have `authenticated: true`

One could argue that the the `session$` is not supposed to know that at logout a redirection is performed, but then the problem will be with the `userSession.user$` since it filters out `userSession.session$` undefined results. Currently, performing a logout will not update the `userSession.user$` subject. Plus, if the idea _was_ stay on the app at logout, a temporary user would need to be created anyway.

To me, since a temporary user is created when no existing user is logged, the user session should always be populated _except_ at app initialization, which is (not explicitly) the case currently.

src/app/shared/services/user-session.service.ts
ThomasJuster: I'll do the changes in that PR

src/app/core/components/top-right-controls/top-right-controls.component.html
```diff
@@ -1,15 +1,21 @@
...
+    <alg-language-picker></alg-language-picker>
     <p-button
-      *ngIf="!session || (session && session.user.tempUser)"
       i18n-label label="Log in / Sign in"
       icon="pi pi-power-off"
       iconPos="right"
       styleClass="login-button"
       (click)="login()"
     ></p-button>
   </ng-template>
```
ThomasJuster, line 20: This change is not in the scope of this PR, so I'll detail it shortly:
The same condition `!session || !session.user.tempUser` was repeated thrice, I gathered them in one and split the view into 2 template.

src/app/core/components/top-right-controls/top-right-controls.component.ts
```diff
@@ -20,8 +20,4 @@ export class TopRightControlsComponent{
     this.sessionService.login();
   }
 
-  logout(): void {
-    this.sessionService.logout();
-  }
-
```
ThomasJuster, line 26: Dead code

src/app/shared/services/user-session.service.ts
ThomasJuster: :warning: not fixed (yet)

src/app/shared/services/user-session.service.ts
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: I have another idea on how to perform the refresh, what do you think ↓ ?
```ts
const update$ = this.currentUserService.update(changes);

of(null).pipe(
  skip(1),
  repeatWhen(() => update$),
).subscribe(() => this.refresh$.next());

return update$;
```

ThomasJuster: I found another way to perform this `refresh$.next()`, what do you think ↓ ?
```ts
updateCurrentUser(changes: UpdateUserBody): Observable<void> {
  const update$ = this.currentUserService.update(changes).pipe(share());

  of(null).pipe(
    delayWhen(() => update$),
  ).subscribe(() => this.refresh$.next());

  return update$;
}
```

ThomasJuster: I also tested what happens when the caller of `updateCurrentUser()` delays the subscription and it works fine, the http request is not sent until the observable is subscribed to

ThomasJuster: Fixed

src/app/core/components/top-right-controls/top-right-controls.component.ts
smadbe: 👍

src/app/core/components/top-right-menu/top-right-menu.component.ts
```diff
@@ -10,13 +13,22 @@ export class TopRightMenuComponent {
...
     { label: 'Profile', icon: 'pi pi-user', routerLink: [ 'groups', 'me' ] },
+    ...(!appConfig.production
+      ? [{ label: 'Invalidate token', icon: 'pi pi-refresh', command: (): void => this.invalidateToken() }]
+      : []
+    ),
     { label: 'Log out', icon: 'pi pi-power-off', command: ():void => this.sessionService.logout() },
   ]
 
-  sessions$ = this.sessionService.session$;
+  userLogin$ = this.sessionService.session$.pipe(map(session => session?.user.login), distinctUntilChanged())
```
smadbe, line 23: so I assume it can follow `userProfile$` instead.

src/app/core/components/top-right-menu/top-right-menu.component.ts
```diff
@@ -10,13 +13,22 @@ export class TopRightMenuComponent {
 
   readonly menuItems = [
     { label: 'Profile', icon: 'pi pi-user', routerLink: [ 'groups', 'me' ] },
+    ...(!appConfig.production
+      ? [{ label: 'Invalidate token', icon: 'pi pi-refresh', command: (): void => this.invalidateToken() }]
+      : []
+    ),
```
smadbe, line 19: you can probably leave this in the code, just move this is a function not to pollute the readability of this list too much.

src/app/modules/item/pages/item-progress/item-progress.component.html
```diff
@@ -1,5 +1,5 @@
-<ng-container *ngIf="session$ | async as session">
-  <alg-section icon="fa fa-list" i18n-label label="{{session.watchedGroup ? 'Group' : 'User'}} situation">
+<ng-container *ngIf="watchedGroup$ | async as watchedGroup">
+  <alg-section icon="fa fa-list" i18n-label label="Group situation">
```
smadbe, line 2: So it never display "User situation" anymore?
Actually, these changes will have severe conflict with #427 that Aleksandr is working on.

src/app/modules/item/services/item-datasource.service.ts
```diff
@@ -52,7 +52,7 @@ export class ItemDataSource implements OnDestroy {
 
     ).subscribe(state => this.state.next(state));
 
-    this.subscription = this.userSessionService.session$.subscribe(_s => this.refreshItem());
+    this.subscription = this.userSessionService.userChanged$.subscribe(_s => this.refreshItem());
```
smadbe, line 55: I assume we could get rid of this subscription by using `repeatLatestWhen(this.userSessionService.userChanged$)` on the the previous operation... but that out of scope of this PR.

src/app/shared/services/user-session.service.ts
```diff
@@ -17,35 +18,46 @@ export interface UserSession {
 export class UserSessionService implements OnDestroy {
 
   session$ = new BehaviorSubject<UserSession|undefined>(undefined)
+  userProfileError$ = new Subject<void>()
+  watchedGroup$ = this.session$.pipe(map(session => session?.watchedGroup), distinctUntilChanged())
```
smadbe, line 22: `distinctUntilChanged` is probably not able to deal properly with a "Group". Maybe that's not very important if the only goal is to get rid of the multiple `undefined` though. (maybe should be written in comments at least)

src/app/shared/services/user-session.service.ts
```diff
@@ -17,35 +18,46 @@ export interface UserSession {
...
 
   /** currently-connected user profile, temporary or not, excluding transient (undefined) states */
-  user$ = this.session$.pipe(
+  userProfile$ = this.session$.pipe(
     filter(isNotUndefined),
     map(session => session.user),
-    distinctUntilChanged((u1, u2) => u1.groupId === u2.groupId)
   );
 
+  /** triggered when the user identity changes, which happens when auth token is invalidated */
```
smadbe, line 30: seems useful to mention that this does not emit an event when the user is first detected/fetched.

src/app/shared/services/user-session.service.ts
```diff
@@ -17,35 +18,46 @@ export interface UserSession {
...
+  userChanged$ = this.userProfile$.pipe(distinctUntilChanged((u1, u2) => u1.groupId === u2.groupId), mapTo(undefined), skip(1))
+
   private subscription?: Subscription;
+  private refresh$ = new Subject<void>()
```
smadbe, line 34: For a user-session service, "refresh$" is not clear enough (could be many things to refresh). 
`userProfileRefresh$`  ?

src/app/shared/services/user-session.service.ts
```diff
@@ -17,35 +18,46 @@ export interface UserSession {
...
     this.subscription = this.authService.status$.pipe(
+      repeatLatestWhen(this.refresh$),
       switchMap(auth => {
-        if (!auth.authenticated) return of<UserProfile|undefined>(undefined);
-        return this.http.getProfileInfo().pipe(
-          catchError(_e => EMPTY)
+        if (!auth.authenticated) return of<UserProfile | undefined>(undefined);
+        return this.currentUserService.getProfileInfo().pipe(
+          catchError(_e => {
+            this.userProfileError$.next();
```
smadbe, line 46: no side effects in pipes! 
Actually, this should be done in subscribe (error case).

src/app/shared/services/user-session.service.ts
```diff
@@ -17,35 +18,46 @@ export interface UserSession {
...
+    private currentUserService: CurrentUserHttpService,
   ) {
     this.subscription = this.authService.status$.pipe(
+      repeatLatestWhen(this.refresh$),
       switchMap(auth => {
-        if (!auth.authenticated) return of<UserProfile|undefined>(undefined);
-        return this.http.getProfileInfo().pipe(
-          catchError(_e => EMPTY)
+        if (!auth.authenticated) return of<UserProfile | undefined>(undefined);
+        return this.currentUserService.getProfileInfo().pipe(
```
smadbe, line 44: Add a `retry(1)` on that request.

src/app/shared/services/user-session.service.ts
```diff
@@ -65,6 +77,12 @@ export class UserSessionService implements OnDestroy {
     return !session || session.user.tempUser;
   }
 
+  updateCurrentUser(changes: UpdateUserBody): Observable<void> {
+    const update$ = this.currentUserService.update(changes).pipe(shareReplay());
```
smadbe, line 81: I think you have to mention the size of the buffer
```suggestion
    const update$ = this.currentUserService.update(changes).pipe(shareReplay(1));
```

src/app/core/components/top-right-menu/top-right-menu.component.ts
ThomasJuster: I think we can't because the user profile filters out the empty sessions. If the token is invalidated and no session is started for any reason (bad connection, etc.), the previous login would be displayed. Now I'm not sure whether this specific situation might happen, so I assumed it was a possibility.

src/app/core/components/top-right-menu/top-right-menu.component.ts
ThomasJuster: Fixed

src/app/modules/item/pages/item-progress/item-progress.component.html
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: Fixed

src/app/shared/services/user-session.service.ts
ThomasJuster: The buffer size is optional, but one could argue that providing the buffer size is more explicit for readability.

From a code point of view, I'm not sure it matters much. There's one value only in this observable, which will be repeated. If someday there are more values, they will all be repeated with current share replay config. From this point of view it looks fine to me.

I'll change it anyway to avoid starting a discussion that might be more entertaining than useful.
Fixed

src/app/core/components/top-right-menu/top-right-menu.component.ts
smadbe: It might happen but the UI will lock with the error, so that's not very problematic. Your point is correct as well.

src/app/shared/services/user-session.service.ts
```diff
@@ -44,16 +44,13 @@ export class UserSessionService implements OnDestroy {
...
-          })
-        );
+        return this.currentUserService.getProfileInfo().pipe(retry(1));
       }),
       distinctUntilChanged(), // skip two undefined values in a row
     ).subscribe(profile => {
       this.session$.next(profile ? { user: profile } : undefined);
+    }, () => {
+      this.userProfileError$.next();
     });
```
smadbe, line 54: `}, () => {` should not be at the same indent as `).subscribe(profile => {` as it is an arg of subscribe.
Usually I write subscription with error like that:
```
).subscribe(
  () => {...},
  () => {... },
)
```
Here, as it is one instructions it should even be: 
```
).subscribe(
  () => this.session$.next(profile ? { user: profile } : undefined),
  () => this.userProfileError$.next(),
)
```

smadbe (approved):

PR #436, opened 2021-05-25
Avoids adding duplicated items

### LINKED ISSUE #430: Fix: do not allow several time the same item among item children
In item edit, it is currently possible to add several time the same item in the list, it will fail at "save". The UI should prevent that early.  For that, the search component should receive the list of `id` "already added" and replace the "+ add" button by a greyed-out "already added" one when the search result is in this list.

Iloveall:
Fixes #430

### DIFF
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
@@ -8,6 +8,7 @@ import { AddedContent } from '../../../shared-components/components/add-content/
 import { ItemRouter } from '../../../../shared/routing/item-router';
 import { bestAttemptFromResults } from '../../../../shared/helpers/attempts';
 import { canCurrentUserViewItemContent } from '../../helpers/item-permissions';
+import { isNotUndefined } from '../../../../shared/helpers/null-undefined-predicates';
 
 export interface ChildData {
   id?: string,
@@ -43,6 +44,7 @@ export class ItemChildrenEditComponent implements OnChanges {
   data: ChildData[] = [];
   selectedRows: ChildData[] = [];
   scoreWeightEnabled = false;
+  addedItemIds: string[] = [];
 
   private subscription?: Subscription;
   @Output() childrenChanges = new EventEmitter<ChildData[]>();
@@ -84,6 +86,7 @@ export class ItemChildrenEditComponent implements OnChanges {
         ).subscribe(children => {
           this.data = children;
           this.scoreWeightEnabled = this.data.some(c => c.scoreWeight !== 1);
+          this.onChildAdded();
           this.state = 'ready';
         },
         _err => {
@@ -96,6 +99,7 @@ export class ItemChildrenEditComponent implements OnChanges {
 
   addChild(child: AddedContent<ItemType>): void {
     this.data.push({ ...child, scoreWeight: DEFAULT_SCORE_WEIGHT });
+    this.onChildAdded();
     this.childrenChanges.emit(this.data);
   }
 
@@ -105,6 +109,7 @@ export class ItemChildrenEditComponent implements OnChanges {
 
   onRemove(): void {
     this.data = this.data.filter(elm => !this.selectedRows.includes(elm));
+    this.onChildAdded();
     this.childrenChanges.emit(this.data);
     this.selectedRows = [];
   }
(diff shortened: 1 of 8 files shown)


src/app/modules/shared-components/components/add-content/add-content.component.ts
```diff
@@ -29,7 +29,9 @@ export class AddContentComponent<Type> implements OnInit, OnDestroy {
   @Input() allowedTypesForNewContent: NewContentType<Type>[] = [];
   @Input() searchFunction?: (searchValue: string) => Observable<AddedContent<Type>[]>;
   @Input() loading = false;
+  @Input() addedItemIds: string[] = [];
```
smadbe, line 32: Add content component is not item specific !

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -107,10 +112,19 @@ export class ItemChildrenEditComponent implements OnChanges {
...
     this.data = this.data.map(c => ({ ...c, scoreWeight: DEFAULT_SCORE_WEIGHT }));
+    this.onChildAdded();
     this.onScoreWeightChange();
   }
 
   onScoreWeightChange(): void {
     this.childrenChanges.emit(this.data);
   }
+
+  onChildAdded(): void {
```
smadbe, line 123: You call in `onRemove` so it is not really `onChildAdded`.
`onChildrenListUpdate"? Actually as it is a function and not really a button callback I would call it by what it does, so "recomputeAddedIds". 
... but it must be called together with the data update, not forgetting any case, otherwise you would have an inconsistent state... I think that would be cleaner to just use the map in the template.

smadbe (changes requested):

src/app/modules/shared-components/components/add-content/add-content.component.ts
Iloveall: Fixed

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
Iloveall: I prefer onChildrenListUpdate as it common pattern, also it could be extended in future for another on updated operations.

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -107,10 +112,19 @@ export class ItemChildrenEditComponent implements OnChanges {
...
     this.data = this.data.map(c => ({ ...c, scoreWeight: DEFAULT_SCORE_WEIGHT }));
+    this.onChildrenListUpdate();
     this.onScoreWeightChange();
   }
 
   onScoreWeightChange(): void {
     this.childrenChanges.emit(this.data);
   }
+
+  onChildrenListUpdate(): void {
```
smadbe, line 123: should be private I assume

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -107,10 +112,19 @@ export class ItemChildrenEditComponent implements OnChanges {
...
+    this.addedItemIds = this.getAddedItemIds();
+  }
+
+  getAddedItemIds(): string[] {
+    return this.data.map(item => item.id).filter(isNotUndefined);
```
smadbe, line 128: If used at a single place, I prefer you put this code directly it is used. (this should not have used outside setting the new value to this.addedItemIds).
(alternatively, you call your function "recomputeAddedItemIds" and this function assign the value directly.

smadbe (changes requested):
small remarks

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
Iloveall: Fixed

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
Iloveall: Fixed

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -107,10 +112,19 @@ export class ItemChildrenEditComponent implements OnChanges {
...
 
   onScoreWeightChange(): void {
     this.childrenChanges.emit(this.data);
   }
+
+  private onChildrenListUpdate(): void {
+    this.recomputeAddedItemIds();
+  }
+
+  recomputeAddedItemIds(): void {
```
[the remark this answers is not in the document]
Iloveall: Fixed

PR #439, opened 2021-05-26
Fixes breadcrumbs when too long

### LINKED ISSUE #431: Fix: truncate breadcrumbs when too long
Of course, when the breadcrumb is very long, it should not push the dark top right menu away from the screen.
Fix it by truncating the beginning of the breadcrumb in CSS so that we can always see the top right bar.
Screenshot 2021-05-21 at 11 08 28

Iloveall:
Fixes  #431


smadbe (changes requested):
Screenshot 2021-05-26 at 17 40 41
The request was " Fix it by truncating the beginning of the breadcrumb in CSS ", so "..." the first elements of the breadcrumbs but keep the latest elements full. 
So in the example, we should see something like "... / Séance 5 : entrainement for et variables / Peinture au numéro 3 - Python".

smadbe (changes requested):
My example with 2 elements was just an example, only what cannot be displayed should be truncated.
The rule should be roughly:
- the last element should always be displayed... if too long, its end could be truncated
- the first element should be displayed except if not place for it
- from right to left, if the element can be display, it has to be displayed, otherwise "..." is shown.

Sorry, it was the clear at first. Ask if it is still unclear if you want to suggest something better.

src/app/core/common/hide-overflow-list.directive.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      if (hiddenElements.length !== targetElements.length - 1 && checkAgain) {
+        this.nextElIdxForHide += 1;
+      }
+
+      this.renderer.addClass(this.el.nativeElement, this.algHideOverflowListStyleClass);
+    }
+
+    if (checkAgain) {
+      this.checkAndHideOverflowedElements();
```
smadbe, line 79: Recursive call with no clear end-condition. :-/
If really needed, it should be a loop (not recursive) with either a provable exit condition, or a way to leave the loop after a max number of calls.
You use the instance variable `nextElIdxForHide` for storing a state of your processing ? :-/ If it is something related to one run of your processing, it should not be an instance variable.

src/app/core/components/breadcrumb/breadcrumb.component.ts
```diff
@@ -22,4 +22,7 @@ export class BreadcrumbComponent {
     }
   }
 
+  trackByFn(index: number): number {
+    return index;
+  }
```
smadbe, line 27: I don't get what it is used for.

src/app/core/components/breadcrumb/breadcrumb.component.scss
```diff
@@ -2,27 +2,59 @@
...
+      .home {
+        display: none;
+      }
+
+      li {
+        &.hidden + li {
+          &:before {
+            content: '...';
+            margin-right: 1rem;
+          }
```
smadbe, line 42: 👍

src/app/core/common/hide-overflow-list.directive.ts
```diff
@@ -0,0 +1,82 @@
...
+  Directive,
+  ElementRef,
+  Input, OnDestroy, OnInit,
+  Renderer2
+} from '@angular/core';
+import { Subscription, timer } from 'rxjs';
+import { distinctUntilChanged, map } from 'rxjs/operators';
+
+@Directive({
+  selector: '[algHideOverflowList]',
```
smadbe, line 11: Why using a directive? Wouldnt be easier/cleaner in the component directly?

smadbe (commented):
It is more complex than expected. I am not convinced by the current implementation with the recursive call. Instead of spending more time into this, I will discuss what would be the ideal final solution to this problem (a way to list the "hidden" parents etc).  So let's close this PR for the moment and I'll come back with a full proposal soon.

Note: When reducing the size, there is a moment where this is not exactly as expected:
Screenshot 2021-06-02 at 12 08 27
(but that's not a big deal)

src/app/core/common/hide-overflow-list.directive.ts
Iloveall: The directive is thing for manipulate HTML structure, it could be done in component it could be done even in index.html but this is not correct.

src/app/core/components/breadcrumb/breadcrumb.component.ts
Iloveall: trackBy need for avoid rerenders which angular does by default

src/app/core/common/hide-overflow-list.directive.ts
Iloveall: Why recursive, imagine we have breadcrumb:

Item 1 -> Item 2 -> Item 3 -> Item 4 -> Item 5.

So when breadcrumb container properties (we are interest scrollWidth*) are changed it runs checkAndHideOverflowedElements method.

checkAndHideOverflowedElements method hides first item when we have overflow (when scroll inside breadcrumb) and asking checkAgain this condition, like if it's enough for hide scroll or no. If there is still scroll - hide next one item.

*Scroll means text was overflowed

smadbe: So probably it should be a loop, but not a recursion.

PR #440, opened 2021-05-26
Assert platform language at app initialisation

### LINKED ISSUE #424: Assert platform language at app initialisation
## Motivations

The platform language should be known. If the language is not provided or if it is not among the available ones, we should display an error preventing access to the app and asking to pick an available language.

## Subtasks

* In the configuration, mark languages as required
* In locale service, mark platform language and languages as required
* On initialisation, assert that platform language is defined (which also means it exists among the configuration languages)
* In the app component, if the platform language is not defined, display a configuration error blocking the UI

ThomasJuster:
Fixes #424

### DIFF
--- a/src/app/core/app.component.html
+++ b/src/app/core/app.component.html
@@ -59,16 +59,3 @@
   <p i18n="@@auth-error-modal-context">Oops, we are unable to make the site work properly. Are you connected to the Internet?</p>
   <p i18n="@@auth-error-modal-todo">Try to refresh the page. If this problem persists, please contact us.</p>
 </p-dialog>
-<p-dialog
-  *ngIf="hasLanguageError"
-  i18n-header header="Configuration error"
-  [visible]="true"
-  [style]="{width: '50vw'}"
-  [modal]="true"
-  [closeOnEscape]="false"
-  [closable]="false"
->
-  <p i18n>
-    Language unknown, try <a href="/">reloading the application</a>. If the problem persist, please contact us.
-  </p>
-</p-dialog>
--- a/src/app/core/app.component.html
+++ b/src/app/core/app.component.html
@@ -59,3 +59,16 @@
   <p i18n="@@auth-error-modal-context">Oops, we are unable to make the site work properly. Are you connected to the Internet?</p>
   <p i18n="@@auth-error-modal-todo">Try to refresh the page. If this problem persists, please contact us.</p>
 </p-dialog>
+<p-dialog
+  *ngIf="hasLanguageError"
+  i18n-header header="Configuration error"
+  [visible]="true"
+  [style]="{width: '50vw'}"
+  [modal]="true"
+  [closeOnEscape]="false"
+  [closable]="false"
+>
+  <p i18n>
+    Language unknown, try <a href="/">reloading the application</a>. If the problem persist, please contact us.
+  </p>
+</p-dialog>
--- a/src/app/core/components/language-picker/language-picker.component.html
+++ b/src/app/core/components/language-picker/language-picker.component.html
@@ -1,8 +1,17 @@
-<p-dropdown
-  *ngIf="languages && current"
-  [options]="languages"
-  [ngModel]="current"
-  optionLabel="tag"
-  [styleClass]="styleClass || 'alg-language-picker-dropdown'"
-  (onChange)="languageChanged($event)"
-></p-dropdown>
+<ng-container *ngIf="languages.length > 1; then picker else readonlyContainer"></ng-container>
+<ng-template #picker>
+  <p-dropdown
+    *ngIf="current"
+    [options]="languages"
+    [ngModel]="current"
+    optionLabel="tag"
+    [styleClass]="styleClass || 'alg-language-picker-dropdown'"
+    (onChange)="languageChanged($event)"
+  ></p-dropdown>
+</ng-template>
+
+<ng-template #readonlyContainer>
+  <span *ngIf="current">
+    {{ current.tag | uppercase }}
+  </span>
+</ng-template>
--- a/src/app/core/components/language-picker/language-picker.component.html
+++ b/src/app/core/components/language-picker/language-picker.component.html
@@ -1,17 +1,8 @@
-<ng-container *ngIf="languages.length > 1; then picker else readonlyContainer"></ng-container>
-<ng-template #picker>
-  <p-dropdown
-    *ngIf="current"
-    [options]="languages"
-    [ngModel]="current"
-    optionLabel="tag"
-    [styleClass]="styleClass || 'alg-language-picker-dropdown'"
-    (onChange)="languageChanged($event)"
-  ></p-dropdown>
-</ng-template>
-
-<ng-template #readonlyContainer>
-  <span *ngIf="current">
-    {{ current.tag | uppercase }}
-  </span>
-</ng-template>
+<p-dropdown
+  *ngIf="languages.length > 1 && current"
+  [options]="languages"
+  [ngModel]="current"
+  optionLabel="tag"
+  [styleClass]="styleClass || 'alg-language-picker-dropdown'"
+  (onChange)="languageChanged($event)"
+></p-dropdown>
--- a/src/app/core/components/language-picker/language-picker.component.html
+++ b/src/app/core/components/language-picker/language-picker.component.html
@@ -2,7 +2,6 @@
   *ngIf="languages.length > 1 && current"
   [options]="languages"
   [ngModel]="current"
-  optionLabel="tag"
   [styleClass]="styleClass || 'alg-language-picker-dropdown'"
   (onChange)="languageChanged($event)"
 ></p-dropdown>
--- a/src/app/core/components/language-picker/language-picker.component.ts
+++ b/src/app/core/components/language-picker/language-picker.component.ts
@@ -13,7 +13,7 @@ export class LanguagePickerComponent implements OnChanges {
   @Output() changeLang = new EventEmitter<string>();
 
   readonly languages = this.localeService.languages;
-  current = this.languages?.find(l => l.tag === this.localeService.currentTag);
+  current = this.languages.find(l => l.tag === this.localeService.currentTag);
 
   constructor(
     private localeService: LocaleService,
--- a/src/app/core/components/language-picker/language-picker.component.ts
+++ b/src/app/core/components/language-picker/language-picker.component.ts
@@ -1,5 +1,4 @@
 import { Component, Input, OnChanges, Output, EventEmitter, OnInit } from '@angular/core';
-import { LanguageConfig } from 'src/app/shared/helpers/config';
 import { LocaleService } from '../../services/localeService';
 
 @Component({
@@ -13,8 +12,8 @@ export class LanguagePickerComponent implements OnInit, OnChanges {
   @Input() redirectOnChange = true;
   @Output() changeLang = new EventEmitter<string>();
 
-  readonly languages = this.localeService.languages;
-  current?: LanguageConfig
+  readonly languages = this.localeService.languages.map(({ tag }) => ({ label: tag, value: tag }));
+  current?: string
 
   constructor(
     private localeService: LocaleService,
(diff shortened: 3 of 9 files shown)


src/app/core/services/localeService.ts
```diff
@@ -1,24 +1,41 @@
...
+      : this.getNextPathnameForDevelopment(nextLang);
+    window.location.href = `${pathname}${window.location.hash}`;
+  }
+
+  private getNextPathnameForDevelopment(nextLang: Language): string {
+    return nextLang.path;
+  }
+
+  private getNextPathname(currentLang: Language, nextLang: Language): string {
+    return window.location.pathname.replace(currentLang.path, nextLang.path);
```
ThomasJuster, line 38: This change might not be in the scope of the PR, I'll detail it shortly.

When changing the language, the app is reloaded by replacing in the URL the current tag by next tag. But in development, the current tag is not in the URL, as a consequence the URL computed for next tag is exactly the same, causing the app not to be reloaded.

I will need this change for the last step (which is displaying a language mismatch dialog), since this PR is quite small I assumed it was ok to do it here.

src/app/core/services/localeService.ts
```diff
@@ -1,24 +1,41 @@
...
+    window.location.href = `${pathname}${window.location.hash}`;
+  }
+
+  private getNextPathnameForDevelopment(nextLang: LanguageConfig): string {
+    return nextLang.path;
+  }
+
+  private getNextPathname(currentLang: LanguageConfig, nextLang: LanguageConfig): string {
+    return window.location.pathname.replace(currentLang.path, nextLang.path);
   }
```
ThomasJuster, line 39: This change might not be in the scope of the PR, I'll detail it shortly.

When changing the language, the app is reloaded by replacing in the URL the current tag by the next tag, but in development, the current tag is not in the URL. As a consequence, the URL computed for next tag is exactly the same the current URL, causing the app *not* to be reloaded.

I will need this change for the last step (which is displaying a language mismatch dialog). Since this PR is quite small I assumed it was ok to do it here.

src/app/core/services/localeService.ts
```diff
@@ -1,24 +1,41 @@
...
-  readonly currentTag?: string;
+  readonly languages: LanguageConfig[];
+  readonly currentTag: string;
+  readonly currentTagError: boolean
 
   constructor() {
     this.languages = appConfig.languages;
-    this.currentTag = this.languages?.find(l => window.location.pathname.endsWith(l.path))?.tag;
+    const fallbackTag = appConfig.production ? '' : 'en';
+    this.currentTag = this.languages.find(l => window.location.pathname.endsWith(l.path))?.tag ?? fallbackTag;
```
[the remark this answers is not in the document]
ThomasJuster: Adding the language config `{ path: '/', tag: 'en' }` does not work because the path `"/"` matches every algorea url, it will fallback to english every time the url language is none of the configured ones. There will never be any issue with that config but the language error will be silent.
Plus, it adds a tricky situation because since it matches every url, it must stay at the end of the config languages array.

Since I considered that matching `"/"` was not a solution, I needed to handle the development use-case.
Maybe it will be easier to talk about it :sweat_smile:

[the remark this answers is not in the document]
ThomasJuster: Another solution is to change the language matching logic: instead of using `pathname.endsWith(langTag)`, we would have to match it entirely: `pathname === langTag`. But I don't know if that would be a breaking change in production.

src/app/core/app.component.html
```diff
@@ -59,3 +59,16 @@
...
+  *ngIf="hasLanguageError"
+  i18n-header header="Configuration error"
+  [visible]="true"
+  [style]="{width: '50vw'}"
+  [modal]="true"
+  [closeOnEscape]="false"
+  [closable]="false"
+>
+  <p i18n>
+    Language unknown, try <a href="/">reloading the application</a>. If the problem persist, please contact us.
```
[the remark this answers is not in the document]
ThomasJuster: At first I thought using the fatal error as well, then I tested how the app behaved for [dev.algorea.com](dev.algorea.com) and considered that it was the way the production worked too. Assuming that, the behavior is:

Case 1: if the language is _defined_ but not found, it means the app is hosted for a language that is not in the configured ones. That is a weird case of a devops error, only the administrator(s) of the platform would know what to do.
Reloading with `"/"` would lead to `dev.algorea.com` which will then redirect `dev.algorea.com/en`, providing a nice fallback for the user who will at least be able to access the app. He will then be able to change the language if needed. However, this case 1 scenario is very very unlikely to happen.

Case 2: the language is not _defined_ (when `pathname === '/'`), here the server already falls back to `/en`

Bottom note: That was for the production behavior only

src/app/core/services/localeService.ts
ThomasJuster: This discussion leads to the same as https://github.com/France-ioi/AlgoreaFrontend/pull/440#discussion_r639838165

src/app/core/services/localeService.ts
```diff
@@ -1,24 +1,27 @@
...
 
 @Injectable({
   providedIn: 'root'
 })
 export class LocaleService {
 
-  readonly languages?: { tag: string, path: string }[];
-  readonly currentTag?: string;
+  readonly languages: LanguageConfig[];
+  readonly currentLang: LanguageConfig | undefined;
```
ThomasJuster, line 12: `currentLang` cannot be defined-only by design because even if we display a fatal error, the app is still displayed in the background. In that use case only the `currentLang` will be `undefined`, but that case is also handled

src/app/core/services/localeService.ts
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/core/app.component.html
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/core/services/localeService.ts
ThomasJuster: Fixed

src/app/core/components/language-picker/language-picker.component.html
```diff
@@ -1,8 +1,17 @@
...
+    optionLabel="tag"
+    [styleClass]="styleClass || 'alg-language-picker-dropdown'"
+    (onChange)="languageChanged($event)"
+  ></p-dropdown>
+</ng-template>
+
+<ng-template #readonlyContainer>
+  <span *ngIf="current">
+    {{ current.tag | uppercase }}
+  </span>
```
smadbe, line 16: For a single-language platform, I don't think that's useful to waste top-bar space for indicating the language.

src/app/core/components/language-picker/language-picker.component.ts
```diff
@@ -13,15 +13,15 @@ export class LanguagePickerComponent implements OnChanges {
...
+  current = this.localeService.currentLang;
 
   constructor(
     private localeService: LocaleService,
   ) {}
 
   ngOnChanges(): void {
     if (this.defaultLang) {
-      this.current = this.languages?.find(l => l.tag === this.defaultLang);
+      this.current = this.languages.find(l => l.tag === this.defaultLang);
```
smadbe, line 24: You both set `current` to the currentLang of locale service and set it to the default ?
This default to local current is strange when you set a value...
Maybe you could not set default value to current and in `ngOnChanges`:
```ts
this.current = this.defaultLang ? this.languages.find(l => l.tag === this.defaultLang) : this.localeService.currentLang;
```

src/app/core/services/localeService.ts
smadbe: Ok.
So write it:
```ts
readonly currentLang?: LanguageConfig;
```

src/app/core/services/localeService.ts
```diff
@@ -9,12 +10,12 @@ export class LocaleService {
 
   readonly languages: LanguageConfig[];
   readonly currentLang: LanguageConfig | undefined;
-  readonly currentTagError
+  readonly currentLangError$: Observable<boolean>
```
smadbe, line 13: why boolean?

src/app/core/components/language-picker/language-picker.component.ts
```diff
@@ -13,15 +13,15 @@ export class LanguagePickerComponent implements OnChanges {
   @Output() changeLang = new EventEmitter<string>();
 
   readonly languages = this.localeService.languages;
-  current = this.languages?.find(l => l.tag === this.localeService.currentTag);
+  current = this.localeService.currentLang;
```
smadbe, line 16: As the value used by the picker is only the label as the navigation is done only by the locale service now, I think the language picker should only deal with strings and string lists.

src/app/shared/helpers/config.ts
```diff
@@ -17,7 +22,7 @@ export type Environment = {
   defaultActivityId: string;
   defaultSkillId: string;
 
-  languages?: { tag: string, path: string }[];
+  languages: LanguageConfig[];
```
smadbe, line 25: There must be an explanation on how to use this (or/and an example), if the match all "/" has to be at the end, it must be mentioned.

smadbe (changes requested):
And also, probably changed in a previous PR:
`current-user.component.ts:L40`, there is no need for the `if (changes.default_language) { ` is the value is not optional. (or it may be optional)

src/app/shared/helpers/config.ts
ThomasJuster: Fixed

src/app/core/components/language-picker/language-picker.component.ts
ThomasJuster: Fixed

src/app/core/components/language-picker/language-picker.component.ts
ThomasJuster: Fixed

src/app/core/components/language-picker/language-picker.component.html
ThomasJuster: Fixed

src/app/core/services/localeService.ts
ThomasJuster: Fixed

src/app/core/components/language-picker/language-picker.component.ts
```diff
@@ -1,28 +1,31 @@
...
 
+  ngOnInit(): void {
+    this.initCurrentLang();
+  }
+
   ngOnChanges(): void {
-    if (this.defaultLang) {
-      this.current = this.languages.find(l => l.tag === this.defaultLang);
-    }
+    this.initCurrentLang();
```
smadbe, line 27: ngChanges is called at init as well (check it), so no need to have ngInit and a separated function

src/app/core/components/language-picker/language-picker.component.ts
```diff
@@ -28,16 +27,16 @@ export class LanguagePickerComponent implements OnInit, OnChanges {
...
 
     if (this.redirectOnChange) {
-      this.localeService.navigateTo(lang.value.tag);
+      this.localeService.navigateTo(lang.value);
     }
   }
 
   private initCurrentLang(): void {
-    this.current = this.defaultLang ? this.languages.find(({ tag }) => tag === this.defaultLang) : this.localeService.currentLang;
+    this.current = this.defaultLang || this.localeService.currentLang?.tag;
```
smadbe, line 39: In TS, should be `this.defaultLang ?? this.localeService.currentLang?.tag;`

smadbe (changes requested):

src/app/core/components/language-picker/language-picker.component.ts
ThomasJuster: I checked it before adding the `ngOnInit` method, the `ngOnChanges` is not triggered at init

src/app/core/components/language-picker/language-picker.component.ts
ThomasJuster: Fixed.

smadbe (approved):
looks good 👍

PR #450, opened 2021-05-31
provide platform language when creating temp user

### LINKED ISSUE #425: Provide platform language when creating a temporary user
## Motivations

Currently, temporary users are created using `fr` as default language. This will be problematic when we will handle current user language not matching the platform language. For temporary users (any unregistered visitor), whenever the platform will be launched in English, there will always be a dialog prompting that the language doesn't match even before they could pick one.

## Subtasks

* Update the auth http service method `createTempUser` to accept `default_language`
* In the auth service, access the locale service to retrieve the platform language, use it to create temporary users: in the constructor and when the token is invalid.

ThomasJuster:
Fixes #425

### DIFF
--- a/src/app/shared/auth/auth.service.ts
+++ b/src/app/shared/auth/auth.service.ts
@@ -14,6 +14,7 @@ import {
   hasForcedToken,
   forcedTokenAuthFromStorage
 } from './auth-info';
+import { LocaleService } from 'src/app/core/services/localeService';
 
 // Lifetime under which we refresh the token.
 export const minTokenLifetime = 5*MINUTES;
@@ -40,7 +41,8 @@ export class AuthService implements OnDestroy {
 
   constructor(
     private oauthService: OAuthService,
-    private authHttp: AuthHttpService
+    private authHttp: AuthHttpService,
+    private localeService: LocaleService
   ) {
     // (1) check if a code/state is given in URL (i.e., we are back from a oauth login redirect) and try to get a token from it.
     oauthService.tryCompletingCodeFlowLogin().pipe(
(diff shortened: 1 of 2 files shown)


src/app/shared/auth/auth.service.ts
```diff
@@ -50,10 +52,13 @@ export class AuthService implements OnDestroy {
...
         else return this.authHttp.refreshCookie(); // will fail if the browser has no cookie
       }),
-      catchError(_e =>
+      catchError(_e => {
         // (3) otherwise, create a temp session
-        this.authHttp.createTempUser().pipe(retry(2))
-      ),
+        const defaultLanguage = this.localeService.currentLang?.tag;
+        // If no default language, the app is in error state, no need to create a temp user.
+        if (!defaultLanguage) return EMPTY;
```
smadbe, line 59: Probably that's better to throw an error so that the observable fails instead of completing without emitting anything. It may make debugging easier.

src/app/shared/http-services/auth.http-service.ts
```diff
@@ -20,21 +20,23 @@ const longAuthServicesTimeout = 10000;
 export class AuthHttpService {
 
   private http: HttpClient; // an http client specific to this class, skipping all http interceptors
-  private cookieParams = new HttpParams({ fromObject: appConfig.authType === 'cookies' ? {
+  private cookieParams = appConfig.authType === 'cookies' ? {
     use_cookie: '1',
     cookie_secure: appConfig.secure ? '1' : '0',
     cookie_same_site: appConfig.sameSite ? '1' : '0',
-  } : {} });
+  } : undefined;
```
smadbe, line 27: `{}`?

I supposed you have tested that `...undefined` works as expected.

smadbe (approved):

src/app/shared/http-services/auth.http-service.ts
ThomasJuster: Yes, it does and that works for any falsy values actually. But only for objects, with arrays it breaks.
Using the empty object is invalid for TypeScript because `HttpParams` `fromObject` option expects either an undefined value, either a populated object. :shrug:

src/app/shared/auth/auth.service.ts
ThomasJuster: Fixed

smadbe (approved):

PR #451, opened 2021-05-31
handle language mismatch

### LINKED ISSUE #390: Handle current language not matching user's

ThomasJuster:
Fixes #390

### DIFF
--- a/src/app/core/components/language-mismatch/language-mismatch.component.ts
+++ b/src/app/core/components/language-mismatch/language-mismatch.component.ts
@@ -1,0 +1,43 @@
+import { Component } from '@angular/core';
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { mapPending } from 'src/app/shared/operators/map-pending';
+import { UserSessionService } from 'src/app/shared/services/user-session.service';
+import { LocaleService } from '../../services/localeService';
+
+@Component({
+  selector: 'alg-language-mismatch',
+  templateUrl: './language-mismatch.component.html',
+  styleUrls: [ './language-mismatch.component.scss' ],
+})
+export class LanguageMismatchComponent {
+
+  readonly currentLanguage = this.localeService.currentLang?.tag;
+  readonly languages = this.localeService.languages;
+
+  languageMismatch$ = this.sessionService.userProfile$.pipe(
+    map(profile => {
+      const hasLanguageMismatch = !!this.currentLanguage && profile.defaultLanguage !== this.currentLanguage;
+      return hasLanguageMismatch ? {
+        userDefaultLanguage: profile.defaultLanguage,
+        userDefaultLanguageIsSupported: this.languages.some(({ tag }) => tag === profile.defaultLanguage),
+      } : undefined;
+    }),
+  )
+  updating$?: Observable<boolean>
+
+  constructor(
+    private localeService: LocaleService,
+    private sessionService: UserSessionService,
+  ) { }
+
+  onUpdateUserLanguage(language: string | undefined): void {
+    if (!language) return;
+    this.updating$ = this.sessionService.updateCurrentUser({ default_language: language }).pipe(mapPending());
+  }
+
+  onVisitPlatformInUserLanguage(language: string): void {
+    this.localeService.navigateTo(language);
+  }
+
+}
--- a/src/app/core/components/language-mismatch/language-mismatch.component.ts
+++ b/src/app/core/components/language-mismatch/language-mismatch.component.ts
@@ -19,8 +19,8 @@ export class LanguageMismatchComponent {
       userDefaultLanguage: profile.defaultLanguage,
       userDefaultLanguageIsSupported: this.localeService.languages.some(({ tag }) => tag === profile.defaultLanguage),
     })),
-  )
-  updating$?: Observable<boolean>
+  );
+  updating$?: Observable<boolean>;
 
   constructor(
     private localeService: LocaleService,
--- a/src/app/core/components/language-mismatch/language-mismatch.component.ts
+++ b/src/app/core/components/language-mismatch/language-mismatch.component.ts
@@ -32,7 +32,7 @@ export class LanguageMismatchComponent {
   ) { }
 
   onUpdateUserLanguage(language?: string): void {
-    if (!language) return;
+    if (!language) throw new Error('language should be defined');
     this.updating$ = this.sessionService.updateCurrentUser({ default_language: language }).pipe(mapPending());
   }
 
--- a/src/app/core/components/language-mismatch/language-mismatch.component.ts
+++ b/src/app/core/components/language-mismatch/language-mismatch.component.ts
@@ -1,5 +1,4 @@
 import { Component } from '@angular/core';
-import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { mapPending } from 'src/app/shared/operators/map-pending';
 import { UserSessionService } from 'src/app/shared/services/user-session.service';
@@ -20,7 +19,7 @@ export class LanguageMismatchComponent {
       userDefaultLanguageIsSupported: this.localeService.languages.some(({ tag }) => tag === profile.defaultLanguage),
     })),
   );
-  updating$?: Observable<boolean>;
+  updating = false;
 
   constructor(
     private localeService: LocaleService,
@@ -29,7 +28,9 @@ export class LanguageMismatchComponent {
 
   onUpdateUserLanguage(language?: string): void {
     if (!language) throw new Error('language should be defined');
-    this.updating$ = this.sessionService.updateCurrentUser({ default_language: language }).pipe(mapPending());
+    this.sessionService.updateCurrentUser({ default_language: language })
+      .pipe(mapPending())
+      .subscribe(updating => (this.updating = updating));
   }
 
   onVisitPlatformInUserLanguage(language: string): void {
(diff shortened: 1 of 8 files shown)


src/app/core/components/language-mismatch/language-mismatch.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  languageMismatch$ = this.sessionService.userProfile$.pipe(
+    map(profile => {
+      const hasLanguageMismatch = !!this.currentLanguage && profile.defaultLanguage !== this.currentLanguage;
+      return hasLanguageMismatch ? {
+        userDefaultLanguage: profile.defaultLanguage,
+        userDefaultLanguageIsSupported: this.languages.some(({ tag }) => tag === profile.defaultLanguage),
+      } : undefined;
+    }),
+  )
+  updating$?: Observable<boolean>
```
smadbe, line 27: We currently put ";" at the end of declaration lines. (missing twice)

src/app/core/components/language-mismatch/language-mismatch.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  languageMismatch$ = this.sessionService.userProfile$.pipe(
+    map(profile => {
+      const hasLanguageMismatch = !!this.currentLanguage && profile.defaultLanguage !== this.currentLanguage;
+      return hasLanguageMismatch ? {
+        userDefaultLanguage: profile.defaultLanguage,
+        userDefaultLanguageIsSupported: this.languages.some(({ tag }) => tag === profile.defaultLanguage),
+      } : undefined;
+    }),
+  )
```
smadbe, line 26: Code style: 
- `!!this.currentLanguage` is useless in  `!!this.currentLanguage && profile.defaultLanguage !== this.currentLanguage;` as it would not match anyway with "===" if it was undefined. (except if you want to exclude the empty string but I do not see the point)
- so L20 becomes `const hasLanguageMismatch = profile.defaultLanguage !== this.currentLanguage;`
- so no need to define an extra constant
- so there is one line of code in the block 
- so we could just write:
```suggestion
    map(profile => (profile.defaultLanguage !== this.currentLanguage ? {
      userDefaultLanguage: profile.defaultLanguage,
      userDefaultLanguageIsSupported: this.languages.some(({ tag }) => tag === profile.defaultLanguage),
    } : undefined)),
  )
```

src/app/core/components/language-mismatch/language-mismatch.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  readonly languages = this.localeService.languages;
+
+  languageMismatch$ = this.sessionService.userProfile$.pipe(
+    map(profile => {
+      const hasLanguageMismatch = !!this.currentLanguage && profile.defaultLanguage !== this.currentLanguage;
+      return hasLanguageMismatch ? {
+        userDefaultLanguage: profile.defaultLanguage,
+        userDefaultLanguageIsSupported: this.languages.some(({ tag }) => tag === profile.defaultLanguage),
+      } : undefined;
+    }),
```
smadbe, line 25: Code style: 
- `!!this.currentLanguage` is useless in  `!!this.currentLanguage && profile.defaultLanguage !== this.currentLanguage;` as it would not match anyway with "===" if it was undefined. (except if you want to exclude the empty string but I do not see the point)
- so L20 becomes `const hasLanguageMismatch = profile.defaultLanguage !== this.currentLanguage;`
- so no need to define an extra constant
- so there is one line of code in the block 
- so we could just write:
```suggestion
    map(profile => (profile.defaultLanguage !== this.currentLanguage ? {
      userDefaultLanguage: profile.defaultLanguage,
      userDefaultLanguageIsSupported: this.languages.some(({ tag }) => tag === profile.defaultLanguage),
    } : undefined)),
```
I even prefer
```suggestion
    map(profile => (profile.defaultLanguage === this.currentLanguage ? undefined : {
      userDefaultLanguage: profile.defaultLanguage,
      userDefaultLanguageIsSupported: this.languages.some(({ tag }) => tag === profile.defaultLanguage),
    })),
```

src/app/core/components/language-mismatch/language-mismatch.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Component({
+  selector: 'alg-language-mismatch',
+  templateUrl: './language-mismatch.component.html',
+  styleUrls: [ './language-mismatch.component.scss' ],
+})
+export class LanguageMismatchComponent {
+
+  readonly currentLanguage = this.localeService.currentLang?.tag;
+  readonly languages = this.localeService.languages;
```
smadbe, line 16: This seems to be used at only one place, just use `this.localeService.languages` directly on line 23

src/app/core/components/language-mismatch/language-mismatch.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  )
+  updating$?: Observable<boolean>
+
+  constructor(
+    private localeService: LocaleService,
+    private sessionService: UserSessionService,
+  ) { }
+
+  onUpdateUserLanguage(language?: string): void {
+    if (!language) return;
```
smadbe, line 35: This is unexpected. Let's "throw" in such a case so that we can notice the error.

src/app/shared/operators/map-pending.ts
```diff
@@ -0,0 +1,16 @@
...
+import { catchError, mapTo, startWith } from 'rxjs/operators';
+
+/**
+ * Operator to indicate if the source action is pending
+ * - Until a value is added to the stream, the action is pending.
+ * - When a value is added to the stream, the action is finished.
+ * - When an error is received, the action is finished.
+ */
+export function mapPending() {
+  return (source: Observable<any>): Observable<boolean> => source.pipe(
```
smadbe, line 11: You should add unit tests :-)

src/app/core/components/language-mismatch/language-mismatch.component.ts
smadbe, line 27: I'm a bit annoyed by the fact this variable is optional and not read-only. We are a bit playing with the limits of Angular template engine to know how `*ngIf="updating$ | async"` works when it is not set, when it is set once, when it is replaced by another observable.  🤔 I have to think about this.

smadbe (changes requested):
👍 
Initial review.

src/app/core/components/language-mismatch/language-mismatch.component.ts
ThomasJuster: Shouldn't there be an eslint rule or a formatter for such things ?

ThomasJuster: Fixed

src/app/core/components/language-mismatch/language-mismatch.component.ts
ThomasJuster: Fixed

src/app/core/components/language-mismatch/language-mismatch.component.ts
ThomasJuster: Fixed

src/app/core/components/language-mismatch/language-mismatch.component.ts
ThomasJuster: Fixed

src/app/core/components/language-mismatch/language-mismatch.component.ts
ThomasJuster: Fixed

src/app/shared/operators/map-pending.ts
ThomasJuster: Fixed

src/app/core/components/language-mismatch/language-mismatch.component.ts
ThomasJuster: Fixed

smadbe: Yes sure, please check if you find such a rule.

smadbe (approved):
👍 👍

PR #455, opened 2021-06-02
use fp-ts es6 build (by removing commonjs imports)

ThomasJuster:


smadbe (approved):
Just changing the path makes things different on what is built?

ThomasJuster:
Yes, `fp-ts/lib/Either` points explicitly to Either commonjs build, `fp-ts/es6/Either` points explicitly to Either module build, and `fp-ts/Either` points to `fp-ts/Either/package.json` which has commonjs and module entry points.
Using the export to `fp-ts/Either` will let the bundler decide which version to take.

smadbe:
👍

PR #464, opened 2021-06-07
Merges userDisplay with userCaption pipe

### LINKED ISSUE #452: Bug in user formatting
User formatting does not work correctly in multiple places: 
Screenshot 2021-06-01 at 10 58 58
Screenshot 2021-06-01 at 11 02 25

- probably because they are undefined while your pipe do not handle `undefined`.
- because the list of members use `let-rowData` which returns a `any` 😡. Would you have a proposal to get the types for the table rows?
- UserDescendant (and probably similar services) assumes first/lastnames cannot be `undefined` while they may. So fix them. (and maybe use the decoder instead)

Iloveall:
Fixes #452

### DIFF
--- a/src/app/modules/group/components/pending-request/pending-request.component.html
+++ b/src/app/modules/group/components/pending-request/pending-request.component.html
@@ -68,7 +68,7 @@
                   /> -->
                   <div class="user-info">
                     <span class="user-info-name">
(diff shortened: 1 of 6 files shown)


smadbe (approved):
I haven't fully tested it. I hope you did :-)

PR #466, opened 2021-06-07
move group watching detail from page navigation to group header

### LINKED ISSUE #465: Move group watching detail from page navigation to group header

ThomasJuster:
Fixes #465


smadbe (approved):
👍 
So, no template at the end.

ThomasJuster:
Indeed, after viewing the component page-navigator it meant either deleting it and move buttons in the group-header and item-header or leave it as is. I opted for the 2nd choice, the 1st one was too impacting. Maybe later if we run into further problems ?

PR #470, opened 2021-06-09
`mapToFetchState`: add previous data to fetching state

smadbe:
Until now, using `mapToFetchState` did not allow to know the latest ready data when refetching. This is useful to reload while showing (even greyed out) the former data. For instance, not to have the full table emptying then back full when updating the members of a group.
This PR add to the "fetching state" the latest data, if any.

Better reviewed at a whole as I changed the way to proceed on the last commit as the first implementation was not fully working in any case.

I have still to update at least one place (group member remove) where the component had to handle that itself. But I'll do it in a separated PR.

### DIFF
--- a/src/app/shared/operators/state.spec.ts
+++ b/src/app/shared/operators/state.spec.ts
@@ -148,6 +148,62 @@ describe('mapToState', () => {
     });
   });
 
+  it('should allow several fetching states with no data', () => {
+    testScheduler.run(helpers => {
+      const { cold, expectObservable } = helpers;
+      const e1 =  cold('---1|');
+      const res = cold('-0------|'); // the values are not important for the resetter
+      const expected = 'ff--a---|';
+
+      expectObservable(e1.pipe(
+        mapToFetchState({ resetter: res }),
+        stateToLetter(),
+      )).toBe(expected);
+    });
+  });
+
+  it('should pass data between fetching states', () => {
+    testScheduler.run(helpers => {
+      const { cold, expectObservable } = helpers;
+      const e1 =  cold('--1|');
+      const res = cold('----00----|'); // the values are not important for the resetter
+      const expected = 'f-a-gg-a--|';
+
+      expectObservable(e1.pipe(
+        mapToFetchState({ resetter: res }),
+        stateToLetter(),
+      )).toBe(expected);
+    });
+  });
+
+  it('should not replay the last data after a failure', () => {
+    testScheduler.run(helpers => {
+      const { cold, expectObservable } = helpers;
+      const e1 =  cold('-1-#');
+      const res = cold('-------0------|');
+      const expected = 'fa-x---fa-x---|';
+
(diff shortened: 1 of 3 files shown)


src/app/shared/operators/state.spec.ts
```diff
@@ -130,9 +137,65 @@ describe('mapToState', () => {
...
+      )).toBe(expected);
+    });
+  });
+
+  it('should change the data in state as expected', () => {
+    testScheduler.run(helpers => {
+      const { cold, expectObservable } = helpers;
+      const e1 =  cold('--1--2|');
+      const res = cold('----0------0-----|'); // the values are not important for the resetter
+      const expected = 'f-a-g-a--b-h-a--b|';
```
ThomasJuster, line 198: Although the mapping letter/state is quite readable, as the test is growing the same mapping becomes harder and harder to read, and probably to maintain too.
I think it would be worth looking for another way, digging around the second arguments of `cold()` and `expectObservable(…).toBe()`
As an example, it could give something like this:
```ts
const e1 =  cold('--1--2', { 1: 'hello', 2: 'world' });
const res = cold('--0---0-'); // values are not important
const expected = 'f-a---g-a--';
expectObservable(e1.pipe(
  mapToFetchState(),
  map(state => `tag: ${state.tag}, value: ${String(state.data)}`), // any serialization does the trick.
)).toBe(expected, {
  f: 'tag: fetching, value: undefined',
  g: 'tag: fetching, value: hello',
  a: 'tag: ready, value: hello world',
});
```
(Note: The test is invalid, it's a quick sample.)

That way, there would be no need to know the whole letter/state mapping, each test would encapsulate its input/output expectation.

What do you think of the general idea ?

src/app/shared/operators/state.ts
```diff
@@ -18,6 +19,11 @@ export function mapToFetchState<T>(config?: { resetter?: Observable<unknown> }):
       startWith(noop),
       switchMapTo(source)
     ),
+    map(state => {
+      if (state.isReady) previousData = state.data;
+      if (state.isError) previousData = undefined;
+      return state.isFetching ? fetchingState(previousData) : state;
+    }),
```
ThomasJuster, line 26: Using a `let` variable looks like a side effect no ? Or maybe it's okay because it is a local variable of the operator.
I think using `pairwise()` would be more suited, but it currently does the trick well enough, your call.

smadbe: Actually, the actual implementation of operator use states (for instance [switchMap](https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/switchMap.ts)) for their internal mechanic. It is their usage which should not use anything else than pure functional. 
My initial implementation was using `pairwise` but that does not work: if I get 2 fetching in a row, I do not have any clue of the data to be put in the second fetching. (pairwise pairs the 2 latest inputs, not the latest output with the input)

ThomasJuster (approved):

PR #471, opened 2021-06-11
Feature/unify other and own user pages

### LINKED ISSUE #403: Unify other user's and profile pages
## Motivations
Unify other user's and (own) profile pages as some of their content are shared.

## Subtasks
Make the personal page a page with 2 tabs:
- "Progress" tab is displayed on `/users/:id` (only display the tab when current user's page) and has the progress content from #401 . Viewing your progress is the same as you would "watch" yourself, but so the "watched_group_id" parameter does not have to be given in the service.
- "Personal data" tab is displayed on `/users/:my_user_id/personal_data` (replace `/groups/me` by this one!).
  * Reject ("You cannot access this page for this user") if the the input id is not the current user
  * Display the content of the previous `/groups/me` page.

Iloveall:
Fixes #403


smadbe (changes requested):
If you cannot access the personal data page and you are on the "progress" "subpage" just do not display the tabs. 
If you cannot access the personal data page and you are on the " personal data " "subpage" leave the tabs (so that the user can navigate on progress) but display the error on a white background like all the other pages under tabs.

smadbe (changes requested):
I don't feel " If you cannot access the personal data page and you are on the " personal data " "subpage" leave the tabs (so that the user can navigate on progress) but display the error on a white background like all the other pages under tabs." is correctly implemented.
Screenshot 2021-07-05 at 12 26 36

src/app/modules/group/pages/group-user/group-user.component.html
```diff
@@ -21,5 +21,42 @@
...
+    </ng-container>
+
+    <ng-container *ngIf="personalData?.isActive">
+      <ng-container *ngIf="currentUserGroupId$ | async as currentUserGroupId">
+        <div class="bg-white" *ngIf="currentUserGroupId !== user.groupId">
+          <p class="no-access" i18n>You cannot access this page for this user</p>
+        </div>
+
+        <alg-current-user *ngIf="currentUserGroupId === user.groupId"></alg-current-user>
+      </ng-container>
```
smadbe, line 58: May be more readable:
```suggestion
      <ng-container *ngIf="(currentUserGroupId$ | async) === user.groupId; then currentUser; else forbidden "></ng-container>
      <ng-template #currentUser>
        <alg-current-user></alg-current-user>
      </ng-template>
      <ng-template #forbidden>
        <div class="bg-white">
            <p class="no-access" i18n>You cannot access this page for this user</p>
          </div>
      </ng-template>
```

src/app/modules/group/pages/group-user/group-user.component.ts
```diff
@@ -1,24 +1,80 @@
...
+  @ViewChild('personalData') personalData?: RouterLinkActive;
+
   readonly state$ = this.route.params.pipe(
     switchMap(({ id }) => this.getUser$(id)),
-    mapToFetchState()
+    mapToFetchState(),
+  );
+
+  readonly currentUserGroupId$ = this.userSessionService.userProfile$.pipe(
+    delay(0),
```
smadbe, line 26: Have you experienced that this `delay(0)` was required?

src/app/modules/group/pages/group-user/group-user.component.ts
```diff
@@ -1,24 +1,80 @@
...
+
+  ngOnInit(): void {
+    this.subscription = this.router.events
+      .pipe(
+        filter(event => event instanceof NavigationEnd),
+        startWith(null),
+      )
+      .subscribe(() => {
+        this.updateBreadcrumbs();
+      });
```
smadbe, line 50: Just follow your state and display "Users \ `user|caption` \ Personal info or Progress" when you know your user and  "Users \  ... \ Personal info or Progress" on loading/error.
Like in group-by-id.component.ts:50-59, just map your state to a breadcrumb, and then set it in your subscribe.

src/app/modules/group/pages/group-user/group-user.component.ts
```diff
@@ -1,24 +1,80 @@
...
 
-  getUser$(id: string): Observable<User> {
+    this.isInitPagePersonal = this.router.url.includes('personal-data');
+  }
+
+  ngOnDestroy(): void {
+    this.subscription?.unsubscribe();
```
smadbe, line 56: You need to flush the current current when leaving the page.

src/app/modules/group/pages/group-user/group-user.component.ts
```diff
@@ -1,24 +1,80 @@
...
+  }
+
+  private getUser$(id: string): Observable<User> {
```
smadbe, line 59: IMO, it is less clean to have function `getUser$(id)` while `this.getUserService.getForId(id)` was completely understandable if used at line 21 directly. (okay if it was complicated to understand what it does)

src/app/shared/services/user-session.service.ts
```diff
@@ -40,6 +40,10 @@ export class UserSessionService implements OnDestroy {
     return !!this.session$.value?.watchedGroup;
   }
 
+  get userGroupId(): string | undefined {
```
smadbe, line 43: probably not useful anymore with the previous remark.

smadbe (changes requested):
👍

src/app/core/components/top-right-menu/top-right-menu.component.ts
```diff
@@ -13,7 +13,7 @@ import { UserSessionService } from 'src/app/shared/services/user-session.service
 export class TopRightMenuComponent {
 
   readonly menuItems = [
-    { label: 'Profile', icon: 'pi pi-user', routerLink: [ 'groups', 'me' ] },
+    { label: 'Profile', icon: 'pi pi-user', routerLink: [ 'groups', 'users', this.sessionService.userGroupId, 'personal-data' ] },
```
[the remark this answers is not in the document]
Iloveall: Fixed

src/app/modules/group/pages/group-user/group-user.component.ts
```diff
@@ -1,24 +1,80 @@
...
+import { CurrentContentService } from '../../../../shared/services/current-content.service';
+import { UserSessionService } from '../../../../shared/services/user-session.service';
 
 @Component({
   selector: 'alg-group-user',
   templateUrl: './group-user.component.html',
   styleUrls: [ './group-user.component.scss' ]
 })
-export class GroupUserComponent {
+export class GroupUserComponent implements OnInit, OnDestroy {
```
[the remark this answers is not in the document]
Iloveall: Fixed

src/app/modules/group/pages/group-user/group-user.component.html
```diff
@@ -21,5 +21,42 @@
...
+          routerLinkActive #personalData="routerLinkActive"
+          [routerLinkActiveOptions]="{exact: true}"
+          [class.active]="personalData.isActive"
+          i18n
+      >
+        Personal data
+      </a>
+    </nav>
+
+    <ng-container *ngIf="progress?.isActive || !progress && (currentUserGroupId$ | async) !== user.groupId">
```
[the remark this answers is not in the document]
Iloveall: It's changed in feature/group-user-list-progress

src/app/modules/group/pages/group-user/group-user.component.ts
Iloveall: I have found this in our app, I suppose async pipe, in template, doesn't work under *ngIf, delay(0) (which is async) moves this operation to end of stack operations and makes it work under *ngIf.

Iloveall: So this is just a trick

src/app/modules/group/pages/group-user/group-user.component.ts
Iloveall: Fixed

src/app/modules/group/pages/group-user/group-user.component.ts
Iloveall: Fixed

src/app/shared/services/user-session.service.ts
Iloveall: Fixed

src/app/modules/group/pages/group-user/group-user.component.html
Iloveall: Fixed

src/app/modules/group/pages/group-user/group-user.component.ts
smadbe: As in group-by-id.component.ts:72, when you leave a page which has set a current content (to be displayed on the top), you need to set it to null, so `this.currentContent.current.next(null);`

Iloveall: Fixed

smadbe (changes requested):
The breadcrumbs does not update when changing tab.
Screenshot 2021-07-12 at 12 56 50
You may need to combine the state with the routing (for instance with `combineLatest`) or with other ways to know the current tab.

src/app/modules/group/pages/user/user.component.ts
```diff
@@ -0,0 +1,73 @@
...
+            path: [
+              {
+                title: state.isFetching || state.isError ? '...' : formatUser(state.data),
+                navigateTo: this.router.createUrlTree([ 'groups', 'users', this.route.snapshot.params.id ]),
+              },
+              {
+                title: this.router.url.includes('personal-data') ? $localize`Personal info` : $localize`Progress`,
+              }
+            ],
+            currentPageIdx: -1,
```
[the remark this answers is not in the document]
Iloveall: Fixed

src/app/modules/group/pages/user/user.component.ts
```diff
@@ -0,0 +1,73 @@
...
+        map(state => contentInfo({
+          breadcrumbs: {
+            category: $localize`Users`,
+            path: [
+              {
+                title: state.isFetching || state.isError ? '...' : formatUser(state.data),
+                navigateTo: this.router.createUrlTree([ 'groups', 'users', this.route.snapshot.params.id ]),
+              },
+              {
+                title: this.router.url.includes('personal-data') ? $localize`Personal info` : $localize`Progress`,
```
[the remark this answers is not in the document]
smadbe: you may only be able to see the user name and basic info without progress nor personal info. In this case, there is no reason to put in the breadcrumb either `Personal info` or `Progress`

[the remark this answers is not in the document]
Iloveall: We decided we can do the change later in the frontend

smadbe (approved):
I haven't re-reviewed everything but all my remarks are fixed.

PR #472, opened 2021-06-14
Typed forms 

GabrielVidal1:
Add strongly typed forms for the group edit page using the [@ngneat/reactive-forms](https://www.npmjs.com/package/@ngneat/reactive-forms) package.
I made changed for the group edit page to test the new typed forms workflow for now, but this could be extended to the item edit page in this PR.


smadbe:
Probably it is not longer useful to integrate typed forms through an external package as the Angular team is working on a solution for that: https://github.com/angular/angular/discussions/44513 , probably it will included in one of the next versions.

PR #473, opened 2021-06-14
Add task-xd-pr to communicate with tasks

mblockelet:
Add a typescript version of [task-xd-pr from pem-platform](https://github.com/France-ioi/pem-platform/blob/master/task-xd-pr.js), which allows communication between platforms and tasks. It requires jschannel.


mblockelet:
Replaced by #525.

PR #478, opened 2021-06-14
Hides pending request and add subgroups panels for none membership

### LINKED ISSUE #444: Show appropriate components for manager with 'none' level
There are 3 levels of group management: none, memberships, and memberships_and_group. (a user may also not be a manager at all).

When the current user is not a manager with `memberships` or `memberships_and_group`:
- do not show/load pending request panel (with error with 403 currently)
- do not show the "Add subgroups" section under the the list of subgroups/members.

Iloveall:
Fixes #444

### DIFF
--- a/src/app/modules/group/pages/group-details/group-details.component.html
+++ b/src/app/modules/group/pages/group-details/group-details.component.html
@@ -3,7 +3,7 @@
   <ng-container *ngIf="state.isReady && state.data as group">
     <alg-group-header [group]="group"></alg-group-header>
     <alg-pending-join-requests
-        *ngIf="group.isCurrentUserManager && group?.currentUserCanManage !== 'none'"
+        *ngIf="group.isCurrentUserManager && group.currentUserCanManage !== 'none'"
         [groupId]="group.id"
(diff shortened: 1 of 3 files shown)


src/app/modules/group/pages/group-details/group-details.component.html
```diff
@@ -2,7 +2,10 @@
 
   <ng-container *ngIf="state.isReady && state.data as group">
     <alg-group-header [group]="group"></alg-group-header>
-    <alg-pending-join-requests *ngIf="group.isCurrentUserManager" [groupId]="group.id"></alg-pending-join-requests>
+    <alg-pending-join-requests
+        *ngIf="group.isCurrentUserManager && group?.currentUserCanManage !== 'none'"
```
smadbe, line 6: I don't think `group` is optional.

smadbe (changes requested):
I forgot in the spec: could you remove the "remove" button as well when the user has "none" level? 
Screenshot 2021-07-05 at 12 36 28

smadbe (approved):

PR #486, opened 2021-06-15
Changes max items display for group indicator

### LINKED ISSUE #445: Bug in "Indicate why you can see a group"
From #373 , if there are 4 groups in the list, you should display the full list.
Screenshot 2021-05-28 at 11 24 09

In summary: if <= 4, display them all; otherwise display the first three "and x more ...".
So you should never see "and 1 more" !

Iloveall:
Fixes #445

### DIFF
--- a/src/app/modules/group/components/group-links/group-links.component.html
+++ b/src/app/modules/group/components/group-links/group-links.component.html
@@ -14,6 +14,6 @@
     >, and</span>
   </span>
   <span class="more" *ngIf="(items || []).length > maxItemsDisplay && isLast" i18n>
-    {{ (items || []).length - ((items || []).length > maxItemsDisplay ? maxItemsDisplay - 1 : maxItemsDisplay) }} more
+    {{ (items || []).length - maxItemsDisplay + 1 }} more
   </span>
 </ng-container>
--- a/src/app/modules/group/components/group-links/group-links.component.html
+++ b/src/app/modules/group/components/group-links/group-links.component.html
@@ -1,4 +1,5 @@
-<ng-container *ngFor="let item of items | slice : 0 : maxItemsDisplay; let i = index; last as isLast;">
+<ng-container
+    *ngFor="let item of items | slice : 0 : (items || []).length > maxItemsDisplay ? maxItemsDisplay - 1 : maxItemsDisplay; let i = index; last as isLast;">
   <a
     class="alg-link base-color"
     (click)="onButtonClick(item)"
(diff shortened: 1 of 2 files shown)


smadbe (changes requested):
Not really. The request is:
- 1 group -> display 1 group
- 2 groups -> display 2 groups
- 3 groups -> display 3 groups
- 4 groups -> display 4 groups
- 5 groups -> display 3 groups "and 2 more..."
- 6 groups -> display 3 groups "and 3 more..."
-  ...

So "and 1 more" should not be possible, it cannot be just changing a constant. I supposed (I haven't checked) 5 groups will be displayed as  4 groups and 1 more...

src/app/modules/group/components/group-links/group-links.component.html
```diff
@@ -13,6 +14,6 @@
     >, and</span>
   </span>
   <span class="more" *ngIf="(items || []).length > maxItemsDisplay && isLast" i18n>
-    {{ (items || []).length - maxItemsDisplay }} more
+    {{ (items || []).length - ((items || []).length > maxItemsDisplay ? maxItemsDisplay - 1 : maxItemsDisplay) }} more
```
smadbe, line 17: `(items || []).length > maxItemsDisplay ? ... : ...`  aren't you in case where this is always true?

Iloveall: Items (groups) can be 1, can be 6 and more, maxItemsDisplay equal 4. So I guess it shouldn't be always true. I have tested different cases

smadbe: line 16 is ` *ngIf="(items || []).length > maxItemsDisplay &&  ...` 🤨

Iloveall: That condition does:

If groups count more than 4, then need to show "more" caption after last group, but if there are less or equal 4 groups, then no need to show "more" caption.

smadbe: If line 16 is that L17 is displayed if "`(items || []).length > maxItemsDisplay` and something... we are sure on line 17 that `(items || []).length > maxItemsDisplay` is true... so the ternary operation `(items || []).length > maxItemsDisplay ? ... : ...` is useless (it is always true).

I don't try to get the business logic, just programming logic.

Iloveall: Got it, Fixed

smadbe (approved):

PR #492, opened 2021-06-19
Add layout service, sync headers display with menu

mblockelet:
Add a layout service allowing components to act on the layout (currently whether the left menu and headers are opened). This will be used by the item-display component, to close the left menu and headers automatically when displaying a task.

Also sync the display of the headers with the display of the left menu, as decided, to reduce the number of clicks.

### DIFF
--- a/src/app/core/app.component.html
+++ b/src/app/core/app.component.html
@@ -40,7 +40,7 @@
     <div
       class="right"
       *ngrxLet="leftMenuAndHeaders$ as leftMenuAndHeaders"
-      [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders, 'with-mode-bar': currentMode !== 'normal' }"
+      [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders, 'with-mode-bar': currentMode !== 'normal', 'with-task': withTask$ | async }"
       >
       <div class="main-content">
         <router-outlet></router-outlet>
--- a/src/app/core/app.component.html
+++ b/src/app/core/app.component.html
@@ -40,7 +40,7 @@
     <div
       class="right"
       *ngrxLet="fullFrameContent$ as fullFrameContent"
-      [ngClass]="{ collapsed: !fullFrameContent, expanded: fullFrameContent, 'with-mode-bar': currentMode !== 'normal', 'with-footer': footer$ | async }"
+      [ngClass]="{ collapsed: !fullFrameContent, expanded: fullFrameContent, 'with-mode-bar': currentMode !== 'normal', 'with-footer': contentFooter$ | async }"
       >
       <div class="main-content">
         <router-outlet></router-outlet>
--- a/src/app/core/app.component.html
+++ b/src/app/core/app.component.html
@@ -1,22 +1,19 @@
 <div class="container" *ngrxLet="session$ as session">
-  <div class="left" [ngClass]="{ collapsed: !leftMenuDisplayed, expanded: leftMenuDisplayed }">
-    <alg-left-menu (hideLeftMenu)="toggleLeftMenuDisplay(false)"></alg-left-menu>
+  <div class="left" [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders }">
+    <alg-left-menu></alg-left-menu>
   </div>
   <ng-container *ngIf="currentMode$ | async as currentMode">
-    <div class="top-bar" [ngClass]="{ 'with-mode-bar': currentMode !== 'normal', collapsed: !leftMenuDisplayed, expanded: leftMenuDisplayed }">
-      <alg-left-header
-        *ngIf="!leftMenuDisplayed"
-        (displayLeftMenu)="toggleLeftMenuDisplay($event)"
-        (displayHeaders)="toggleHeadersDisplay($event)"
-        [compactMode]="true"
-      ></alg-left-header>
+    <div class="top-bar"
+      [ngClass]="{ 'with-mode-bar': currentMode !== 'normal', collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders }">
+      <alg-left-header *ngIf="!leftMenuAndHeaders" [compactMode]="true"></alg-left-header>
       <div class="top-right-bar" *ngrxLet="currentContent$ as currentContent">
         <alg-editor-bar *ngIf="currentMode === 'editing'" (cancel)="onEditCancel()"></alg-editor-bar>
-        <alg-observation-bar *ngIf="currentMode === 'watching'" [groupName]="session?.watchedGroup?.name || ''" (cancel)="onWatchCancel()"></alg-observation-bar>
-        <div class="breadcrumb-bar" *ngIf="headersDisplayed && !scrolled">
+        <alg-observation-bar *ngIf="currentMode === 'watching'" [groupName]="session?.watchedGroup?.name || ''"
+          (cancel)="onWatchCancel()"></alg-observation-bar>
+        <div class="breadcrumb-bar" *ngIf="leftMenuAndHeaders && !scrolled">
           <alg-breadcrumb [contentBreadcrumb]="currentContent?.breadcrumbs"></alg-breadcrumb>
         </div>
-        <div class="folded-top-bar" *ngIf="!headersDisplayed || scrolled">
+        <div class="folded-top-bar" *ngIf="!leftMenuAndHeaders || scrolled">
           <!--<alg-score-ring
             [displayedScore]="50"
             [currentScore]="60"
@@ -37,8 +34,9 @@
       </div>
       <alg-top-right-controls [drawLeftBorder]="currentMode === 'normal'"></alg-top-right-controls>
     </div>
-    <div class="right" [ngClass]="{ collapsed: !leftMenuDisplayed, expanded: leftMenuDisplayed, 'with-mode-bar': currentMode !== 'normal' }">
-      <div class="main-content" [ngClass]="{ folded: !headersDisplayed, unfolded: headersDisplayed }">
+    <div class="right"
+      [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders, 'with-mode-bar': currentMode !== 'normal' }">
+      <div class="main-content" [ngClass]="{ folded: !leftMenuAndHeaders, unfolded: leftMenuAndHeaders }">
         <router-outlet></router-outlet>
       </div>
     </div>
@@ -47,16 +45,10 @@
 <p-toast></p-toast>
 <p-confirmDialog></p-confirmDialog>
 <p-confirmPopup key="commonPopup"></p-confirmPopup>
-<p-dialog
-  *ngrxLet="fatalError$"
-  i18n-header="@@auth-error-modal-title" header="Error"
-  [visible]="true"
-  [style]="{width: '50vw'}"
-  [modal]="true"
-  [closeOnEscape]="false"
-  [closable]="false"
->
-  <p i18n="@@auth-error-modal-context">Oops, we are unable to make the site work properly. Are you connected to the Internet?</p>
+<p-dialog *ngrxLet="fatalError$" i18n-header="@@auth-error-modal-title" header="Error" [visible]="true"
+  [style]="{width: '50vw'}" [modal]="true" [closeOnEscape]="false" [closable]="false">
+  <p i18n="@@auth-error-modal-context">Oops, we are unable to make the site work properly. Are you connected to the
+    Internet?</p>
   <p i18n="@@auth-error-modal-todo">Try to refresh the page. If this problem persists, please contact us.</p>
 </p-dialog>
 <alg-language-mismatch></alg-language-mismatch>
--- a/src/app/core/app.component.html
+++ b/src/app/core/app.component.html
@@ -1,10 +1,13 @@
 <div class="container" *ngrxLet="session$ as session">
-  <div class="left" [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders }">
+  <div class="left" *ngrxLet="leftMenuAndHeaders$ as leftMenuAndHeaders" [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders }">
     <alg-left-menu></alg-left-menu>
   </div>
   <ng-container *ngIf="currentMode$ | async as currentMode">
-    <div class="top-bar"
-      [ngClass]="{ 'with-mode-bar': currentMode !== 'normal', collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders }">
+    <div
+      class="top-bar"
+      *ngrxLet="leftMenuAndHeaders$ as leftMenuAndHeaders"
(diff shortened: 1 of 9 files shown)


mblockelet:
I don't know why the e2e tests are failing on CircleCI, they work fine for me locally...

smadbe:
> I don't know why the e2e tests are failing on CircleCI, they work fine for me locally...

I've relaunched them and they pass.

src/app/core/components/left-header/left-header.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="nav-header" *ngIf="!compactMode">
   <div class="nav-control">
-    <span class="nav-collapse" (click)="hideLeftMenu()" tabindex="0">
+    <span class="nav-collapse" (click)="hideLeftMenuAndHeaders()" tabindex="0">
       <!-- <i class="fa fa-bars"></i> -->
       <img src="assets/images/menu-open.svg" style="width: 2rem; height: 2rem;" />
     </span>
```
smadbe, line 6: If now it is just a full screen button, let's just use a full screen icon?
https://fontawesome.com/v5.15/icons/expand-alt?style=solid or https://fontawesome.com/v5.15/icons/expand-arrows-alt?style=solid
and https://fontawesome.com/v5.15/icons/compress-alt?style=solid

src/app/core/app.component.html
```diff
@@ -47,16 +45,10 @@
...
-  [style]="{width: '50vw'}"
-  [modal]="true"
-  [closeOnEscape]="false"
-  [closable]="false"
->
-  <p i18n="@@auth-error-modal-context">Oops, we are unable to make the site work properly. Are you connected to the Internet?</p>
+<p-dialog *ngrxLet="fatalError$" i18n-header="@@auth-error-modal-title" header="Error" [visible]="true"
+  [style]="{width: '50vw'}" [modal]="true" [closeOnEscape]="false" [closable]="false">
+  <p i18n="@@auth-error-modal-context">Oops, we are unable to make the site work properly. Are you connected to the
+    Internet?</p>
```
smadbe, line 51: Why is this change? The previous version follows the way we display multiline html.

src/app/core/app.component.ts
```diff
@@ -26,8 +27,8 @@ export class AppComponent implements OnInit, OnDestroy {
     this.localeService.currentLangError$,
   );
 
-  leftMenuDisplayed = true;
-  headersDisplayed = true;
+  leftMenuAndHeaders = true;
+  leftMenuAndHeadersSubscription? : Subscription;
```
smadbe, line 31: Is there a reason for keeping a extra copy of this state? 
You should just use the observable directly
```ts
leftMenuAndHeaders$ = this.layoutService.leftMenuAndHeadersDisplayed$;
```
and use it with the `async` pipe in the code.

src/app/core/app.component.html
```diff
@@ -37,8 +34,9 @@
       </div>
       <alg-top-right-controls [drawLeftBorder]="currentMode === 'normal'"></alg-top-right-controls>
     </div>
-    <div class="right" [ngClass]="{ collapsed: !leftMenuDisplayed, expanded: leftMenuDisplayed, 'with-mode-bar': currentMode !== 'normal' }">
-      <div class="main-content" [ngClass]="{ folded: !headersDisplayed, unfolded: headersDisplayed }">
+    <div class="right"
```
smadbe, line 37: In this project:
If it fits on one line:
```<div class="right" [ngClass]=...>```
Otherwise:
```
<div 
  class="right" 
  [ngClass]=...
>
```

This remarks applies everywhere in this file.

File: src/app/core/app.component.html [hunk 5]
```diff
@@ -37,8 +34,9 @@
...
+      [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders, 'with-mode-bar': currentMode !== 'normal' }">
+      <div class="main-content" [ngClass]="{ folded: !leftMenuAndHeaders, unfolded: leftMenuAndHeaders }">
```
smadbe, line 39: I suppose it should inherit from the parent div with expanded/collapsed instead of defining 2 conditions again here.

src/app/shared/services/layout.service.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  toggleLeftMenuAndHeaders(shown?: boolean): void {
+    if (shown !== undefined) {
+      this.leftMenuAndHeadersDisplayed.next(shown);
+    } else {
+      this.leftMenuAndHeadersDisplayed.next(!this.leftMenuAndHeadersDisplayed.value);
+    }
+  }
+
+  toggleWithTask(newState: boolean): void {
```
smadbe, line 25: As you do not use it in the PR, I do not know what you want to do with it exactly but it looks a bit strange to have to independent observable for things which are not independent.

src/app/shared/services/layout.service.ts
```diff
@@ -0,0 +1,28 @@
...
+  leftMenuAndHeadersDisplayed$ = this.leftMenuAndHeadersDisplayed.asObservable().pipe(delay(0));
+
+  private withTask = new BehaviorSubject<boolean>(false);
+  withTask$ = this.withTask.asObservable().pipe(delay(0));
+
+  toggleLeftMenuAndHeaders(shown?: boolean): void {
+    if (shown !== undefined) {
+      this.leftMenuAndHeadersDisplayed.next(shown);
+    } else {
+      this.leftMenuAndHeadersDisplayed.next(!this.leftMenuAndHeadersDisplayed.value);
```
smadbe, line 21: I see only usage with a given parameter so I does not look relevant case.

smadbe (changes requested):

src/app/core/app.component.html
mblockelet: Do you have a suggestion of HTML formatter which does that?

src/app/core/app.component.html
mblockelet: Done, thanks.

src/app/core/app.component.html
mblockelet: Sorry, default HTML formatter had modified this and I hadn't noticed it.

src/app/core/app.component.ts
mblockelet: Thanks. I used `*ngrxLet` in the template as I saw another line doing this, and it avoids adding a lot of async pipes.

src/app/core/components/left-header/left-header.component.html
mblockelet: Used the expand-arrows-alt and compress-arrows-alt, thanks for the suggestion.

src/app/shared/services/layout.service.ts
mblockelet: Removed.

src/app/shared/services/layout.service.ts
mblockelet: Added usage of withTask$ : removes padding-bottom in the right side of app.component.html when a task is loaded, so the task can be displayed with all available area.

src/app/core/app.component.html
smadbe: No, sorry. I do it by hand.

src/app/core/app.component.html
```diff
@@ -1,10 +1,13 @@
...
-  <div class="left" [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders }">
+  <div class="left" *ngrxLet="leftMenuAndHeaders$ as leftMenuAndHeaders" [ngClass]="{ collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders }">
     <alg-left-menu></alg-left-menu>
   </div>
   <ng-container *ngIf="currentMode$ | async as currentMode">
-    <div class="top-bar"
-      [ngClass]="{ 'with-mode-bar': currentMode !== 'normal', collapsed: !leftMenuAndHeaders, expanded: leftMenuAndHeaders }">
+    <div
+      class="top-bar"
+      *ngrxLet="leftMenuAndHeaders$ as leftMenuAndHeaders"
```
smadbe, line 8: Just use `leftMenuAndHeaders` defined above in this block, no need for another subscription (`ngrxlet`)

src/app/shared/services/layout.service.ts
```diff
@@ -0,0 +1,24 @@
...
+  providedIn: 'root'
+})
+export class LayoutService {
+  // Service allowing modifications of the layout
+
+  private leftMenuAndHeadersDisplayed = new BehaviorSubject<boolean>(true);
+  leftMenuAndHeadersDisplayed$ = this.leftMenuAndHeadersDisplayed.asObservable().pipe(delay(0));
+
+  private withTask = new BehaviorSubject<boolean>(false);
+  withTask$ = this.withTask.asObservable().pipe(delay(0));
```
smadbe, line 15: Do you use `.pipe(delay(0))` because it was required or because you've copied it for another place?

src/app/core/app.component.ts
```diff
@@ -28,6 +28,7 @@ export class AppComponent implements OnInit, OnDestroy {
   );
 
   leftMenuAndHeaders$ = this.layoutService.leftMenuAndHeadersDisplayed$;
+  withTask$ = this.layoutService.withTask$;
```
smadbe, line 31: Don't you want the same for course? If so, using "task" in the name is not very appropriate.

src/app/core/components/left-header/left-header.component.html
```diff
@@ -12,6 +11,6 @@
 
 <div class="nav-control alt" *ngIf="compactMode">
   <span class="nav-collapse" (click)="showLeftMenuAndHeaders()" tabindex="0">
-    <img src="assets/images/menu-open.svg" style="width: 2rem; height: 2rem;" />
+    <i class="fas fa-compress-arrows-alt" style="font-size: 2rem;"></i>
```
smadbe, line 14: This one looks a bit like the sign for emergency rendez-vous point. ;-)

src/app/core/components/left-header/left-header.component.html
```diff
@@ -1,8 +1,7 @@
 <div class="nav-header" *ngIf="!compactMode">
   <div class="nav-control">
     <span class="nav-collapse" (click)="hideLeftMenuAndHeaders()" tabindex="0">
-      <!-- <i class="fa fa-bars"></i> -->
-      <img src="assets/images/menu-open.svg" style="width: 2rem; height: 2rem;" />
+      <i class="fas fa-expand-arrows-alt" style="font-size: 2rem;"></i>
```
smadbe, line 4: As the it extends the right part, probably it would be better if it would be on the right, and not in the part which disappears. (can be done later)

smadbe (changes requested):

src/app/core/app.component.ts
mblockelet: TODO : Rename leftMenuAndHeaders to "fullFrameContent", "withTask" to "footer"

src/app/core/components/left-header/left-header.component.html
mblockelet: TODO : Move the icon to the right of the title "Algorea Platform", make it a left/right chevron.

src/app/core/app.component.html
mblockelet: They're not actually in the same block, so if I remove it it stops working.

src/app/shared/services/layout.service.ts
mblockelet: Yes.

src/app/core/app.component.ts
mblockelet: Renamed.

src/app/core/components/left-header/left-header.component.html
mblockelet: Changed icon.

src/app/core/components/left-header/left-header.component.html
mblockelet: Changed icon (to `<<`) and position of the button.

image

image

Other choices considered :

`fa-chevron-left` : I prefer the other one when comparing

image

`fa-caret-square-left` : looks too heavy to me

image

src/app/shared/services/layout.service.ts
```diff
@@ -8,17 +8,17 @@ import { delay } from 'rxjs/operators';
...
   // Service allowing modifications of the layout
 
-  private leftMenuAndHeadersDisplayed = new BehaviorSubject<boolean>(true);
-  leftMenuAndHeadersDisplayed$ = this.leftMenuAndHeadersDisplayed.asObservable().pipe(delay(0));
+  private fullFrameContent = new BehaviorSubject<boolean>(true);
+  fullFrameContent$ = this.fullFrameContent.asObservable().pipe(delay(0));
 
-  private withTask = new BehaviorSubject<boolean>(false);
-  withTask$ = this.withTask.asObservable().pipe(delay(0));
+  private footer = new BehaviorSubject<boolean>(false);
```
smadbe, line 14: contentFooter maybe?
And I assume "footer false" should mean that the footer is not shown. So as you changed withTask -> footer, I would expect you invert the logic (so the default value).  Or you have to call the attribute `noContentFooter`, which is valid as well (double negation is not great though)

src/app/shared/services/layout.service.ts
```diff
@@ -8,17 +8,17 @@ import { delay } from 'rxjs/operators';
...
+  footer$ = this.footer.asObservable().pipe(delay(0));
```
smadbe, line 15: I think it would be useful to add one sentence of comment for `fullFrameContent` and `*footer` so that it is clear what is the intend of these.

src/app/core/components/left-header/left-header.component.html
```diff
@@ -1,16 +1,16 @@
...
+    <span class="nav-collapse" (click)="setFullFrameContent()" tabindex="0">
+      <i class="fas fa-angle-double-left"></i>
+    </span>
   </div>
 </div>
 
 <div class="nav-control alt" *ngIf="compactMode">
   <span class="nav-collapse" (click)="unsetFullFrameContent()" tabindex="0">
-    <i class="fas fa-compress-arrows-alt" style="font-size: 2rem;"></i>
+    <i class="fas fa-angle-double-right"></i>
```
smadbe, line 14: When testing it, I think I prefer `fa-chevron-*` or `fa-chevron-circle-*` as the chevron is more commonly used for hiding/showing menu than the double chevron.
Screenshot 2021-07-20 at 09 50 13
Screenshot 2021-07-20 at 09 49 58

smadbe (changes requested):
Small details

src/app/shared/services/layout.service.ts
mblockelet: I had changed the logic to fit the new name "footer" (if true, adds the footer margin). I just renamed it to contentFooter which is a better name.

This comment made me realize that I didn't update the logic when renaming `leftMenuAndHeaders` to `fullFrameContent`, so I just fixed that.

src/app/shared/services/layout.service.ts
mblockelet: Added comments on the public observable and function.

src/app/core/components/left-header/left-header.component.html
mblockelet: Changed to `fa-chevron-circle`.

smadbe (approved):
👍

PR #500, opened 2021-06-22
Fixes links for groups in manager list component

### LINKED ISSUE #497: In manager list: fix links to groups
In group manager list, all managers are not users while they might be groups. In this case the link is not valid.
Now the service returns the first-/lastname and login. 

If there is a login, you know that the manager is a user.

If it is a user, display the usual user display and a link to the user page.
If if is not, display the group name and a link to the group page.

(to test with a group which is a manager: http://dev.algorea.org/en/#/groups/by-id/6884210558506774905;path=/details/managers)

Iloveall:
Fixes #497

### DIFF
--- a/src/app/modules/group/components/group-manager-list/group-manager-list.component.html
+++ b/src/app/modules/group/components/group-manager-list/group-manager-list.component.html
@@ -64,26 +64,26 @@
             </a>
           </td>
           <td style="min-width: 110px;">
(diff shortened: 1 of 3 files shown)


smadbe (changes requested):
Ok for the link but I think you missed the "If it is a user, display the usual user display and a link to the user page.".

smadbe (approved):

PR #502, opened 2021-07-05
Item edition: allow "0" entry_max_team_size when team participation is not enabled

### LINKED ISSUE #493: Item edition: allow "0" entry_max_team_size when team participation is not enabled
We had a bug: most editable items cannot be saved, indicating that we need to solve the issue in red first... while there were nothing in red.
After some investigation, it was because team size was set to "0" (the db default value, and because this item was not supporting teams anyway)... and the red was only shown when the "enable team participation" was on...

So the main issue to be fixed here is: 
We should not disallow "0" for max team size when team participation is disabled. The cleanest way is probably to have a validator which check that "entry_max_team_size > 0 when team participation is enabled"

ThomasJuster:
Fixes #493

### DIFF
--- a/src/app/modules/item/pages/item-edit/item-edit.component.ts
+++ b/src/app/modules/item/pages/item-edit/item-edit.component.ts
@@ -1,6 +1,6 @@
 import { Component, OnDestroy, ViewChild } from '@angular/core';
 import { ItemDataSource } from '../../services/item-datasource.service';
-import { AbstractControl, FormBuilder, Validators } from '@angular/forms';
+import { AbstractControl, FormBuilder, ValidatorFn, Validators } from '@angular/forms';
 import { forkJoin, Observable, of, Subscription, throwError } from 'rxjs';
 import { map, switchMap } from 'rxjs/operators';
 import { ItemStringChanges, UpdateItemStringService } from '../../http-services/update-item-string.service';
@@ -48,6 +48,8 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
     entry_frozen_teams: [ false ],
     entry_max_team_size: [ '' ],
     entry_min_admitted_members_ratio: [ '' ],
+  }, {
+    validators: [ this.maxTeamSizeValidator() ],
   });
   itemChanges: { children?: ChildData[] } = {};
 
@@ -55,7 +57,6 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
   initialFormData?: Item & {durationEnabled?: boolean, enteringTimeMaxEnabled?: boolean};
 
   subscription?: Subscription;
-  maxTeamSizeSubscription?: Subscription;
 
   get enableParticipation(): boolean {
     return this.initialFormData?.type !== 'Skill';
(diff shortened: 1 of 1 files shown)


src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -86,11 +87,20 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
         this.initialFormData = data;
         this.resetForm();
       });
+
+    this.maxTeamSizeSubscription = this.itemForm
+      .get('entry_participant_type')
+      ?.valueChanges
```
smadbe, line 93: usually we display these on a single line
```suggestion
      .get('entry_participant_type')?.valueChanges
```

smadbe (approved):
Ideally, I would have preferred a validation of the team field-set as a whole instead of another subscription which makes things hard to follow (and to debug). But probably your change is quite short and so more pragmatic.

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -368,4 +370,16 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
...
+    return (itemForm): null => {
+      const isParticipationAsTeamOnly = itemForm.get('entry_participant_type')?.value as boolean;
+      const maxTeamSizeControl = itemForm.get('entry_max_team_size');
+      if (!maxTeamSizeControl) return null;
+
+      const errors = isParticipationAsTeamOnly ? Validators.min(1)(maxTeamSizeControl) : null;
+      maxTeamSizeControl.setErrors(errors);
+      return null;
+    };
+  }
```
ThomasJuster, line 384: I would have preferred a fieldset level validator too, it would look like this. If it's ok with you, we can leave it as is.

However, the trouble with this approach is that I need to manually attach the error to the max team size control instead of returning the validation errors (which would attach them to the form group control), and that is a side effect to me.
And the reason why I need to attach the error to the max_team_size control is because it's currently the only/preferred way to monitor errors on the ui.

src/app/modules/item/pages/item-edit/item-edit.component.ts
```diff
@@ -368,4 +370,16 @@ export class ItemEditComponent implements OnDestroy, PendingChangesComponent {
...
   }
+
+  private maxTeamSizeValidator(): ValidatorFn {
+    return (itemForm): null => {
+      const isParticipationAsTeamOnly = itemForm.get('entry_participant_type')?.value as boolean;
+      const maxTeamSizeControl = itemForm.get('entry_max_team_size');
+      if (!maxTeamSizeControl) return null;
+
+      const errors = isParticipationAsTeamOnly ? Validators.min(1)(maxTeamSizeControl) : null;
+      maxTeamSizeControl.setErrors(errors);
```
smadbe, line 381:
```suggestion
      if (isParticipationAsTeamOnly) maxTeamSizeControl.setErrors(Validators.min(1)(maxTeamSizeControl));
```
would have be more readable from my point of view

smadbe (approved):
That looks much better to me.
The "side effect" is not so problematic IMO.

src/app/modules/item/pages/item-edit/item-edit.component.ts
ThomasJuster: this sample doesn't cover the case when switching from participation as team to the opposite, but I get the idea, I'll change it accordingly.

ThomasJuster: Fixed

PR #508, opened 2021-07-05
fix problem with auth by cookie when cross site

### LINKED ISSUE #507: Fix cookie problem on cross site

ThomasJuster:
Fixes #507

## Solution

Under the hood, Angular uses `XMLHttpRequest`. When the app is configured to be deployed under another domain, setting [withCredentials](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) option to `true` allows the server to set third party cookies.
The app configuration properties which pilot this behavior are `authType` and `sameSite`. When `authType === 'cookies' && sameSite === false`, credentials are enabled.

If authentication still doesn't work for algorea apps with another domain, it means that the user's browser is configured to block third party cookies, s⋅he will have to change to browser configuration.


ThomasJuster:
Close in favor of #522

PR #512, opened 2021-07-06
Fixes permissions indication

### LINKED ISSUE #505: Bugfix: Wrong manager permission indication
Investigate why "X" are indicated for "can grant group access" and "can watch members" while the service returns "true". Fix it.

Screenshot 2021-07-05 at 16 14 41

Iloveall:
Fixes #505


smadbe (approved):
🤷‍♂️

PR #522, opened 2021-07-07
auth: force 'same-site' cookie option

### LINKED ISSUE #507: Fix cookie problem on cross site

ThomasJuster:
Fixes #507 … by not fixing it actually.

### DIFF
--- a/src/app/shared/helpers/config.ts
+++ b/src/app/shared/helpers/config.ts
@@ -1,7 +1,7 @@
 import { environment } from 'src/environments/environment';
 
 interface TokenConfig { authType: 'tokens' }
-type CookieConfig = { authType: 'cookies' } & ({ secure: true, sameSite: boolean } | { secure: boolean, sameSite: true });
+interface CookieConfig { authType: 'cookies', secure: boolean }
 export type AuthTypeConfig = TokenConfig | CookieConfig;
 
 export interface LanguageConfig {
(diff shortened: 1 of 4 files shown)


src/app/shared/http-services/auth.http-service.ts
```diff
@@ -23,7 +23,14 @@ export class AuthHttpService {
...
     use_cookie: '1',
     cookie_secure: appConfig.secure ? '1' : '0',
-    cookie_same_site: appConfig.sameSite ? '1' : '0',
+    /**
+     * NOTE on cookie_same_site:
+     * The 'same-site' policy only applies for auth by third-party cookie.
+     * Safari and Firefox both disable third-party cookies by default, and Chrome will in 2021 (but postponed for 2023)
+     * Since third-party cookies will be removed in a near future, we chose not to enable this type of authentication,
+     * making the 'same-site' policy always true.
+     */
```
smadbe, line 32: 👍

src/app/shared/helpers/config.ts
```diff
@@ -1,7 +1,7 @@
 import { environment } from 'src/environments/environment';
 
 interface TokenConfig { authType: 'tokens' }
-type CookieConfig = { authType: 'cookies' } & ({ secure: true, sameSite: boolean } | { secure: boolean, sameSite: true });
+interface CookieConfig { authType: 'cookies', secure: boolean }
```
smadbe, line 4: Hum... this `secure` is just whether the backend is in https or not. 
We should probably move `authType` to `Environment` directly and compute `cookie_secure` in code.

Can be done in another PR.

smadbe (approved):
👍 
Small suggestion of improvement but I approve this PR so that you can do it in another PR if you prefer.

ThomasJuster:
Close in favor of #531

PR #524, opened 2021-07-07
handle can_edit children item permission

### LINKED ISSUE #506: Item edition: handle correctly "can_edit:children" permission
## Motivations
The `can_edit` permission given to groups on item (see [doc](https://france-ioi.github.io/algorea-devdoc/design/access-rights/items/#can_edit)) has a "children" option which allows users to edit items, but only to modify children on this items (so typically only for chapters/skills). 

For the moment, for this case, the UI display the edit button but shows a error message when pressing it. 
(for instance, with the usual user on item [1625159049301502151](http://dev.algorea.org/en/#/activities/by-id/1625159049301502151;attempId=0;path=4702/details)).

## Subtasks
* Do not display the error like now
* Only allow changing children if the user has 'children' as can edit permissions.
* Test extensively.

ThomasJuster:
Fixes #506


smadbe (approved):
Most of the PR is not about "Fixes #506" but that's good 👍

PR #525, opened 2021-07-10
Convert task-xd-pr to RxJS

mblockelet:
This PR :
* adds a layer over jschannel to use RxJS Observables instead of callbacks
* simplifies task-xd-pr by removing unused functions and using RxJS logic 
* resolves comments from #473 
* modifies item-display *minimally* to be compatibles

Things out of scope of this PR :
* better typing and type validation will come in a separate PR (#526)
* item-display changes were minimal to make it work with the new version ; a better RxJS logic will be implemented along with the changes originally in the item-display-2 branch


src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+
+export class Platform {
+  task : Task;
+  constructor(task : Task) {
+    this.task = task;
+  }
+
+  getTask() : Task {
+    return this.task;
+  }
```
smadbe, line 254: A priori (I don't have the full picture), I would say a listener should not have the reference of the object it is listening to. Is it used ?

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+
+  constructor(chan: RxMessagingChannel) {
+    this.chan = chan;
+  }
+
+  destroy(): void {
+    this.chan.destroy();
+  }
+
+  setPlatform(platform : Platform): void {
```
smadbe, line 118: bindPlatform/bindListener (if platform has been renamed to listener)

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+    // Legacy calls
+    this.chan.bind('platform.updateHeight', (height : number) => platform.updateDisplay({ height: height }));
+  }
+
+  /**
+   * Task API functions
+   */
+  load(views : Object): Observable<void> {
+    return this.chan.call({
+      method: "task.load",
```
smadbe, line 139: Normally the linter should only allow single quotes, strange.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+    });
+  }
+
+  showViews(views : Object) : Observable<void> {
+    return this.chan.call({
+      method: "task.showViews",
+      params: views,
+      timeout: 2000
+    });
+  }
```
smadbe, line 221: I find quite confusing to have `showViews` here and in the "listeners`. I think they are two different things but I am not 100% sure. If they are different, probably they should have a different name.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+    if (key) {
+      if (key !== 'options' && key in res) {
+        return of(res[key]);
+      } else {
+        return of(defaultValue);
+      }
+    }
+    return of(res);
+  }
+}
```
smadbe, line 303: I don't see the reason to make them observable. From what I understand, there are signal from the task. So they can be treated with no delay. 
`getTaskParams` shows that, observables look useless there.

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -51,20 +50,20 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
...
     const sToken = '';
-    this.setUrl(this.taskProxyManager.getUrl(url, sToken, 'http://algorea.pem.dev', 'task-'));
+    this.setUrl(getTaskUrl(url, sToken, 'http://algorea.pem.dev', 'task-'));
   }
 
   ngAfterViewInit(): void {
     if (this.iframe) {
       const iframe = this.iframe.nativeElement;
-      this.taskProxyManager.getTaskProxy(iframe, this.taskIframeLoaded.bind(this), false);
+      getTaskProxy(iframe).subscribe((task: Task) => this.taskIframeLoaded(task));
```
smadbe, line 59: You still have this cascade of callback `ngAfterViewInit` -> `taskIframeLoaded -> `taskLoaded` -> `setViews`. From what I understand, that's out of scope of this PR.

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -86,26 +85,34 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  reloadAnswerState(answer : string, state : string): void {
+    if (!this.task) {
+      return;
+    }
+    const task = this.task;
+    task.reloadAnswer(answer)
+      .pipe(switchMap(() =>
+        task.reloadState(state)
+      ))
+      .subscribe(() => {});
```
smadbe, line 115:
```suggestion
      .subscribe();
```

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -117,7 +124,7 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   updateHeight(): void {
```
smadbe, line 126: For later probably:
Hard to get when seeing `updateHeight, if this function change the height of task, change immediately the height of the display or schedule the update of height of the display in accordance with the height of the task

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -117,7 +124,7 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
...
-    this.task?.getHeight(this.setHeight.bind(this));
+    this.task?.getHeight().subscribe(this.setHeight.bind(this));
```
smadbe, line 127: More readable and the way we do in the project:
```suggestion
    this.task?.getHeight().subscribe(val => this.setHeight(val));
```

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -140,26 +147,32 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
 }
 
 export class ItemDisplayPlatform extends Platform {
```
smadbe, line 149: I don't have the full picture yet but why not having Platform (or TaskListener) to be an interface and having `ItemDisplayComponent` implement this interface directly? Intuitively I think that's how I would have done it but maybe there are reasons to do it this way.

src/app/modules/item/task/rxjschannel.ts
```diff
@@ -0,0 +1,75 @@
+import { build, ChannelConfiguration, MessageTransaction, MessagingChannel } from "jschannel";
```
smadbe, line 1: I would prefer you put this files in a more explicit directory, for instance  `src/app/modules/item/task-communication` (or a better name)

smadbe (changes requested):
Many comments, maybe not all relevant :-)

smadbe:
Ah I haven't reviewed `rxjschannel.ts ` yet. Maybe that's the main file you wanted me to review :-D

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+
+function getRandomID(): string {
+  const low = Math.floor(Math.random() * 922337203).toString();
+  const high = Math.floor(Math.random() * 2000000000).toString();
+  return high + low;
+}
+
+/** Get URL for a task with specific parameters */
+export function getTaskUrl(taskUrl : string, token : string, platform : string, prefix? : string, locale? : string) : string {
+  const channelId = (prefix || '') + getRandomID();
```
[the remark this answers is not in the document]
mblockelet: Thanks, fixed.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+}
+
+function getRandomID(): string {
+  const low = Math.floor(Math.random() * 922337203).toString();
+  const high = Math.floor(Math.random() * 2000000000).toString();
+  return high + low;
+}
+
+/** Get URL for a task with specific parameters */
+export function getTaskUrl(taskUrl : string, token : string, platform : string, prefix? : string, locale? : string) : string {
```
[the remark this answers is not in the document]
mblockelet: I agree, renamed to `taskUrlWithParameters`.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+      + '&sPlatform=' + encodeURIComponent(platform)
+      + '&channelId=' + encodeURIComponent(channelId);
+  if (locale) {
+    url += '&sLocale=' + encodeURIComponent(locale);
+  }
+  return url;
+}
+
+/** Get Task object from an iframe */
+export function getTaskProxy(iframe : HTMLIFrameElement): Observable<Task> {
```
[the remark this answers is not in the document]
mblockelet: Renamed to `taskProxyFromIframe`.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+  // Check every second whether the iframe loaded
```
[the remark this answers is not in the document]
mblockelet: You're right, changed to 100ms.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+          }
+          return n;
+        }
+      ),
+
+      // Get contentWindow when it exists
+      filter(_ => !!iframe.contentWindow),
+      take(1),
+      map(_ => iframe.contentWindow as Window),
+
```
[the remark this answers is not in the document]
mblockelet: It is more elegant and readable, thank you.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+      // Get the RxMessagingChannel from the contentWindow
+      switchMap((contentWindow: Window) =>
+        rxBuild({
+          window: contentWindow,
+          origin: "*",
```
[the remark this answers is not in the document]
mblockelet: Changed to single but no the linter isn't complaining. It doesn't seem like there's any rule set in .eslintrc.js about quotes.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+      ),
+
+      // Get the Task from the channel
+      map((chan: RxMessagingChannel) => new Task(chan))
+    );
+}
+
+
+/**
+ * Task object, created from an iframe DOM element
```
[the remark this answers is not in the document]
mblockelet: Fixed the comment. I only had noticed it and fixed it in the next PR.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+ */
+export class Task {
+  platformSet = false;;
```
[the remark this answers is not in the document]
mblockelet: Thanks, fixed.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+  chan: RxMessagingChannel;
```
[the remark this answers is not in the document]
mblockelet: Changed to private.

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,303 @@
...
+      timeout: 2000
+    });
+  }
+}
+
+/*
+ * Platform object definition, created from a Task object (see below)
+ */
+
+export class Platform {
```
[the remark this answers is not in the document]
mblockelet: In the API it's called Platform, and in this file the name is used for instance in the method names bound in Task.setPlatform (`'platform.validate'`) which can't be changed because of the API, so changing the name here would probably be confusing.

src/app/modules/item/task/task-xd-pr.ts
mblockelet: It is used by the "Platform", for instance in [platform.validate in item-display](https://github.com/France-ioi/AlgoreaFrontend/blob/1478336586c442642df72db451b669d4978ea08d/src/app/modules/item/pages/item-display/item-display.component.ts#L164), as the Platform needs to get the current answer to be validated.

src/app/modules/item/task/task-xd-pr.ts
mblockelet: Renamed to `bindPlatform`.

src/app/modules/item/task/task-xd-pr.ts
mblockelet: Fixed.

src/app/modules/item/task/task-xd-pr.ts
mblockelet: As it's a function defined in the API, I can't really rename it. They're not really different, as the calls are basically meant to synchronize the frontend and the task as to what should be displayed (task statement, editor, hints, ...). Most of the time, the platform will tell the task "display the editor now", but sometimes the task can change view and hence needs to tell the platform "I'm actually displaying the hints now".

The tabs in the frontend, above the task, will allow to select the task view to be displayed.

src/app/modules/item/task/task-xd-pr.ts
mblockelet: They can't always be treated with no delay, for instance when it comes to validate or askHint which require calls to the backend and/or the task. Moreover, for instance for validate, making it an observable allows to keep the rxjs logic, [example here with this first version of validate](https://github.com/France-ioi/AlgoreaFrontend/blob/1478336586c442642df72db451b669d4978ea08d/src/app/modules/item/pages/item-display/item-display.component.ts#L164).

As some seem better to be written with Observables, I just made everything Observable to avoid having to write different definitions for `channel.bind`.

src/app/modules/item/pages/item-display/item-display.component.ts
mblockelet: Yes, that's out of scope of this PR. I didn't try to change it yet as anyway the PR focusing on item-display will change that code a lot and make something better.

src/app/modules/item/pages/item-display/item-display.component.ts
mblockelet: Fixed.

src/app/modules/item/pages/item-display/item-display.component.ts
mblockelet: I'd rename it to `updateIframeHeight` if that sounds clearer.

src/app/modules/item/pages/item-display/item-display.component.ts
mblockelet: Fixed.

src/app/modules/item/pages/item-display/item-display.component.ts
mblockelet: I wasn't sure whether it's clean to have a component also implement another interface which is unrelated to it being a component. I'll do that.

src/app/modules/item/task/rxjschannel.ts
mblockelet: Sure, I'll rename them. I'm not sure whether GitHub will transfer comments properly after a rename, so I'll do that once this PR is closer to completion.

src/app/modules/item/pages/item-display/item-display.component.ts
smadbe: I tried to guess what  `updateIframeHeight` would mean before checking at the code... and it was not really doing what I expected. The minimum would be to add explicit comments (even short). I think I would prefer a more explicit name, as `updateIframeToTaskHeight`.

src/app/modules/item/task/task-xd-pr.ts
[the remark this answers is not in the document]
smadbe: I don't know how come it disappears from the checks... I asked Aleksandr to re-add this rule in eslint.

src/app/modules/item/task/task-xd-pr.ts
smadbe: But without changing the "method" name in the API the function can have a more explicit name without problem, right? 
I wonder if it wouldn't make clearer to have the TS functions named "showViewsInTask" / "showViewsInPlatform"

src/app/modules/item/pages/item-display/item-display.component.ts
smadbe: I seems much cleaner if I haven't missed an aspect.  And so the "Platform" would just be an interface. 
In this context, I think the interface should not be called "Platform" but more "TaskListener"... IMO that's not a problem if the api method mention "platform" for that, which is the correct term in the API, but in the platform itself having a "platform" concept to represent oneself is quite confusing.

src/app/modules/item/task/rxjschannel.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  bind<T>(method: string, observable?: (params: T) => Observable<any>, doNotPublish?: boolean): MessagingChannel {
+    // Create a callback wrapping the observable bound
+    function callback(transaction: MessageTransaction, params: T): void {
+      if (!observable) {
+        return;
+      }
+      const cb$ = observable(params);
+      cb$.subscribe({
+        next: transaction.complete,
```
smadbe, line 45: `next` is not `complete`, it may be called several times.
```suggestion
        complete: transaction.complete,
```

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -140,26 +148,32 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
...
   constructor(task: Task, taskParams: TaskParams) {
     super(task);
     this.taskParams = taskParams;
   }
 
-  validate(mode : string, success : CompleteFunction, _error : ErrorFunction) : void {
+  validate(mode : string) : Observable<void> {
     if (mode == 'cancel') {
       // TODO reload answer
+      return of();
```
smadbe, line 160:
```suggestion
      return EMPTY;
```

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -140,26 +148,32 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
...
-        this.task.gradeAnswer(answer, '', (results : any) => {
-          success(results);
-        });
-      });
+      return this.task.getAnswer()
+        .pipe(
+          switchMap((answer: string) => this.task.gradeAnswer(answer, '')),
+          switchMap((_results : any) =>
+            // TODO Do something with the results
+            of()
```
smadbe, line 168:
```suggestion
            EMPTY
```

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -140,26 +148,32 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
...
+          )
+        );
     }
+    // Other unimplemented modes
+    return of();
```
smadbe, line 173:
```suggestion
    return EMPTY;
```

src/app/modules/item/task/rxjschannel.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  constructor(innerChan: MessagingChannel) {
+    this.innerChan = innerChan;
+  }
+
+  unbind(method: string, doNotPublish?: boolean): boolean {
+    return this.innerChan.unbind(method, doNotPublish);
+  }
+
+  bind<T>(method: string, observable?: (params: T) => Observable<any>, doNotPublish?: boolean): MessagingChannel {
```
smadbe, line 37:
```suggestion
  bind<T>(method: string, observable?: (params: T) => Observable<unknown>, doNotPublish?: boolean): MessagingChannel {
```

src/app/modules/item/task/rxjschannel.ts
```diff
@@ -0,0 +1,75 @@
...
+      cb$.subscribe({
+        next: transaction.complete,
+        error: error => transaction.error(error, '')
+      });
+      transaction.delayReturn(true);
+    }
+    return this.innerChan.bind(method, callback, doNotPublish);
+  }
+
+  call<S, T>(message: RxMessage<S>, validator?: (result?: any) => T): Observable<T> {
```
smadbe, line 53: I have several remarks but I supposed you have changed everything in your PR about typing.

src/app/modules/item/task/rxjschannel.ts
```diff
@@ -0,0 +1,75 @@
+import { build, ChannelConfiguration, MessageTransaction, MessagingChannel } from 'jschannel';
+import { Observable } from 'rxjs';
+
+export interface RxMessage<T> {
```
smadbe, line 4: For the current usage (there is no use of "notify" for now), the generic type is useless here as you do not care about the type of your arg, so you could set it`unknown` and remove the `<T>`.

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -140,26 +148,32 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
...
 export class ItemDisplayPlatform extends Platform {
-  taskParams: any;
+  taskParams: TaskParams;
   constructor(task: Task, taskParams: TaskParams) {
     super(task);
     this.taskParams = taskParams;
   }
 
-  validate(mode : string, success : CompleteFunction, _error : ErrorFunction) : void {
+  validate(mode : string) : Observable<void> {
```
smadbe, line 157: if it is the only valid values:
```suggestion
  validate(mode : 'validate'|'cancel') : Observable<void> {
```

src/app/modules/item/pages/item-display/item-display.component.ts
```diff
@@ -140,26 +148,32 @@ export class ItemDisplayComponent implements OnInit, AfterViewInit, OnDestroy {
...
     }
     if (mode == 'validate') {
-      this.task.getAnswer((answer: string) => {
-        this.task.gradeAnswer(answer, '', (results : any) => {
-          success(results);
-        });
-      });
+      return this.task.getAnswer()
+        .pipe(
+          switchMap((answer: string) => this.task.gradeAnswer(answer, '')),
```
smadbe, line 165: as the observable is typed, it is useless to repeat the type
```suggestion
          switchMap(answer => this.task.gradeAnswer(answer, '')),
```

smadbe (changes requested):
Review of `rxjschannel.ts` and more random remarks.

smadbe:
As we usually do in this project (probably written in https://france-ioi.github.io/contributing/github#the-review as well), do not "resolve" yourself so that I can check and approve each of your answer.

src/app/modules/item/task/task-xd-pr.ts
smadbe: Okay for the delay. 
For `getTaskParams` which returns a response, I think `bind` does not do anything with the observable result, does it?

src/app/modules/item/task/task-xd-pr.ts
```diff
@@ -0,0 +1,293 @@
...
+export type TaskGrade = any;
+export type TaskResources = any;
+export type TaskDisplayData = any;
+export type TaskLog = any;
+
+function getUrlParameterByName(name : string, url : string) : string {
+  const regex = new RegExp('[\\?&]' + name + '=([^&#]*)'),
+    results = regex.exec(url);
+  return results && results[1] ? decodeURIComponent(results[1].replace(/\+/g, ' ')) : '';
+}
```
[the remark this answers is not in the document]
mblockelet: Honestly I only kept the function from the original file without thinking, I should have checked. There is indeed a function in helpers which can be used instead of this, so I did that.

src/app/modules/item/task/rxjschannel.ts
mblockelet: `transaction.complete` from jschannel needs to receive the return value from the Observable, and is supposed to be called only once. Should I add a take(1) to ensure next is called only once?

src/app/modules/item/task/rxjschannel.ts
mblockelet: Yes, I changed all the typing/validation side in the PR.

src/app/modules/item/pages/item-display/item-display.component.ts
mblockelet: No there can be other values too, there are a lot of possible ones and the ones not recognized by AlgoreaFrontend should just be ignored.

src/app/modules/item/task/rxjschannel.ts
mblockelet: Removed the generic type.

mblockelet:
Following your suggestion, I renamed the folder to `task-communication`, and also `task-xd-pr` to a clearer name `task-proxy`. Also batch-commited a few of your suggestions through GitHub.

src/app/modules/item/task/rxjschannel.ts
smadbe: > Should I add a take(1) to ensure next is called only once?

Yes

smadbe (commented):
I think these comments still need to be addressed or answered:
https://github.com/France-ioi/AlgoreaFrontend/pull/525#discussion_r669383887
https://github.com/France-ioi/AlgoreaFrontend/pull/525#discussion_r669461449
https://github.com/France-ioi/AlgoreaFrontend/pull/525#discussion_r669380975
https://github.com/France-ioi/AlgoreaFrontend/pull/525#discussion_r674809368

src/app/modules/item/task/task-xd-pr.ts
mblockelet: I renamed the functions to `showViewsInTask` and `viewsShownByTask` ; for the second name, I think it indicates better what happens (some views were just shown by the task, which then communicates it to the platform).

In the next PR (item-display modifications), many of these functions will actually be used in a proper way, and I'll take the chance to add comments describing their use and behavior. Moreover the "platform" functions will be moved to item-display.

src/app/modules/item/task/task-xd-pr.ts
mblockelet: `bind` [sends the result to the task through jschannel](https://github.com/France-ioi/AlgoreaFrontend/pull/525/files#diff-b0866128bd3efd33f554be8000bf7103073584b8206ad42a2a1a1297960b77caR48).

src/app/modules/item/task/rxjschannel.ts
mblockelet: Done.

src/app/modules/item/pages/item-display/item-display.component.ts
mblockelet: Renamed to TaskListener. Note that as discussed, the TaskListener functions in `task-proxy.ts` will be merged into the item-display component.

src/app/modules/item/pages/item-display/item-display.component.ts
mblockelet: This will be addressed in the next PR (which will probably remove the function altogether).

mblockelet:
Sorry, I should have addressed all comments now. Some modifications are still to be addressed in the future item-display PR.

src/app/modules/item/task/task-xd-pr.ts
smadbe: Ah yes `viewsShownByTask` is better, actually I was wrong about I think it was doing.

src/app/modules/item/task/task-xd-pr.ts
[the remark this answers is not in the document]
mblockelet: Renamed to TaskListener. We could see about renaming other functions too if their name is not clear.

PR #526, opened 2021-07-12
Typing for task-xd-pr

mblockelet:
This PR adds typing for task-xd-pr, and uses io-ts to ensure types of data received from the task are the expected ones.

Two functions being currently unused, typing was not done yet for those, as typing those is complex, and they will probably end up never being used anyway.


src/app/modules/item/task-communication/rxjschannel.ts
```diff
@@ -35,31 +40,70 @@ export class RxMessagingChannel {
...
-    function callback(transaction: MessageTransaction, params: T): void {
+    function callback(transaction: MessageTransaction, ...params: any[]): void {
       if (!observable) {
         return;
       }
-      const cb$ = observable(params);
+      // Select params
+      const actualSelector = selector
+        ? selector
+        : (result: any[]) : unknown => (result.length > 0 ? result[0] : undefined);
```
ThomasJuster, line 54: Note to self:
To me, selector logic should not be internalise to the bind method since
- it is not the method purpose to map anything
- it can be done easily using rxjs
- it is (currently) used only once

src/app/modules/item/task-communication/rxjschannel.ts
```diff
@@ -35,31 +40,70 @@ export class RxMessagingChannel {
...
+      const decodedParams = validator
+        ? fppipe(
+          validator.decode(selectedParams),
+          fold(
+            error => {
+              throw new Error(D.draw(error));
+            },
+            decoded => decoded
+          ))
+        : selectedParams as T;
```
ThomasJuster, line 68: Note to self:
Same as the selector logic, to me it should be done via rxjs, the method should not internalise decoders

src/app/modules/item/task-communication/rxjschannel.ts
```diff
@@ -35,31 +40,70 @@ export class RxMessagingChannel {
...
 
-  call<T>(message: RxMessage, validator?: (result?: any) => T): Observable<T> {
+  /** Call a remote method through jschannel, return the result through an Observable */
+  call<T>(message: RxMessage, validator?: D.Decoder<unknown, T>): Observable<T> {
     // Create an Observable wrapping the inner jschannel call
     return new Observable<T>(subscriber => {
+      const selector = message.selector
+        ? message.selector
+        : (result: any[]) : unknown => (result.length > 0 ? result[0] : undefined);
+
```
ThomasJuster, line 89: Note to self:
Same as the `bind()` method, to me it should be done via rxjs, the method should not internalise decoders

src/app/modules/item/task-communication/rxjschannel.ts
```diff
@@ -35,31 +40,70 @@ export class RxMessagingChannel {
...
+          const decodedResult = validator
+            ? fppipe(
+              validator.decode(selectedResult),
+              fold(
+                error => {
+                  throw new Error(D.draw(error));
+                },
+                decoded => decoded
+              ))
+            : selectedResult as T;
```
ThomasJuster, line 104: Note to self:
Same as the `bind()` method, to me it should be done via rxjs, the method should not internalise decoders

src/app/modules/item/task-communication/types.ts
```diff
@@ -0,0 +1,70 @@
...
+});
+export type UpdateDisplayParams = D.TypeOf<typeof updateDisplayParamsDecoder>;
+
+// Log data sent by the task
+export const taskLogDecoder = D.array(D.string);
+export type TaskLog = D.TypeOf<typeof taskLogDecoder>;
+
+// Currently unused
+export type TaskMetaData = any;
+export type TaskResources = any;
```
ThomasJuster, line 70: Note to self:
Replace `any` by `never`

ThomasJuster (commented):
Notes to self

PR #527, opened 2021-07-12
keep permission propagation in item children edit

### LINKED ISSUE #516: Keep permission propagation in item children edit
## Motivations
(first make sure you have understood [permission propagation rules](https://france-ioi.github.io/algorea-devdoc/design/access-rights/items/#propagation-across-parent-child-items))
Currently, propagation may be lost when item children are changed as they are fetched but not saved by the frontend when pushing the children.

## Acceptance criteria
* [x] Make sure the `itemUpdate` service is given the propagation value received for the children which were existing before the edition, even if items are reordered.
* [x] For new children added, do not set propagations, the default will be set appropriately by the backend

ThomasJuster:
Fixes #516

### DIFF
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
@@ -20,6 +20,11 @@ export interface ChildData {
     validated: boolean,
     score: number,
   },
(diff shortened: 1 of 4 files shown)


smadbe (approved):
It looks like it works 👍

PR #531, opened 2021-07-13
fix auth by cookie on cross site for subdomains only

### LINKED ISSUE #507: Fix cookie problem on cross site

ThomasJuster:
Fixes #507 for requests across subdomains only, not cross-domain requests


src/app/shared/helpers/domain.ts
```diff
@@ -0,0 +1,6 @@
+export function getDomain(href: string): string {
```
smadbe, line 1: should it be called "getRootDomain"? From my point of view, the domain of `https://sub.example.com/api/stuff ` is `sub.example.com`, not `example.com`

src/app/shared/http-services/auth.http-service.ts
```diff
@@ -20,10 +21,31 @@ const longAuthServicesTimeout = 10000;
...
+     * A cookie party is defined by its domain:
+     * - first-party cookies have domain "dev.algorea.org", "api.algorea.org" or "algorea.org"
+     * - third-party cookies have domain or "complete-other-site.com"
+     * Safari and Firefox both disable third-party cookies by default, and Chrome will in 2021 (but postponed for 2023)
+     * Since third-party cookies will be removed in a near future, we chose not to enable this type of authentication,
+     * making the 'same-site' policy always true.
+     */
+  } : undefined;
+  private requestConfig = appConfig.authType === 'cookies' ? {
+    withCredentials: !this.isSameDomain,
```
smadbe, line 42: `withCredentials` looks safe to set in all cookie cases, same domain or not, as it just marks the request as having credentials. (and it might save us issue with future changes in browsers)

src/app/shared/http-services/auth.http-service.ts
```diff
@@ -20,10 +21,31 @@ const longAuthServicesTimeout = 10000;
...
+    /**
+     * NOTE:
+     * withCredentials is necessary to allow cross-origin resources to set cookies, even if the resources share the same domain.
+     * For instance, dev.algorea.org calling api.algorea.org is a cross-origin request,
+     * but a cookie with domain "algorea.org" is a first-party cookie, see note on cookie_same_site.
+     */
```
smadbe, line 48: From what I understand, you claim here that dev.algorea.org is cross-origin api.algorea.org so should use `withCredentials` but that the way `isSameDomain` is computed makes in true in this case.

src/app/shared/http-services/auth.http-service.ts
```diff
@@ -20,10 +21,31 @@ const longAuthServicesTimeout = 10000;
...
     use_cookie: '1',
-    cookie_secure: appConfig.secure ? '1' : '0',
-    cookie_same_site: appConfig.sameSite ? '1' : '0',
+    cookie_secure: this.apiUrl.protocol === 'https:' || this.apiUrl.hostname === 'localhost' ? '1' : '0',
+    cookie_same_site: '1',
+    /**
+     * NOTE on cookie_same_site:
+     * The 'same-site' policy only applies for auth by third-party cookie.
+     * A cookie party is defined by its domain:
+     * - first-party cookies have domain "dev.algorea.org", "api.algorea.org" or "algorea.org"
```
smadbe, line 34: this sentence is not clear. As written, it seems that in absolute first-party cookies are those with domain "dev.algorea.org", "api.algorea.org" or "algorea.org".

src/app/shared/interceptors/credentials.interceptor.ts
```diff
@@ -0,0 +1,26 @@
...
+  private withCredentials = appConfig.authType === 'cookies' && !this.isSameOrigin;
+
+  constructor() {}
+
+  intercept(req: HttpRequest<unknown>, next: HttpHandler): Observable<HttpEvent<unknown>> {
+    if (!req.url.toLowerCase().startsWith(appConfig.apiUrl)) {
+      return next.handle(req);
+    }
+
+    return next.handle(req.clone({ withCredentials: this.withCredentials }));
```
smadbe, line 24: same remark about `withCredentials` as above.

smadbe (changes requested):

src/app/shared/helpers/domain.ts
ThomasJuster: Fixed. It became dead code.

src/app/shared/http-services/auth.http-service.ts
ThomasJuster: Fixed.

src/app/shared/http-services/auth.http-service.ts
ThomasJuster: Fixed.

src/app/shared/http-services/auth.http-service.ts
ThomasJuster: I tried making the whole comment clearer

src/app/shared/http-services/auth.http-service.ts
```diff
@@ -28,16 +28,35 @@ export class AuthHttpService {
...
-     * - third-party cookies have domain or "complete-other-site.com"
+     * NOTE on cookie_same_site and the cookie attribute "SameSite":
      * Safari and Firefox both disable third-party cookies by default, and Chrome will in 2021 (but postponed for 2023)
      * Since third-party cookies will be removed in a near future, we chose not to enable this type of authentication,
      * making the 'same-site' policy always true.
+     *
+     * NOTE on cookie party:
+     * A cookie party is determined by its "Domain" attribute, which defaults to the hostname.
+     * As defined by the MDN glossary, the domain represents an "authority", an organization to which belongs the cookie.
+     * In our case, the authority is "algorea.org".
```
smadbe, line 37: It is more "for instance" than "In our case", this domain may change for this app.

smadbe (approved):

PR #537, opened 2021-07-16
Fixes breadcrumbs

### LINKED ISSUE #438: Bug: fix breadcrumb nav
For instance, nav from http://dev.algorea.org/en/#/activities/by-id/83802952425549250;parentAttempId=0;path=4702,4102,899800937596940136,883636806228183397/edit to "Parcours officiel" through the breadcrumb sends to skills (?) unexpected !

Iloveall:
Fixes #438

### DIFF
--- a/src/app/modules/item/http-services/get-breadcrumb.service.ts
+++ b/src/app/modules/item/http-services/get-breadcrumb.service.ts
@@ -49,6 +49,7 @@ export class GetBreadcrumbService {
               path: items.slice(0,idx).map(it => it.item_id),
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
               attemptId: item.attempt_id!, // the service ensures the attempt for all but last is given
+              contentType: itemRoute.contentType,
             } as ItemRoute,
             attemptCnt: item.attempt_number ? +item.attempt_number : undefined,
           })).concat([{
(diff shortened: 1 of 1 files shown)


src/app/modules/item/http-services/get-breadcrumb.service.ts
```diff
@@ -49,6 +49,7 @@ export class GetBreadcrumbService {
               path: items.slice(0,idx).map(it => it.item_id),
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
               attemptId: item.attempt_id!, // the service ensures the attempt for all but last is given
+              contentType: itemRoute.contentType,
```
smadbe, line 52: It means you'll navigate to the same "content type" as the current page, right? It might not be the case as you may visit a chapter as a child of skill (maybe not for now, but it still makes sense). What's important here is to use as content type the one from the item of concern, so just infer it from the type of the item, I suppose you have it.

smadbe (changes requested):

src/app/modules/item/http-services/get-breadcrumb.service.ts
Iloveall: GetBreadcrumb response doesn't have content type. The structure is:

attempt_id,
item_id,
language_tag,
title,

smadbe: Ok, so let's add the type... should be done in a few days.

smadbe: https://github.com/France-ioi/AlgoreaBackend/issues/773

smadbe: Done. So you now have the type in the breadcrumb service: https://france-ioi.github.io/algorea-devdoc/api/#operation/itemBreadcrumbsGet

Iloveall: Fixed

smadbe (approved):

PR #539, opened 2021-07-16
Feature/indicate when item group search is partial

### LINKED ISSUE #532: Indicate when item/group search is partial
## Overview

When searching for an item or group, when the maximum number of elements is displayed, indicate that it is a partial result.

## Acceptance criteria
- [ ] Search for 5 elements instead of 4, to be able to know if there are >4 results.
- [ ] If there are <=4 elements returned, leave the result as it is.
- [ ] if the are >4 elements returned, display only 4 but add a sentence below the results (centered, as bit lighter than the list?) "More than 4 results match your search terms. If you have not found what you are looking for, make your search more specific."
- [ ] Make sure it works for items and groups.

Iloveall:
Fixes #532

### DIFF
--- a/src/app/modules/group/http-services/search-group.service.ts
+++ b/src/app/modules/group/http-services/search-group.service.ts
@@ -31,7 +31,7 @@ export class SearchGroupService {
 
   search(
     searchString: string,
-    limit = 4,
+    limit = 5,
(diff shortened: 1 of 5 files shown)


smadbe (approved):
👍 great

PR #548, opened 2021-07-20
Feature/create group directly

### LINKED ISSUE #540: Create group directly (not only via adding subgroups)
## Overview

Let users create new groups, not only by creating subgroup via group composition.

## Acceptance criteria
- [x] In the "My groups" page, under the "The groups you manage" table, add a widget similar to: 
Screenshot 2021-07-19 at 09 58 34
- [x] with, as title, "Create a new group"
- [x] with only the left part, to create a new group
- [x] the placeholder should be something like "Enter a title to create a new group" (the placeholder should probably be a parameter of the component)
- [x] When created, go directly on the group page.
- [ ] Test that it works correctly.

Iloveall:
Fixes #540


src/app/modules/group/components/add-group/add-group.component.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  constructor(
+    private groupCreationService: GroupCreationService,
+    private router: Router,
+  ) {}
+
+  addChild(group: AddedContent<GroupType>): void {
+    this.groupCreationService.create(group.title, group.type).subscribe(createdId => {
+      void this.router.navigate([ 'groups', 'by-id', createdId, 'details' ]);
+    });
```
smadbe, line 49: For all editions, you need to handle errors! 
If there is an error, you have to display a message on the top right as usual.

And what happens if the request is slow and the user click several times on the icon? The widget should be frozen during the fetching.

smadbe (changes requested):
Each time there is an action (not a get) you need to handle the error and waiting time!

Can you fix the height of these boxes? They are half filled with void, that looks strange.
Screenshot 2021-07-20 at 16 50 10

src/app/modules/group/components/add-group/add-group.component.ts
Iloveall: Fixed

src/app/modules/group/components/add-group/add-group.component.ts
```diff
@@ -38,14 +39,25 @@ export class AddGroupComponent {
...
   ) {}
 
   addChild(group: AddedContent<GroupType>): void {
-    this.groupCreationService.create(group.title, group.type).subscribe(createdId => {
-      void this.router.navigate([ 'groups', 'by-id', createdId, 'details' ]);
+    this.state = 'addingGroup';
+    this.groupCreationService.create(group.title, group.type).subscribe({
+      next: createdId => {
+        this.state = 'ready';
+        void this.router.navigate([ 'groups', 'by-id', createdId, 'details' ]);
```
smadbe, line 55: Sorry I forgot: Could you add success action feedback on success as well ? ("Group successfully created")

Iloveall: Fixed

smadbe (approved):

PR #559, opened 2021-07-23
Fixes save edit item for empty child items

### LINKED ISSUE #558: Bug: item-edit save not doing anything
## Overview
On http://dev.algorea.org/en/#/activities/by-id/7766562542777237704;attempId=0;path=4702,1352246428241737349,561969155071897034/edit (with the usual user).
If I remove the existing child and "save", the UI greys out as expected but forever and no network call has been emitted. 

## Acceptance criteria
- [ ] Check what's happening. 
- [ ] If it is a simple bug, fix it. If not, ask Damien.

Iloveall:
Fixes #558


smadbe (approved):
👍 
😮 This is not new so this bug was there for a while.
To remember: when forkJoin may be given an empty array, it does not complete... so we have to address this specific case separately!

PR #561, opened 2021-07-26
Feature/improve header hiding

### LINKED ISSUE #545: Improve header hiding (make it work on non-item pages)
## Overview

When pressing the arrow icon on the top of the left menu, the top of the page is forced scrolled of a fixed number of pixel (the `folded`/`unfolded` css class) . While it works well for some pages, for instance most items, it is buggy on others, for instance groups, or anywhere the header size may change depending on its content.

## Acceptance criteria
- [x] Instead of scrolling as it is done for now, hide/show (with animation if possible) specific html elements marked as "to be hidden when header is hiden", so using specific css class, ng directive, or condition depending on the new `LayoutService.fullFrameContent$`.
- [ ] Check if it may be useful to use https://www.primefaces.org/primeng/showcase/#/styleclass to avoid having to handle the show/hide in every page component. ???
- [ ] Make sure it works correctly on every page: 
   - [ ] items
   - [ ] groups with various manager/membership info
   - [ ] personal pages
   - [ ] my groups page
   - [ ] in case of errors when loading item/group

Iloveall:
Fixes #545

### DIFF
--- a/src/app/core/app.component.scss
+++ b/src/app/core/app.component.scss
@@ -234,11 +234,7 @@ a:hover {
       -o-transition: margin-left 1s;
 
       .main-content {
-        margin-top: -12.0833rem;
-        transition: margin-top 1s;
-        -webkit-transition: margin-top 1s;
-        -moz-transition: margin-top 1s;
-        -o-transition: margin-top 1s;
+        margin-top: 1rem;
       }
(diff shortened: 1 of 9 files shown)


smadbe (approved):
👍

PR #564, opened 2021-07-26
Adds permissions for searched items

### LINKED ISSUE #560: Allow propagation edition: initial value for searched item
## Overview
Permissions has been added to the item search services. You can now infer the propagation options for items added from search.

## Acceptance criteria
- [ ] Display the propagation widget for items added from search.
- [ ] Apply the same rules (as in #529) to determine what options should be available.

Iloveall:
Fixes #560

### DIFF
--- a/src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
+++ b/src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
@@ -145,8 +145,8 @@ export class ItemChildrenEditComponent implements OnChanges {
       ...child,
       scoreWeight: DEFAULT_SCORE_WEIGHT,
       isVisible: true,
-      contentViewPropagation: child.permissions?.canGrantView === 'none' ? 'none' : 'as_info',
-      permissions: child.permissions || permissionsForCreatedItem,
+      contentViewPropagation: (child.permissions ?? permissionsForCreatedItem).canGrantView === 'none' ? 'none' : 'as_info',
+      permissions: child.permissions ?? permissionsForCreatedItem,
     });
     this.onChildrenListUpdate();
     this.childrenChanges.emit(this.data);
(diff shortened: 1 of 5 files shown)


smadbe (changes requested):
I think there is an error: 
The current user does have any "grant view" permission of "ALGOREA ADVENTURE". However, if I go to "Tas de sable" (http://dev.algorea.org/branch/feature/allow-propagation-edit-for-search-added-items/en/#/activities/by-id/1012286565380301759;attempId=0;path=4702,4102/edit) and add "ALGOREA ADVENTURE" I have the options "locked and hidden" and "locked" while I suppose "locked" should not be listed... if I save the backend refuses anyway...

You should also rebase (or merge master) before my code review, because some other changes have been done to the same component, so it is better to integrate the change as early as possible.

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
```diff
@@ -139,14 +139,12 @@ export class ItemChildrenEditComponent implements OnChanges {
...
       isVisible: true,
-      ...(!child.id ? {
-        contentViewPropagation: 'as_info',
-        permissions: {
-          canView: 'solution',
-          canEdit: 'all_with_grant',
-          canGrantView: 'solution_with_grant',
-        },
-      } : {})
+      contentViewPropagation: child?.permissions?.canGrantView === 'none' ? 'none' : 'as_info',
```
smadbe, line 142: I assume we rely here that if it is a new child, it will be undefined, so `!== none`.
I am feel that it is a bit error prone, the next developer changing that may introduce an error.
I prefer you create a constant `permissionsForCreatedItem` (with the value you have on line 143-177), then use this constant when you do not have permissions.

BTW, `child?.pe...` -> child is not optional so I don't see the point of the optional chaining.

smadbe (changes requested):
small remark

src/app/modules/item/components/item-children-edit/item-children-edit.component.ts
Iloveall: Fixed

smadbe (approved):

PR #566, opened 2021-07-27
Adds observation mode for users

### LINKED ISSUE #553: Allow observation on individual users
## Overview
The same way groups are, the individual user may be watched by some users who has to specific permission.

## Acceptance criteria
- [ ] If the current user is not the visited user AND the current user can watch the visited user, display the "observation" button at the right of the header (as in groups). 
- [ ] Make sure the behavior is the same as in the group, the button should be yellow if the user is already being observed

Iloveall:
Fixes #553

### DIFF
--- a/src/app/modules/group/components/user-header/user-header.component.html
+++ b/src/app/modules/group/components/user-header/user-header.component.html
@@ -1,0 +1,28 @@
+<div class="alg-page-header" *ngIf="user">
+  <div>
+    <div class="alg-page-header-title">
+      {{ user | userCaption }}
+    </div>
+
+    <div class="alg-page-header-subtitle" *ngIf="user.webSite">
+      <i class="fa fa-external-link-alt"></i>
+      <a class="web-site alg-link base-color" [attr.href]="user.webSite" target="_blank">{{ user.webSite }}</a>
+    </div>
+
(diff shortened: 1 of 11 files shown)


src/app/modules/group/components/user-header/user-header.component.html
```diff
@@ -0,0 +1,28 @@
...
+  <alg-page-navigator
+      *ngIf="true"
+      [allowEditing]="false"
+      [allowWatching]="user.currentUserCanWatchUser && !user.isCurrentUser"
+      [isWatched]="!!(isCurrentGroupWatched$ | async)"
+      [allowFullScreen]="false"
+      navigationMode="nextAndPrev"
+      (edit)="onEditButtonClicked()"
+      (watch)="onStartWatchButtonClicked()"
+      (stopWatch)="onStopWatchButtonClicked()"
```
smadbe, line 26: indent of 4 instead of 2?

smadbe (approved):
Looks good 👍

src/app/modules/group/components/user-header/user-header.component.html
Iloveall: Fixed

PR #571, opened 2021-08-01
Adds switch control for entering min time

### LINKED ISSUE #546: Improve handling of "First allowed entering time"
## Overview
Improve handling of "First allowed entering time". Currently when "1000-01-01", the value is always set to "now" to have a calendar not starting from year 1000.  As a consequence, when changing another attribute and not touching this one, this value is set to the current time while it does not have any meaning.

Instead of this, add a switch to enable/disable setting a "First allowed entering time". 

## Acceptance criteria
- [ ] Add a switch on the left of the calendar (maybe useful to create a component shared with "Latest allowed entering time"?). 
- [ ] When disabled, do not display the calendar. Interpret this value as "1000-01-01".
- [ ] When enabled, set it to "now" by default.
- [ ] On loading, if the value is "1000-01-01 ...", disable the switch (and hide the calendar), otherwise enable it and show the calendar at the current value.
- [ ] Avoid the change of height when changing switch value as it happens with the "Latest allowed entering time".
- [ ] Test carefully each case.

Iloveall:
Fixes #546

### DIFF
--- a/src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
+++ b/src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
@@ -1,17 +1,15 @@
-import { Component, Input, OnChanges, OnInit, SimpleChanges } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
 import { FormGroup, Validators } from '@angular/forms';
 import { DropdownOption } from 'src/app/modules/shared-components/components/dropdown/dropdown.component';
 import { Item } from '../../http-services/get-item-by-id.service';
 import { HOURS } from 'src/app/shared/helpers/duration';
-
-const DEFAULT_ENTERING_TIME_MIN = '1000-01-01T00:00:00Z';
 
 @Component({
   selector: 'alg-item-edit-advanced-parameters',
   templateUrl: './item-edit-advanced-parameters.component.html',
   styleUrls: [ './item-edit-advanced-parameters.component.scss' ]
 })
-export class ItemEditAdvancedParametersComponent implements OnInit, OnChanges {
+export class ItemEditAdvancedParametersComponent implements OnInit {
   @Input() item?: Item;
   @Input() parentForm?: FormGroup;
 
@@ -65,13 +63,6 @@ export class ItemEditAdvancedParametersComponent implements OnInit, OnChanges {
     this.handleDurationValidation();
   }
 
-  ngOnChanges(changes: SimpleChanges): void {
-    if (changes.item && this.item?.enteringTimeMin.getTime() === new Date(DEFAULT_ENTERING_TIME_MIN).getTime()) {
-      this.parentForm?.get('entering_time_min')?.patchValue(new Date());
-      this.parentForm?.get('entering_time_min')?.markAsPristine();
-    }
-  }
-
   onRequiresExplicitEntryChange(): void {
     this.handleDurationValidation();
   }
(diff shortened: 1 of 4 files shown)


src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
```diff
@@ -90,23 +81,41 @@ export class ItemEditAdvancedParametersComponent implements OnInit, OnChanges {
...
   }
 
+  onEnteringTimeMinEnabledChange(event: boolean): void {
+    if (!event) {
+      return;
+    }
+
+    this.parentForm?.get('entering_time_min')?.patchValue(
+      new Date()
+    );
```
smadbe, line 91: So, just switching it off and on directly clear the previous value. This is not what we want. Should set to now only if it equaled DEFAULT_ENTERING_TIME_MIN.

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
```diff
@@ -90,23 +81,41 @@ export class ItemEditAdvancedParametersComponent implements OnInit, OnChanges {
...
+
+  computeEnteringTimeMaxValue(): void {
+    const enteringTimeMinEnabled = this.parentForm?.get('entering_time_min_enabled')?.value as boolean;
     const enteringTimeMin = this.parentForm?.get('entering_time_min')?.value as string;
-    const enteringTimeMinDate = new Date(enteringTimeMin);
+    const enteringTimeMinDate = enteringTimeMinEnabled ? new Date(enteringTimeMin) : new Date();
     this.parentForm?.get('entering_time_max')?.patchValue(
       new Date(enteringTimeMinDate.getTime() + HOURS)
     );
   }
```
smadbe, line 111: Same here. I would say that when enabling max time, if it has its default value (`9999-12-31` I think), you leave this value, otherwise you set it `max(now, min time) + 1h`

smadbe (changes requested):

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
Iloveall: Fixed

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
Iloveall: Fixed

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
```diff
@@ -62,25 +61,18 @@ export class ItemEditAdvancedParametersComponent implements OnInit, OnChanges {
...
+    this.handleDurationValidators();
   }
 
   onDurationEnabledChange(): void {
-    this.handleDurationValidation();
+    this.handleDurationValidators();
   }
 
-  handleDurationValidation(): void {
+  handleDurationValidators(): void {
```
smadbe, line 75:
```suggestion
  private handleDurationValidators(): void {
```

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
```diff
@@ -90,23 +82,51 @@ export class ItemEditAdvancedParametersComponent implements OnInit, OnChanges {
...
+    }
+
+    const enteringTimeMin = this.parentForm?.get('entering_time_min')?.value as Date;
+
+    if (enteringTimeMin.getTime() === new Date(DEFAULT_ENTERING_TIME_MIN).getTime()) {
+      this.parentForm?.get('entering_time_min')?.patchValue(
+        new Date()
+      );
+
+      this.computeEnteringTimeMaxValue();
```
smadbe, line 97: I am not sure it is useful in your implementation:
- if max is not enabled, it will not enter into the condition on line 113, so will not do anything
- if max was already enabled, it has already been set to "now" when enabled, so it will not enter into the condition on line 113, so will not do anything

smadbe (changes requested):
👍 only small details

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
Iloveall: Fixed

smadbe (approved):
👍

PR #577, opened 2021-08-04
Adds handling permissions for progress tab

### LINKED ISSUE #568: When the item cannot be watched, do not display the progress tab in observation mode
## Overview
In observation mode, the item "progress" tab should contain only the result from the observed group.
In case the item is not observable (we cannot see the result of any group on this item), there is no reason to display this tab.

## Acceptance criteria
If the current user cannot watch the item (can_watch < 'result'):
- [ ] don't display the "progress" tab when not on the progress tab
- [ ] if on the progress tab (via the url), display it but display under "Situation of ..." an error: "You are not allowed to watch the progress of other users on this content."

Iloveall:
Fixes #568

### DIFF
--- a/src/app/modules/item/pages/item-details/item-details.component.html
+++ b/src/app/modules/item/pages/item-details/item-details.component.html
@@ -1,4 +1,4 @@
-<ng-container *ngIf="itemLoadingstate$ | async as state">
+<ng-container *ngIf="itemLoadingState$ | async as state">
 
   <ng-container *ngIf="state.isReady && state.data as itemData">
 
@@ -22,10 +22,11 @@
         Content
       </a>
       <a
+        *ngIf="!(watchedGroup$ | async) || !!(watchedGroup$ | async) && itemData.item.permissions.canView !== 'none'"
         class="nav-tab-item"
         [routerLink]="'./progress'"
         routerLinkActive #progressTab="routerLinkActive"
-        [class.active]="progressTab.isActive">
+        [class.active]="progressTab?.isActive">
         <i class="fa fa-chart-line"></i>
         &nbsp;
         <span i18n>Progress</span>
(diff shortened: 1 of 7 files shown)


smadbe (changes requested):
First test: 
With the usual user:
- I start observing a group
- I go in activities, on "parcours officiel" at the root.
- the progress tab is visible while the user has `can_watch='none'` on the item.
- 
That does not fit the specs.

src/app/modules/item/pages/item-details/item-details.component.html
```diff
@@ -22,18 +22,22 @@
         Content
       </a>
       <a
+        *ngIf="!(watchedGroup$ | async) || !!(watchedGroup$ | async) && itemData.item.permissions.canWatch !== 'none' || !!(watchedGroup$ | async) && itemData.item.permissions.canWatch === 'none' && !contentTab.isActive"
```
smadbe, line 25: The condition looks good but can be simplified to 
```
!(watchedGroup$ | async) || itemData.item.permissions.canWatch !== 'none' || !contentTab.isActive"
```
easier 🙂

src/app/modules/item/pages/item-progress/item-progress.component.html
```diff
@@ -4,56 +4,65 @@
...
-
-      <div class="chapter-group-progress">
-        <alg-chapter-group-progress
-            *ngIf="chapterGroupProgressTab?.isActive"
+
+    <p class="not-allow-caption" *ngIf="!!session?.watchedGroup && itemData.item.permissions.canWatch === 'none'" i18n>
+      You are not allowed to watch the progress of other users on this content.
+    </p>
+
+    <ng-container *ngIf="!session?.watchedGroup || !!session?.watchedGroup && itemData.item.permissions.canWatch !== 'none'">
```
smadbe, line 12: Use a else, it is DRYer and safer than one of the two blocks is shown.

smadbe (changes requested):
just details to be fixed

src/app/modules/item/pages/item-details/item-details.component.html
Iloveall: Thank you, fixed

src/app/modules/item/pages/item-progress/item-progress.component.html
Iloveall: Fixed

smadbe (approved):

PR #581, opened 2021-08-05
Moves scroll logic outside of angular

### LINKED ISSUE #578: Move scroll listener outside angular
## Overview
The `@HostListener('window:scroll', [ '$event' ])` in app component makes all the components re-render for every single scroll event which may be expensive (especially if some have functions in their rendering)

## Acceptance criteria
- [ ] Move the scroll detection outside angular change detection scope (it only affects html, it can use basic visibility to hide/show based on position).
- [ ] Test that the behavior is exactly the same as now.
- [ ] Test that all components are not re-rendered anymore on scroll event (e.g. print the current time in a template)

Iloveall:
Fixes #578


smadbe (approved):
Seems to work and produce much less updates. 👍

PR #582, opened 2021-08-05
bugfix: ensure we do not expect children on chapter with can_view=info perm 

smadbe:
Fix a bug: infinite loading spinner in menu when the item has can_view:info but has children.

Add extra checks to prevent this issue (will also ask the backend this "fix" that)

### DIFF
--- a/src/app/core/http-services/item-navigation.service.ts
+++ b/src/app/core/http-services/item-navigation.service.ts
@@ -126,7 +126,7 @@ function createNavMenuItem(raw: {
   return {
     id: raw.id,
     title: raw.string.title ?? '',
-    hasChildren: raw.has_visible_children && raw.permissions.can_view !== 'none' && raw.permissions.can_view !== 'info',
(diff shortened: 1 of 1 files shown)


src/app/core/http-services/item-navigation.service.ts
```diff
@@ -126,7 +126,7 @@ function createNavMenuItem(raw: {
   return {
     id: raw.id,
     title: raw.string.title ?? '',
-    hasChildren: raw.has_visible_children,
+    hasChildren: raw.has_visible_children && raw.permissions.can_view !== 'none' && raw.permissions.can_view !== 'info',
```
Iloveall, line 129: Idea for improve - I prefer to use:

!['none', 'info'].includes(raw.permissions.can_view);

smadbe:
```suggestion
    hasChildren: raw.has_visible_children && ![ 'none', 'info' ].includes(raw.permissions.can_view),
```

Iloveall (approved):

PR #587, opened 2021-08-06
bugfix: make sure the 'via group ... ' indication is indicated for every root item, even non-chapters

smadbe:
The 'via group...' was not given for tasks, courses and leaf skills.

### DIFF
--- a/src/app/core/components/left-nav-tree/left-nav-tree.component.html
+++ b/src/app/core/components/left-nav-tree/left-nav-tree.component.html
@@ -132,6 +132,7 @@
           <span class="label-container">
             <span class="node-label-title">
               {{ node.data.element.title }}
+              <span *ngIf="node.data.element.groupName">(via group "{{ node.data.element.groupName }}")</span>
             </span>
           </span>
           <alg-skill-progress
@@ -228,6 +229,7 @@
           <span class="label-container">
             <span class="node-label-title">
               {{ node.data.element.title }}
+              <span *ngIf="node.data.element.groupName">(via group "{{ node.data.element.groupName }}")</span>
             </span>
           </span>
         </span>


src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -132,6 +132,7 @@
           <span class="label-container">
             <span class="node-label-title">
               {{ node.data.element.title }}
+              <span *ngIf="node.data.element.groupName">(via group "{{ node.data.element.groupName }}")</span>
             </span>
```
Iloveall, line 136: Must be translated?

Iloveall (changes requested):

src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -228,6 +229,7 @@
           <span class="label-container">
             <span class="node-label-title">
               {{ node.data.element.title }}
+              <span *ngIf="node.data.element.groupName">(via group "{{ node.data.element.groupName }}")</span>
```
Iloveall, line 232: Must be translated?

src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -132,6 +132,7 @@
...
```
Iloveall, line 135: Must be translated?

Iloveall (changes requested):

smadbe (commented):
Probably but that's out of scope of this PR as it is the same at the other place.
But this a temporary think while the designer creates something better so I don't think  that it is important for now.

Iloveall (approved):

PR #590, opened 2021-08-09
Hides user column when observing single user

### LINKED ISSUE #574: item log view: do not display the user column when watching a single user
## Overview
In item log view, it is not really useful to display the "user" column when the page is only related to 1 user.
Screenshot 2021-08-02 at 18 52 44

## Acceptance criteria
- [ ] if the watched group is a user, do not display the "user" column
- [ ] the condition ` ... [ 'Chapter', 'Task', 'Course' ].includes(type),` on that column looks strange. This part of the check can be removed, it makes no sense.

Iloveall:
Fixes #574

### DIFF
--- a/src/app/modules/item/pages/item-log-view/item-log-view.component.ts
+++ b/src/app/modules/item/pages/item-log-view/item-log-view.component.ts
@@ -6,7 +6,7 @@ import { distinct, switchMap, map } from 'rxjs/operators';
 import { mapToFetchState } from 'src/app/shared/operators/state';
 import { ItemType } from '../../../../shared/helpers/item-type';
 import { Item } from '../../http-services/get-item-by-id.service';
-import { UserSessionService } from '../../../../shared/services/user-session.service';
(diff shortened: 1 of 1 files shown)


src/app/modules/item/pages/item-log-view/item-log-view.component.ts
```diff
@@ -83,7 +83,7 @@ export class ItemLogViewComponent implements OnChanges, OnDestroy {
       {
         field: 'item.user',
         header: $localize`User`,
-        enabled: watchingGroupId && [ 'Chapter', 'Task', 'Course' ].includes(type),
+        enabled: watchingGroup && !watchingGroup.login,
```
smadbe, line 86: add a comment next it to explain what it means
```suggestion
        enabled: watchingGroup && !watchingGroup.login, // only if in watching mode and watch group is not a user
```

smadbe (approved):

PR #594, opened 2021-08-10
Adds disabled state for left nav with locked icon and tooltip

### LINKED ISSUE #583: In the left menu, display locks for "locked" items
## Overview

The same way a lock is shown in the item children list when the item cannot be explored/started (when can_view='info'), display a white lock on the right in the left menu when the user cannot participate in this item.

(example with a temp user (non authenticated user) in `/en/#/activities/by-id/4102;attempId=0;path=4702/details`)
Screenshot 2021-08-05 at 17 48 13

## Acceptance criteria
- [ ] Show a white lock icon on the right of item row with permission `can_view='info'` (for any type of item)
- [ ] On hover, display the tooltip: 
  - for chapter, skills: "Your current access rights do not allow you to list the content of this chapter/skill."
  - for task/course: "Your current access rights do not allow you to start the activity."

Iloveall:
Fixes #583

### DIFF
--- a/src/app/core/components/left-nav-tree/left-nav-tree.component.html
+++ b/src/app/core/components/left-nav-tree/left-nav-tree.component.html
@@ -59,6 +59,13 @@
             <span class="node-state" *ngIf="node.data.status === 'error'">
               <i class="fa fa-exclamation-triangle"></i>
             </span>
+            <span
+              class="node-state"
+              *ngIf="node.data.element.isLocked"
+              i18n-pTooltip pTooltip="{{ 'Your current access rights do not allow you to list the content of this chapter/skill.' }}"
+            >
+              <i class="fa fa-lock"></i>
+            </span>
           </span>
         </span>
       </span>
@@ -116,7 +123,7 @@
     <span
       id="nav-{{ node.data.element.id }}"
       class="node-tree-item folder"
-      [ngClass]="{ expanded: node.expanded, locked: node.locked, current: node.data && node.data.current }"
+      [ngClass]="{ expanded: node.expanded, locked: node.locked, current: node.data && node.data.current, disabled: node.data.element.disabled }"
     >
       <span class="node-item-branch">
         <span class="top-branch">
@@ -136,6 +143,13 @@
             <span class="node-label-title">
               {{ node.data.element.title }}
               <span *ngIf="node.data.element.groupName">(via group "{{ node.data.element.groupName }}")</span>
+            </span>
+            <span
+              class="node-state"
+              *ngIf="node.data.element.disabled"
+              i18n-pTooltip pTooltip="Your current access rights do not allow you to list the content of this chapter/skill."
+            >
+              <i class="fa fa-lock"></i>
             </span>
           </span>
           <alg-skill-progress
@@ -204,7 +218,7 @@
     <span
       id="nav-{{ node.data.element.id }}"
       class="node-tree-item folder"
-      [ngClass]="{ expanded: node.expanded, locked: node.locked, current: node.data && node.data.current }"
+      [ngClass]="{ expanded: node.expanded, locked: node.locked, current: node.data && node.data.current, disabled: node.data.element.disabled }"
     >
       <span class="node-item-branch">
         <span class="top-branch">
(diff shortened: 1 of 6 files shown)


smadbe (changes requested):
Unexpected behavior, if I go to "Parcours officiels\Parcours officiels\4LA 1ère année UE Informatique", via the left menu, then I cannot click on the other "locked" item in the left menu.

smadbe (changes requested):
The white selection background has transparency as well, this is not what we want.
Screenshot 2021-08-12 at 09 08 40
(even, the icon on the left and right should not be with opacity, just the text)

src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -59,6 +59,13 @@
             <span class="node-state" *ngIf="node.data.status === 'error'">
               <i class="fa fa-exclamation-triangle"></i>
             </span>
+            <span
+              class="node-state"
+              *ngIf="node.data.element.isLocked"
+              i18n-pTooltip pTooltip="{{ 'Your current access rights do not allow you to list the content of this chapter/skill.' }}"
```
smadbe, line 65:
```suggestion
              i18n-pTooltip pTooltip="{{ 'Your current access rights do not allow you to list the content of this chapter.' }}"
```

src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -137,6 +144,13 @@
               {{ node.data.element.title }}
               <span *ngIf="node.data.element.groupName">(via group "{{ node.data.element.groupName }}")</span>
             </span>
+            <span
+              class="node-state"
+              *ngIf="node.data.element.disabled"
+              i18n-pTooltip pTooltip="Your current access rights do not allow you to list the content of this chapter/skill."
```
smadbe, line 150:
```suggestion
              i18n-pTooltip pTooltip="Your current access rights do not allow you to list the content of this skill."
```

src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -116,7 +123,7 @@
     <span
       id="nav-{{ node.data.element.id }}"
       class="node-tree-item folder"
-      [ngClass]="{ expanded: node.expanded, locked: node.locked, current: node.data && node.data.current }"
+      [ngClass]="{ expanded: node.expanded, locked: node.locked, current: node.data && node.data.current, disabled: node.data.element.disabled }"
```
smadbe, line 126: "locked" seems not to be used anymore, probably it was done for this. 
I think I would prefer you reuse the "locked" class for this. 

And `node.data.element` is a `NavTreeElement`, it does not have any "disabled". It only works because the angular templating forgets about the type and makes everything 'any'.
So just introduce a new `locked` property in `NavTreeElement` and make sure it has a value for items and groups. (it may be made more clean later)

src/app/core/http-services/item-navigation.service.ts
```diff
@@ -88,7 +88,8 @@ export interface NavMenuItem {
   currentScore?: number,
   validated?: boolean,
   canViewContent: boolean,
-  children?: NavMenuItem[] // placeholder for children when fetched (may 'hasChildren' with 'children' not set)
+  children?: NavMenuItem[], // placeholder for children when fetched (may 'hasChildren' with 'children' not set)
+  disabled: boolean,
```
smadbe, line 92: `canViewContent` does already the same ... and it is not used so it can just be renamed.

smadbe (changes requested):

src/app/core/components/left-nav-tree/left-nav-tree.component.html
Iloveall: Fixed

src/app/core/http-services/item-navigation.service.ts
Iloveall: Fixed

src/app/core/components/left-nav-tree/left-nav-tree.component.html
```diff
@@ -31,7 +31,7 @@
           <span class="mask-branch"></span>
         </span>
       </span>
-      <span class="node-item-content" [ngClass]="{ unlocked: !node.locked }" (click)="selectNode(node)">
+      <span class="node-item-content" [ngClass]="{ unlocked: !node.data.element.locked }" (click)="selectNode(node)">
```
smadbe, line 34: What is "unlocked"? Some are locked but the other one should be "normal", I don't see why a need for a specific unlocked class.

smadbe (changes requested):

src/app/core/components/left-nav-tree/left-nav-tree.component.html
Iloveall: Fixed

smadbe (approved):
Looks good

PR #597, opened 2021-08-12
disallow backticks when not required

smadbe:
Does not allow backticks when they are not required. 
The name of this property was not clear but as "true", it allows to use backticks for any string.


Iloveall (approved):
Approved

PR #599, opened 2021-08-12
Feature/handle item with can watch

### LINKED ISSUE #585: Handle item with can_view:info
## Overview
Currently, when the current user cannot start an item (can_view:info), the page it gets on arriving on the page is a bit empty and unclear.

Screenshot 2021-08-05 at 18 22 37

## Acceptance criteria

When the current user has only can_view:info perm on an item:
- [ ] display, in the "Content" tab, on a white background (use something like `validation-text` class):
   - [ ] for chapters: "Your current access rights do not allow you to list the content of this chapter."
   - [ ] for skills: "Your current access rights do not allow you to list the content of this skill."
   - [ ] for task/course: "Your current access rights do not allow you to start the activity."

Iloveall:
Fixes #585

### DIFF
--- a/src/app/modules/item/pages/item-content/item-content.component.html
+++ b/src/app/modules/item/pages/item-content/item-content.component.html
@@ -1,13 +1,31 @@
 <ng-container *ngIf="itemData">
-  <ng-container *ngIf="itemData.currentResult">
+  <ng-container *ngIf="itemData.item.permissions.canView !== 'info'; else notAllowToWatch">
+    <ng-container *ngIf="itemData.currentResult">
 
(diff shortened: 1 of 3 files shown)


smadbe (changes requested):
The spec mentions "use something like validation-text class".
So I would have expected something like that: 
Screenshot 2021-08-13 at 15 23 16
rather than: 
Screenshot 2021-08-13 at 15 23 29

smadbe (approved):

smadbe:
Note that the title of this PR is misleading. "can watch" is something very different from "can view".

PR #602, opened 2021-08-16
Feature/observation suggestions of activities

### LINKED ISSUE #399: Observation: list suggestions of activities
## Motivations
Improve the "observation" mode on groups

## Subtasks
* When clicking on the eye when the group is not the one currently observed (so when activating observation mode)
  - display an primeng OverlayPanel (dismissable, with close icon) which will list some items to navigate to directly. (require a loading spinner)
  - for the moment, display in this list what is returned by the [activitiesView](tttps://france-ioi.github.io/algorea-devdoc/api/#operation/activitiesView ) service with the the current group as `watched_group_id` parameter. 

Note: it might be required to disable the current redirection which takes place when entering into observation mode.

Iloveall:
Fixes #399

### DIFF
--- a/src/app/modules/group/components/suggestion-of-activities/suggestion-of-activities.component.html
+++ b/src/app/modules/group/components/suggestion-of-activities/suggestion-of-activities.component.html
@@ -1,0 +1,25 @@
+<ng-container *ngIf="state$ | async as state">
+  <div class="spinner-container" *ngIf="state.isFetching">
+    <p-progressSpinner [style]="{width: '24px', height: '24px'}"></p-progressSpinner>
+  </div>
+  <div *ngIf="state.isError" i18n>Error while loading the activities</div>
+  <div *ngIf="state.isReady && state.data.length === 0" i18n>
+    There are no activities
+  </div>
+  <p-table
+      class="alg-table --overlay-panel"
+      [value]="state.data"
+      *ngIf="state.isReady && state.data.length > 0"
+  >
+    <ng-template pTemplate="body" let-item>
+      <tr>
+        <td><span class="type">{{ item.activity.type }}</span></td>
+        <td>
+          <a class="alg-link dots" [routerLink]="item.activity | rawItemLink">
+            {{ item.activity.string.title }} ({{ item.name }})
+          </a>
+        </td>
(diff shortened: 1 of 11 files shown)


src/app/modules/group/components/suggestion-of-activities/suggestion-of-activities.component.ts
```diff
@@ -0,0 +1,31 @@
...
+import { isNotUndefined } from '../../../../shared/helpers/null-undefined-predicates';
+import { mapToFetchState } from '../../../../shared/operators/state';
+
+@Component({
+  selector: 'alg-suggestion-of-activities',
+  templateUrl: './suggestion-of-activities.component.html',
+  styleUrls: [ './suggestion-of-activities.component.scss' ],
+})
+export class SuggestionOfActivitiesComponent {
+  readonly state$ = this.sessionService.watchedGroup$.pipe(
```
smadbe, line 14: Nice way to retrieve the id of clicked group without strange ways of passing the parameters! 👍

src/app/modules/group/components/suggestion-of-activities/suggestion-of-activities.component.html
```diff
@@ -0,0 +1,25 @@
...
+  <p-table
+      class="alg-table --overlay-panel"
+      [value]="state.data"
+      *ngIf="state.isReady && state.data.length > 0"
+  >
+    <ng-template pTemplate="body" let-item>
+      <tr>
+        <td><span class="type">{{ item.activity.type }}</span></td>
+        <td>
+          <a class="alg-link dots" [routerLink]="item.activity | rawItemLink">
```
smadbe, line 18: A small improvement... but you'll need my newly merged branch (so to rebase master).
Actually, here you know more that just the id of the item. As the item is a root, you know that its path is `[]`, this way you'll not have to call the service to find the path. 
So either you build a `RouteItem` and convert it to a link before passing it to the template, or you adapt your pipe to give the path... but in this case it is not a rawItemLink anymore...

... ok merge it as it and I'll do the improvement myself. I will be easier. :-)

smadbe (approved):
great 👍

PR #606, opened 2021-08-16
require tests to pass before deploying master: to prevent a merging error to break everything

smadbe:
For branches, they are deployed and it's not a problem (it allows testing faster), but for the "master" branch, let's not deploy it if one one test fail. At worst, it is false negative (the tests need to be re-run) but that's better than breaking the platform.


Iloveall (approved):
Approved

PR #614, opened 2021-08-18
No undefined-item history state

smadbe:
Do not keep an extra state when there is quick redirection because an item has its path/attempt fetched.
This way, the user can use the browser previous/next button without strange behavior.


Iloveall (approved):
Approved

PR #618, opened 2021-08-20
Feature/permission page first stage

### LINKED ISSUE #600: Permission page: First almost empty page
## Overview
The final goal is to have a page for each group which summarize all permissions given the group as shown here:
image

The goal of this first step is to have some base of this page.

## Acceptance criteria
- [ ] A new tab is added on the group page, it is only displayed if the current user has `can_grant_group_access` on the group or if he arrives on the page by other means (e.g. via a link), in which case an error is displayed.
- [ ] If the user sees the page while not having `can_grant_group_access` on the group, an error "You are not allowed to manage permissions of this group." is displayed.
- [ ] A section is shown (as in the design) "Activity associated with this group"
- [ ] Only one row is displayed (for now): the root activity of the group. 
- [ ] (fetch and) display the item title for group root activity.

Iloveall:
Fixes #600

### DIFF
--- a/src/app/modules/group/pages/group-access/group-access.component.html
+++ b/src/app/modules/group/pages/group-access/group-access.component.html
@@ -1,25 +1,27 @@
 <ng-container *ngIf="group?.currentUserCanGrantGroupAccess; else noAccess">
   <alg-section icon="fa fa-users" i18n-label label="Activity associated to {{ group?.name }}">
-    <ng-container *ngIf="itemState$ | async as itemState; else noActivity">
-      <div class="spinner" *ngIf="itemState.isFetching">
+    <ng-container *ngIf="rootActivityState$ | async as rootActivityState">
+      <div class="spinner" *ngIf="rootActivityState.isFetching">
         <p-progressSpinner [style]="{width: '3.3333rem', height: '3.3333rem'}"></p-progressSpinner>
       </div>
 
-      <div class="error alg-error-text" *ngIf="itemState.isError" i18n>
+      <div class="error alg-error-text" *ngIf="rootActivityState.isError" i18n>
         Error while loading the root activity.
       </div>
 
-      <ng-container *ngIf="itemState.isReady">
-        <div class="item">
-          <i class="icon fa fa-folder"></i>
-          <strong i18n>Root activity: {{ itemState.data.string.title }}</strong>
-        </div>
+      <ng-container *ngIf="rootActivityState.isReady">
+        <ng-container *ngIf="rootActivityState.data; else noActivity">
+          <div class="item">
+            <i class="icon fa fa-folder"></i>
+            <strong i18n>Root activity: {{ rootActivityState.data.string.title }}</strong>
+          </div>
+        </ng-container>
+
+        <ng-template #noActivity>
+          <div i18n>This group has no activity.</div>
+        </ng-template>
       </ng-container>
     </ng-container>
-
-    <ng-template #noActivity>
-      <div i18n>This group has no activity.</div>
-    </ng-template>
   </alg-section>
 </ng-container>
 
--- a/src/app/modules/group/pages/group-access/group-access.component.html
+++ b/src/app/modules/group/pages/group-access/group-access.component.html
@@ -1,0 +1,9 @@
+<ng-container *ngIf="group?.currentUserCanGrantGroupAccess; else noAccess">
+  <alg-section icon="fa fa-users" i18n-label label="Activity associated to {{ group?.name }}">
+    Permissions content
+  </alg-section>
+</ng-container>
+
+<ng-template #noAccess>
+  <p i18n>You are not allowed to manage permissions of this group.</p>
+</ng-template>
--- a/src/app/modules/group/pages/group-access/group-access.component.html
+++ b/src/app/modules/group/pages/group-access/group-access.component.html
@@ -1,10 +1,6 @@
 <ng-container *ngIf="group?.currentUserCanGrantGroupAccess; else noAccess">
   <alg-section icon="fa fa-users" i18n-label label="Activity associated to {{ group?.name }}">
-    <div *ngIf="!group?.rootActivityId" i18n>
-      This group has no activity.
-    </div>
-
-    <ng-container *ngIf="itemState$ | async as itemState">
+    <ng-container *ngIf="itemState$ | async as itemState; else noActivity">
       <div class="spinner" *ngIf="itemState.isFetching">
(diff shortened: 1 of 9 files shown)


smadbe (changes requested):
The usual tab issue when not allowed to see a tab: the tab should be shown when visiting it anyway (by a link for instance)
Screenshot 2021-08-23 at 19 04 34

Can you move the tab before 'settings' as well ? :-)

src/app/modules/group/pages/group-access/group-access.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  private readonly rootActivityId$ = new ReplaySubject<string | null>(1);
+
+  itemState$ = this.rootActivityId$.pipe(
+    filter(isNotNull),
+    switchMap(rootActivityId =>
+      this.getItemByIdService.get(rootActivityId).pipe(
+        mapToFetchState(),
+      )
+    ),
+  );
```
smadbe, line 26: This means that if you switch from a group with a root activity to one without, the previous root activity will stay.

You have to do something like that instead (to be checked):
```suggestion
  itemState$ = this.rootActivityId$.pipe(
    switchMap(rootActivityId =>
      rootActivityId === null ? of (null) :     this.getItemByIdService.get(rootActivityId).pipe(
        mapToFetchState(),
      )
    ),
  );
```
And handle the null in the template.

smadbe (changes requested):

src/app/modules/group/pages/group-access/group-access.component.ts
Iloveall: Fixed

src/app/modules/group/pages/group-access/group-access.component.html
```diff
@@ -1,10 +1,6 @@
...
   <alg-section icon="fa fa-users" i18n-label label="Activity associated to {{ group?.name }}">
-    <div *ngIf="!group?.rootActivityId" i18n>
-      This group has no activity.
-    </div>
-
-    <ng-container *ngIf="itemState$ | async as itemState">
+    <ng-container *ngIf="itemState$ | async as itemState; else noActivity">
       <div class="spinner" *ngIf="itemState.isFetching">
         <p-progressSpinner [style]="{width: '3.3333rem', height: '3.3333rem'}"></p-progressSpinner>
       </div>
```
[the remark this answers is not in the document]
Iloveall: I've sent the video

src/app/modules/group/pages/group-access/group-access.component.html
```diff
@@ -1,10 +1,6 @@
 <ng-container *ngIf="group?.currentUserCanGrantGroupAccess; else noAccess">
   <alg-section icon="fa fa-users" i18n-label label="Activity associated to {{ group?.name }}">
-    <div *ngIf="!group?.rootActivityId" i18n>
-      This group has no activity.
-    </div>
-
-    <ng-container *ngIf="itemState$ | async as itemState">
+    <ng-container *ngIf="itemState$ | async as itemState; else noActivity">
```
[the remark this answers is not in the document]
Iloveall: I have look off docs, noActivity will shown while itemState equal false state (null, undefined, "", 0). Also I have tested it so it works, it works even with filter operator.

https://angular.io/api/common/NgIf#storing-a-conditional-result-in-a-variable

Additional: https://prnt.sc/1qsyq1f

src/app/modules/group/pages/group-access/group-access.component.html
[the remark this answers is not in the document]
smadbe: Yes sorry. It's clearly visible for me as well.

src/app/modules/group/pages/group-access/group-access.component.html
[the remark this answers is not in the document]
smadbe: Right. It's me who mixed stuffs here. It is because you do not really use the FetchState as I would expect. See the next comment.

src/app/modules/group/pages/group-access/group-access.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  itemState$ = this.rootActivityId$.pipe(
+    switchMap(rootActivityId => {
+      if (!rootActivityId) {
+        return of(null);
+      }
+      return this.getItemByIdService.get(rootActivityId).pipe(
+        mapToFetchState()
+      );
+    })
+  );
```
smadbe, line 27: What confused me in my previous comment was that I would have expected:
```suggestion
  itemState$ = this.rootActivityId$.pipe(
    switchMap(rootActivityId => {
      if (!rootActivityId) {
        return of(null);
      }
      return this.getItemByIdService.get(rootActivityId);
    }),
    mapToFetchState()
  );
```
Your solution results in a `Observable<FetchState<Item>|null>` (either "null" or a ready/fetching/error state) while I expected a `Observable<FetchState<Item|null>>`, a state which is either error, fetching, or ready of either nothing (if no root activity for this group) or an item. 
For now, that's quite okay, but when we'll add more fetching combined with this one, probably that will not work well anymore.

src/app/modules/group/pages/group-access/group-access.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  selector: 'alg-group-access',
+  templateUrl: './group-access.component.html',
+  styleUrls: [ './group-access.component.scss' ],
+})
+export class GroupAccessComponent implements OnChanges {
+  @Input() group?: Group;
+
+  private readonly rootActivityId$ = new ReplaySubject<string | null>(1);
+
+  itemState$ = this.rootActivityId$.pipe(
```
smadbe, line 18: `itemState` as variable name is a bit strange. Either use `state` as at many places to consider that it is the state of component globally. Or, if we consider we may have multiple states later in this component, that's more `rootActivityState$`. `itemState` does not help understanding what it is for, it is just define the type which may be in the state.

smadbe (approved):

src/app/modules/group/pages/group-access/group-access.component.ts
Iloveall: Yes, I'm expecting multiple states, but `rootActivityState$` more clear. I'll fix

src/app/modules/group/pages/group-access/group-access.component.ts
Iloveall: Fixed

src/app/modules/group/pages/group-access/group-access.component.ts
Iloveall: Fixed

PR #623, opened 2021-08-25
Adds summary column for group progress grid

### LINKED ISSUE #612: Group progress: the parent item score column
## Overview
The group progress grid page shows the results of each members of the observed group on the visited item and its children. 
Screenshot 2021-08-18 at 14 37 03

The problem is that the current display, display the result on each child of the item, but not the score of the item itself. So let us add it. (this has been added to backend service already)

## Acceptance criteria
- [x] Add a new column on the right of the "name" header, so on the left of all the children. The style of the full column should be a bit different (darker?) as the other columns to show that's the summary column. 
- [x] Populate this column with what is received from the existing service (already deployed)

Iloveall:
Fixes #612

### DIFF
--- a/src/app/modules/item/pages/group-progress-grid/group-progress-grid.component.scss
+++ b/src/app/modules/item/pages/group-progress-grid/group-progress-grid.component.scss
@@ -1,3 +1,5 @@
+@import 'src/colors';
+
 :host {
 
   .loader {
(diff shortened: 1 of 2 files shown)


smadbe (changes requested):
Screenshot 2021-08-25 at 17 38 33

Bug: All the users have the same score on the parent.

ThomasJuster (approved):

smadbe (approved):

PR #627, opened 2021-08-26
add path to group link in member list

### LINKED ISSUE #650: Bug: Invalid descendant team link
As shown in the screenshot, the url contains "undefined". 
Screenshot 2021-08-31 at 17 40 19

ThomasJuster:
## Done

Forward `path` as input to member list component. Use it to construct the group url in the member list.

## Tests

1) Go to a group member list (see links below)
2) Click on `Sub groups`
3) Hover any group name

- With the [production app](https://dev.algorea.org/en/#/groups/by-id/672913018859223173;path=52767158366271444/details/members), you won't see any path in the url
- With the [branch app](https://dev.algorea.org/branch/fix/member-list-group-path/en/#/groups/by-id/672913018859223173;path=52767158366271444/details/members), you should see the path in the url

Incidentally, fixes #650


src/app/modules/group/components/member-list/member-list.component.html
```diff
@@ -58,9 +58,9 @@
             </ng-container>
             <ng-container *ngSwitchCase="'name'">
               <a
-                  class="alg-link"
-                  [routerLink]="[ '/', 'groups', 'by-id', rowData.id, 'details' ]"
-                  >
+                class="alg-link"
+                [routerLink]="getSubGroupUrl(rowData.id)"
```
smadbe, line 62: Don't we avoid at all cost function calls in templates?  
Actually I know that 's not as dramatic as I thought...
Anyway, as the path is exactly the same for all children (I assume?), shouldn't you just do a pipe which build this path based on the id and parent info?

smadbe (changes requested):
In "descendants", you do not know their full path, so you should not use the path.

ThomasJuster:
Was only the showcase of #643

src/app/modules/group/components/member-list/member-list.component.html
ThomasJuster: Fixed

smadbe (changes requested):
Unexpected group link for a user.
Screenshot 2021-09-14 at 12 03 03

PR #643, opened 2021-08-31
fix e2e tests

ThomasJuster:


ThomasJuster:
Was only the showcase of #645

PR #644, opened 2021-08-31
Refactor item datasource for using FetchState (and mapToFetchState)

smadbe:
Refactor item datasource for a proper usage of fetch states.

It was in preparation of more changes around the item datasource.

### DIFF
--- a/src/app/modules/item/services/item-datasource.service.ts
+++ b/src/app/modules/item/services/item-datasource.service.ts
@@ -1,6 +1,6 @@
 import { Injectable, OnDestroy } from '@angular/core';
 import { concat, EMPTY, forkJoin, merge, Observable, of, ReplaySubject, Subject, Subscription } from 'rxjs';
-import { map, switchMap } from 'rxjs/operators';
+import { map, shareReplay, switchMap } from 'rxjs/operators';
 import { bestAttemptFromResults, implicitResultStart } from 'src/app/shared/helpers/attempts';
 import { isRouteWithSelfAttempt, FullItemRoute } from 'src/app/shared/routing/item-route';
 import { ResultActionsService } from 'src/app/shared/http-services/result-actions.service';
@@ -28,7 +28,8 @@ export class ItemDataSource implements OnDestroy {
   /* state to put outputted */
   readonly state$ = this.fetchOperation$.pipe(
     switchMap(item => this.fetchItemData(item)),
-    mapToFetchState({ resetter: merge(this.refresh$, this.fetchOperation$) })
+    mapToFetchState({ resetter: merge(this.refresh$, this.fetchOperation$) }),
+    shareReplay(1),
   );
 
   private subscription: Subscription;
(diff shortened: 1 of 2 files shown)


src/app/modules/item/services/item-datasource.service.ts
```diff
@@ -22,11 +22,15 @@ export interface ItemData { route: FullItemRoute, item: Item, breadcrumbs: Bread
...
 
+  private readonly fetchOperation$ = new ReplaySubject<FullItemRoute>(1); // trigger item fetching
+  private readonly refresh$ = new Subject<void>();
+
   /* state to put outputted */
-  private state = new BehaviorSubject<FetchState<ItemData>>(fetchingState());
-  state$ = this.state.asObservable();
+  readonly state$ = this.fetchOperation$.pipe(
+    switchMap(item => this.fetchItemData(item)),
+    mapToFetchState({ resetter: merge(this.refresh$, this.fetchOperation$) })
```
ThomasJuster, line 31: Why do you need the `fetchOperation$` as resetter here ?
To me it's useless since the `fetchOperation$` subject is already the starting point of the observable, did you encounter issues when not providing it as resetter ?

src/app/modules/item/services/item-datasource.service.ts
```diff
@@ -22,11 +22,15 @@ export interface ItemData { route: FullItemRoute, item: Item, breadcrumbs: Bread
...
+  );
```
ThomasJuster, line 32: When refactoring the group data source, I had to add a `shareReplay(1)` operator
As is, if a component subscribes to the datasource _after_ the fetch has succeeded, the component won't have any value

I tested switching between edit and normal mode on [this item](https://dev.algorea.org/branch/use-fetchstate-in-itemdatasource/en/#/activities/by-id/1980584647557587953;path=4702,4102;attempId=0/details), it apparently works but I suspect it is because the item is refetched.

Should this `shareReplay(1)` + double call issue be addressed in a another issue+PR ?

src/app/modules/item/services/item-datasource.service.ts
smadbe: Otherwise, on navigation, the former item is displayed while fetching the next one. Actually, that was the initial goal of the resetter, more than for refreshing (otherwise it would have been called reloader/refresher), but at the end it has not been used for this usage yet. If you think about it, the `mapToFetchState` has no other way the `this.fetchOperation$` has changed... it just knows the output of the `switchMap`, which comes with a delay.

src/app/modules/item/services/item-datasource.service.ts
smadbe: Ok, I can do this.
What you call "double call" is the fact that get-item is called in item-edit (why probably it shouldn't, except to be sure we work on the same version)?
Actually, I am thinking about moving this item-datasource to something private to item-by-id and then just pass the item by input. This is a remaining from a time where I had not better way to pass the item as the subpage are served by routing.

smadbe:
TODO: check that there are no double calls

src/app/modules/item/services/item-datasource.service.ts
smadbe: fixed, and it fix multiple calls

ThomasJuster: Weird, I have 2 cancelled calls at item init, do you have an idea of where it could come from ?
image

ThomasJuster (commented):
I added a comment about 2 cancelled calls at item page startup, those are not in production.
I think we should investigate where they come from, what do you think ?

PR #652, opened 2021-09-01
Fetch item navigation data in item-by-id

smadbe:
Fetch the item navigation data in item-by-id when the result is known.
This makes a 3rd call to the navigation service when navigating on an item, but of course, the goal is remove this from the menu in a next stage, and then remove it from the navigator as well. It will be done by publishing the nav data to the current content info.

### DIFF
--- a/src/app/core/http-services/item-navigation.service.ts
+++ b/src/app/core/http-services/item-navigation.service.ts
@@ -6,6 +6,11 @@ import { bestAttemptFromResults } from 'src/app/shared/helpers/attempts';
 import { isRouteWithSelfAttempt, FullItemRoute } from 'src/app/shared/routing/item-route';
 import { appConfig } from 'src/app/shared/helpers/config';
 import { isASkill, isSkill, ItemType, ItemTypeCategory } from 'src/app/shared/helpers/item-type';
+import { decodeSnakeCase } from 'src/app/shared/operators/decode';
+import { pipe } from 'fp-ts/lib/function';
+import * as D from 'io-ts/Decoder';
+import { permissionsDecoder } from 'src/app/modules/item/helpers/item-permissions';
+import { dateDecoder } from 'src/app/shared/helpers/decoders';
 
 interface ItemStrings {
   title: string|null,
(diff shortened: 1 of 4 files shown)


src/app/core/http-services/item-navigation.service.ts
```diff
@@ -6,6 +6,11 @@ import { bestAttemptFromResults } from 'src/app/shared/helpers/attempts';
 import { isRouteWithSelfAttempt, FullItemRoute } from 'src/app/shared/routing/item-route';
 import { appConfig } from 'src/app/shared/helpers/config';
 import { isASkill, isSkill, ItemType, ItemTypeCategory } from 'src/app/shared/helpers/item-type';
+import { decodeSnakeCase } from 'src/app/shared/operators/decode';
+import { pipe } from 'fp-ts/lib/function';
```
ThomasJuster, line 10:
```suggestion
import { pipe } from 'fp-ts/function';
```

src/app/shared/operators/build-up.spec.ts
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable no-multi-spaces */
+import { EMPTY, of } from 'rxjs';
+import { delay } from 'rxjs/operators';
+import { TestScheduler } from 'rxjs/testing';
+import { buildUp } from './build-up';
+
+describe('progressiveListFromList operator', () => {
```
ThomasJuster, line 7: `progressiveListFromList` hum :thinking: :smile:

src/app/shared/operators/build-up.ts
```diff
@@ -0,0 +1,26 @@
...
+export function buildUp<T extends ObjectT, U extends ObjectT>(project: (value: T) => Observable<U>): OperatorFunction<T, T | (T & U)> {
+  return pipe(
+    switchMap(initialValue => concat(
+      of(initialValue),
+      project(initialValue).pipe(
+        map(additionalValue => ({ ...initialValue, ...additionalValue }))
+      ),
+    ))
+  );
+}
```
ThomasJuster, line 26: I understand the interest of enriching/projecting new data on the source observable data, but I don't understand the interest of emitting the initial value(s)
To me it adds complexity to the operator directly by implying a union type and potential extra computations

For instance if you `buildUp()` more than twice, let's say 5 times, the output type will be unbearable to understand and eventually to check

Why is emitting the initial value(s) needed ?

src/app/core/http-services/item-navigation.service.ts
smadbe: What's the diff?

src/app/shared/operators/build-up.ts
smadbe: That's all the point of the operator 😁, to hide this messy part of emitting the initial value. 
Actually as you saw the example in get-item-by-id and get-group-by-id, you might get the point.
So for items, we want 4 info: the breadcrumb, the item itself, the results and the nav data.
We first fetch (in parallel) breadcrumb and the item itself as "results" needs info from the previous ones, and nav data needs results. However, we do not want to wait for the 4 requests to complete (3 of them in cascade) before displaying anything, most of the interesting information is already there after the first set of 2 requests. So we emit these 2 values, use them in the interface, and in parallel the whole set of data is "build up" on top of that...
So we need 2 emissions for build up, the first one immediately with the data, then the other one, asynced, based on this data.

src/app/shared/operators/build-up.spec.ts
smadbe: fixed

src/app/core/http-services/item-navigation.service.ts
ThomasJuster: https://github.com/France-ioi/AlgoreaFrontend/pull/455#issuecomment-852823351

src/app/shared/operators/build-up.ts
ThomasJuster: I still doubt the idea because emitting new values of the state will trigger unneeded re-renders
To me, either the states are deeply coupled and only usable as a whole, either they should be split into different smaller states which you can subscribe to separately

Knowing that and since this subject is only a matter of opinion, if it is still what you want then fine by me

ThomasJuster (approved):

src/app/shared/operators/build-up.ts
ThomasJuster: I still doubt the idea because emitting new values of the state will trigger unneeded re-renders
To me, either the states are deeply coupled and only usable as a whole, either they should be split into different smaller states which you can subscribe to separately

Knowing that and since this subject is only a matter of opinion, if it is still what you want then fine by me

ThomasJuster: Conversation to resolve after reading

src/app/core/http-services/item-navigation.service.ts
ThomasJuster: Conversation to resolve after reading

PR #655, opened 2021-09-02
fetch group nav data in group datasource

### LINKED ISSUE #653: Fetch group navigation data in group-by-id
Do the same as done in #652 for groups. (for the same objectives)

ThomasJuster:
Fixes #653

## Done

- [x] add method `getGroupNavigation(groupId: string)` to `group-navigation` service
- [x] call `getGroupNavigation` in `group-datasource` service
- [x] update `state$` in `group-by-id` component

## Tests

1. Go to a group page (links below)
2. Check out the requests in the console

With the [branch app](https://dev.algorea.org/branch/feat/fetch-group-nav-data-in-group-datasource/en/#/groups/by-id/7650735426882403033;path=52767158366271444,672913018859223173/details), you should see 4 requests to navigation
With the [production app](https://dev.algorea.org/en/#/groups/by-id/7650735426882403033;path=52767158366271444,672913018859223173/details), you should see 3 requests to navigation

### DIFF
--- a/src/app/modules/group/services/group-datasource.service.ts
+++ b/src/app/modules/group/services/group-datasource.service.ts
@@ -1,11 +1,16 @@
 import { Injectable, OnDestroy } from '@angular/core';
-import { ReplaySubject, Subject } from 'rxjs';
+import { forkJoin, ReplaySubject, Subject } from 'rxjs';
 import { map, shareReplay, switchMap } from 'rxjs/operators';
+import { GroupNavigation, GroupNavigationService } from 'src/app/core/http-services/group-navigation.service';
 import { mapToFetchState } from 'src/app/shared/operators/state';
 import { GroupRoute } from 'src/app/shared/routing/group-route';
 import { GetGroupByIdService, Group } from '../http-services/get-group-by-id.service';
 
-export interface GroupData { route: GroupRoute, group: Group }
+export interface GroupData {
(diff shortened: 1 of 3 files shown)


src/app/modules/group/services/group-datasource.service.ts
```diff
@@ -22,15 +27,19 @@ export class GroupDataSource implements OnDestroy {
   state$ = this.fetchOperation.pipe(
     // switchMap does cancel the previous ongoing processing if a new one comes
     // on new fetch operation to be done: set "fetching" state and fetch the data which will result in a ready or error state
-    switchMap(route => this.getGroupByIdService.get(route.id).pipe(
-      map(group => ({ route, group })),
+    switchMap(route => forkJoin({
+      group: this.getGroupByIdService.get(route.id),
+      navigation: this.groupNavigationService.getGroupNavigation(route.id),
```
smadbe, line 32: We do not need/want to wait for the nav data to display the group page, so either we need to use `buildUp`, but it means the 2nd fetch will not start before the first one has responded, or we "buildUp" the answer without waiting, so using the `zip` operator... but without emitting the nav while the get-group-by-id has not been received... we need a "buildUpParallel" operator 😁
Ok, we can do this later...

smadbe (approved):

PR #657, opened 2021-09-02
group path: avoid failing when path from root fails

ThomasJuster:
## Done

When the group `path-from-root` fails, redirect to same page with `path: []`
Note: for a content route to be displayed, the `path` is mandatory, that's why I pass an empty array rather than `undefined`. If this behavior is incorrect, please notice me :pray: 

## Tests

I did not find a group for which `path-from-root` fails, where can I find one :pray: ?


smadbe (approved):
I don't see any example when it may fail (except a backend bug).

PR #658, opened 2021-09-02
fix user links

ThomasJuster:
## Done

- Retreat all user links:
  - left nav tree
  - breadcrumbs
  - profile link
- allow to give user routes to `urlArrayForGroupRoute` and `group-router`
- create `userLink` pipe (already created in another [PR to be merged](https://github.com/France-ioi/AlgoreaFrontend/pull/627))

## Tests

This is a refactoring PR, therefore the production and branch apps should behave exactly the same

- [x] Breadcrumbs: [production app](https://dev.algorea.org/en/#/groups/users/670968966872011405) / [branch app](https://dev.algorea.org/branch/fix/user-links/en/#/groups/users/670968966872011405)
  1. Click on the breadcrumb
  2. You should have navigated to the progress page
- [x] Profile link: [production app](https://dev.algorea.org/en/#/groups/by-id/7650735426882403033;path=52767158366271444,672913018859223173/details) / [branch app](https://dev.algorea.org/branch/fix/user-links/en/#/groups/by-id/7650735426882403033;path=52767158366271444,672913018859223173/details)
  1. Click on the user login at the top right of the screen
  2. Select option "profile"
  3. You should be redirected to the user page "personal datas"
- [x] left nav tree for group links: [production app](https://dev.algorea.org/branch/fix/user-links/en/#/groups/by-id/52767158366271444;path=/details) / [branch app](https://dev.algorea.org/en/#/groups/by-id/52767158366271444;path=/details)
  1. Click on a subgroup
  2. You should have navigated to the subgroup, the path should have been provided
- [ ] left nav tree for user links:
  - I did not find any case where the node was a user.


src/app/shared/pipes/groupLink.ts
```diff
@@ -4,12 +4,12 @@ import { RawGroupRoute } from '../routing/group-route';
...
 
 /**
- * Functions using item route should always be preferred to raw item.
- * Using raw item means further requests will be required to fetch path and attempt information.
+ * Pipe to format group links, with or without the path
+ * Since a user is considered as a group of one person, user routes are formatted using this pipe too.
  */
 @Pipe({ name: 'groupLink', pure: true })
 export class GroupLinkPipe implements PipeTransform {
   transform(route: RawGroupRoute, isUser?: boolean): UrlCommand {
```
smadbe, line 12: BTW, why a `isUser?` which supports then 3 values: `true`, `false`, `undefined` while I suppose there is only the semantic for 2 values. If it is false when not set, just use a default value for the parameter.

src/app/shared/routing/group-route.ts
```diff
@@ -23,14 +24,27 @@ export function rawGroupRoute(id: GroupId): RawGroupRoute {
...
+export type UserPage = 'personal-data';
+export type GroupPage = 'edit' | 'details';
+
 /**
  * Return a url array (`commands` array) to the given group, on the given page.
  */
-export function urlArrayForGroupRoute(route: RawGroupRoute, page: 'edit' | 'details' = 'details', isUser = false): UrlCommand {
+export function urlArrayForGroupRoute(
+  route: RawGroupRoute,
+  { isUser, page }: { isUser?: boolean, page?: GroupPage | UserPage } = {},
```
smadbe, line 35: The 2 parameters are not independent at all, that's a bit annoying.
Anyway, as the final goal is probably to have the page become a `string|string[]` as in items, probably no need to loose time with that.

smadbe (changes requested):

src/app/core/components/left-nav-tree/left-nav-tree.component.ts
```diff
@@ -90,11 +89,7 @@ export class LeftNavTreeComponent implements OnChanges {
     const routeBase = { id: node.data.element.id, path: node.data.path };
     switch (this.elementType) {
       case 'group':
-        if (node.data?.element?.type === 'User') {
-          void this.router.navigate([ '/', 'groups', 'users', routeBase.id ]);
-        } else {
-          this.groupRouter.navigateTo({ ...routeBase, contentType: 'group' });
-        }
+        this.groupRouter.navigateTo(groupRoute(routeBase.id, routeBase.path), { isUser: node.data?.element?.type === 'User' });
```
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/modules/group/components/group-log-view/group-log-view.component.ts
```diff
@@ -11,7 +12,7 @@ interface Column {
 
 interface Data {
   columns: Column[],
-  rowData: ActivityLog[]
+  rowData: (ActivityLog & { route?: RawGroupRoute })[]
```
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/modules/group/pages/user/user.component.ts
```diff
@@ -65,7 +66,9 @@ export class UserComponent implements OnInit, OnDestroy {
               path: [
                 {
                   title,
-                  navigateTo: this.router.createUrlTree([ 'groups', 'users', this.route.snapshot.params.id ]),
+                  navigateTo: this.router.createUrlTree(
+                    urlArrayForGroupRoute(rawGroupRoute(this.route.snapshot.params.id), { isUser: true })
+                  ),
```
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/shared/pipes/groupLink.ts
ThomasJuster: Fixed

src/app/shared/routing/group-route.ts
ThomasJuster: Fixed, I marked it as `string` for now

src/app/shared/routing/group-route.ts
```diff
@@ -15,22 +17,49 @@ export interface GroupRouteError {
   id?: string;
 }
 
-export function groupRoute(id: GroupId, path: string[]): GroupRoute {
-  return { contentType: 'group', id: id, path };
+
+export type GroupLike =
+  | { id: Group['id'], type?: string }
```
Iloveall, line 22: Do you need | before first type?

src/app/shared/routing/group-route.ts
```diff
@@ -15,22 +17,49 @@ export interface GroupRouteError {
   id?: string;
 }
 
-export function groupRoute(id: GroupId, path: string[]): GroupRoute {
-  return { contentType: 'group', id: id, path };
+
+export type GroupLike =
```
Iloveall, line 21: Probably could be on one line as length is not more than 140

src/app/shared/routing/group-route.ts
ThomasJuster: No, this notation is mainly for further iterations and/or readability so you can compare the union-ed types in a glimpse
But actually the type probably won't change

Should I still change this type and make it one-line ?

Iloveall (changes requested):
Small comments

Iloveall (approved):

src/app/shared/routing/group-route.ts
```diff
@@ -16,33 +17,47 @@ export interface GroupRouteError {
...
+  | { id: Group['id'], type?: string }
+  | Pick<User, 'groupId' | 'login'>;
+
+function isUser(group: GroupLike): boolean {
+  return ('login' in group && !!group.login) || ('type' in group && group.type === 'User');
```
smadbe, line 26: Which means that if you have a group without login and type, you might classify a group as non-user while you have no guarantee.  This is quite misleading as a dev might use one of the pipe/function thinking that it is handled correctly while some data is missing to determine whether it is a user or not.
It might be that the `GroupLike` type is incomplete... it might require a third option with the "caller" forces to user or non-user because he knows it.

src/app/shared/routing/group-route.ts
```diff
@@ -16,33 +17,47 @@ export interface GroupRouteError {
...
 }
 
-export type UserPage = 'personal-data';
-export type GroupPage = 'edit' | 'details';
+export function rawGroupRoute(group: GroupLike): RawGroupRoute {
+  const groupId = 'id' in group ? group.id : group.groupId;
+  return { contentType: 'group', id: groupId, isUser: isUser(group) };
+}
+
+export function isRawGroupRoute(route?: any): route is RawGroupRoute {
```
smadbe, line 39: should be `unknown` I assume.

smadbe (changes requested):

src/app/shared/routing/group-route.ts
ThomasJuster: Fixed

src/app/shared/routing/group-route.ts
ThomasJuster: Fixed

src/app/shared/routing/group-route.ts
```diff
@@ -15,22 +17,49 @@ export interface GroupRouteError {
...
+export type GroupLike =
+  | { id: Group['id'], isUser: boolean }
+  | { id: Group['id'], type: string }
+  | Pick<User, 'groupId' | 'login'>;
+
+function isUser(group: GroupLike): boolean {
+  return ('isUser' in group && group.isUser) || ('login' in group && !!group.login) || ('type' in group && group.type === 'User');
+}
+
+export function rawGroupRoute(group: GroupLike, forceIsUser?: boolean): RawGroupRoute {
```
smadbe, line 30: I don't see the point of `forceIsUser` while it can already be forced somehow (with `isUser`) in `GroupLike`.
It is not even used.

smadbe (approved):

PR #659, opened 2021-09-02
Feat: display group breadcrumbs

ThomasJuster:
## Done

- Create service `GetGroupBreadcrumbs`
- Add `breadcrumbs` property to `group-datasource` `GroupData` and fetch breadcrumbs when fetching a group
- Register breadcrumbs to display in component `group-by-id`

## Tests

- [x] When constructing a path manually by navigating in the app
:one: Wherever on the app ([production]() / [branch]()), click on `Groups` tab
:two: Navigate to groups: `4LTDI1819` > `Pixal` > `demo`
Then:
:arrow_right: You should see in the breadcrumbs `4LTDI1819`, `Pixal` and `demo`, `demo` should be selected
:arrow_right: Clicking on `Pixal` should redirect to the `Pixal` group, `4LTDI1819` remaining in the breadcrumbs
:arrow_right: Clicking on `4LTDI1819` should redirect to the group

- [x] When landing on a group with path constructed by backend by calling `path-from-root`:
:one: Copy-paste the direct url of this group `demo` group: [production](https://dev.algorea.org/en/#/groups/by-id/4603869706499321180/details) / [branch](https://dev.algorea.org/branch/feat/display-group-breadcrumbs/en/#/groups/by-id/4603869706499321180/details)
Then:
:arrow_right: You should see in the breadcrumbs `4LTDI1819`, `Pixal` and `demo`, `demo` should be selected
:arrow_right: Clicking on `Pixal` should redirect to the group with `4LTDI1819` remaining in the breadcrumbs
:arrow_right: Clicking on `4LTDI1819` should redirect to the group

### DIFF
--- a/src/app/modules/group/pages/group-details/group-details.component.ts
+++ b/src/app/modules/group/pages/group-details/group-details.component.ts
@@ -15,6 +15,7 @@ export class GroupDetailsComponent {
   state$ = this.groupDataSource.state$.pipe(mapStateData(state => ({
     group: withManagementAdditions(state.group),
     route: state.route,
+    breadcrumbs: state.breadcrumbs,
   })));
   fullFrameContent$ = this.layoutService.fullFrameContent$;
 
--- a/src/app/modules/group/pages/group-details/group-details.component.ts
+++ b/src/app/modules/group/pages/group-details/group-details.component.ts
@@ -13,10 +13,8 @@ import { LayoutService } from '../../../../shared/services/layout.service';
 export class GroupDetailsComponent {
 
   state$ = this.groupDataSource.state$.pipe(mapStateData(state => ({
+    ...state,
     group: withManagementAdditions(state.group),
-    route: state.route,
-    navigation: state.navigation,
-    breadcrumbs: state.breadcrumbs,
   })));
   fullFrameContent$ = this.layoutService.fullFrameContent$;
 
--- a/src/app/modules/group/services/group-datasource.service.ts
+++ b/src/app/modules/group/services/group-datasource.service.ts
@@ -1,11 +1,16 @@
 import { Injectable, OnDestroy } from '@angular/core';
-import { ReplaySubject, Subject } from 'rxjs';
+import { forkJoin, ReplaySubject, Subject } from 'rxjs';
 import { map, shareReplay, switchMap } from 'rxjs/operators';
 import { mapToFetchState } from 'src/app/shared/operators/state';
 import { GroupRoute } from 'src/app/shared/routing/group-route';
+import { GetGroupBreadcrumbsService, GroupBreadcrumb } from '../http-services/get-group-breadcrumbs.service';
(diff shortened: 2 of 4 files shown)


src/app/modules/group/pages/group-details/group-details.component.ts
```diff
@@ -15,6 +15,7 @@ export class GroupDetailsComponent {
   state$ = this.groupDataSource.state$.pipe(mapStateData(state => ({
     group: withManagementAdditions(state.group),
     route: state.route,
+    breadcrumbs: state.breadcrumbs,
```
smadbe, line 18: If the breadcrumb is not used in groups-details, it should not be in the state.

src/app/modules/group/services/group-datasource.service.ts
```diff
@@ -22,15 +27,19 @@ export class GroupDataSource implements OnDestroy {
   state$ = this.fetchOperation.pipe(
     // switchMap does cancel the previous ongoing processing if a new one comes
     // on new fetch operation to be done: set "fetching" state and fetch the data which will result in a ready or error state
-    switchMap(route => this.getGroupByIdService.get(route.id).pipe(
-      map(group => ({ route, group })),
+    switchMap(route => forkJoin({
+      group: this.getGroupByIdService.get(route.id),
+      breadcrumbs: this.getGroupBreadcrumbsService.getBreadcrumbs(route),
+    }).pipe(
+      map(({ group, breadcrumbs }) => ({ route, group, breadcrumbs })),
```
smadbe, line 34: I'll have a serious conflict with master on merge, so you should rebase/merge first.

src/app/modules/group/pages/group-details/group-details.component.ts
ThomasJuster: I was replicating what was done with item, but it's probably not a good idea
However removing it from the state is not satisfying either because the `GroupData` interface from `group-datasource` is used by group components and coupling their `groupData` type to a page state type is also a bad idea.
I see 2 viable solutions:
1. Enhance the group with management additions, and spread the rest of the state as is to avoid breaking the type instead of specifying each property one by one
2. Make subcomponents having a `groupData` input  retrieve the group data value from the data source and not as an input anymore

ThomasJuster: Although I think option 2 is more appropriate, it's not done this way in other parts of the app. So I opted for option 1.
Fixed

smadbe: Yes, as I wrote somewhere else: Option 2 is only possible because the datasource was thought to bypass a technical issue of passing arg to sub-components built by routing. But now that is not required anymore, DS should be private to group-by-id. So opt2 should not be used as it is a way to use a global variable for storing the group (considering there is only 1 group in the app at a given moment).

smadbe (approved):

PR #662, opened 2021-09-06
Feature/group administration page allow removal

### LINKED ISSUE #82: Group administration page: allow removal
## Overview
Allow removal of managers.

## Acceptance criteria

In the group manager page:

- [x] add selection, including a "select all". It should work like in "composition" tab
- [x] removal button like in "composition" tab (button should be greyed out if not selection)
- [x] require the current user to have "memberships_and_group" management level on the current group to see this button
- removal management like in composition tab: 
   - [x] UI frozen while fetching, 
   - [x] error/success result shown in a toast
   - [x] refresh of data after a change
 
Beware that for now managers cannot be re-added. You should probably add some using the service directly (in curl) when testing

Iloveall:
Fixes #82

### DIFF
--- a/src/app/modules/group/components/group-manager-list/group-manager-list.component.html
+++ b/src/app/modules/group/components/group-manager-list/group-manager-list.component.html
@@ -18,6 +18,7 @@
     >
       <ng-template #headerTemplate let-columns>
         <tr style="border: none;">
+          <th class="thin"></th>
           <th style="min-width: 160px;">
             <div class="slanted-header">
               <div class="slanted-header-content" i18n>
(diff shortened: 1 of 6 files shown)


src/app/modules/group/components/group-manager-list/group-manager-list.component.html
```diff
@@ -5,15 +5,20 @@
 <ng-container *ngIf="state === 'ready'">
```
ThomasJuster, line 5: Tested on [this group](https://dev.algorea.org/branch/feature/group-administration-page-allow-removal/en/#/groups/by-id/49268055312628123;path=/details/managers)

When I don't have the rights to remove managers, I can select managers but do nothing with the selection:
image

Apparently it works exactly the on the composition tab, I doubt this is an expected behavior.
I think it would be better to show selection only when actions are available, for both managers and group composition

Footnote: Since the group composition is not the subject of this PR, it could be fixed in another one I think, I'll create the issue

ThomasJuster (changes requested):

src/app/modules/group/components/group-manager-list/group-manager-list.component.html
Iloveall: Fixed

ThomasJuster (approved):

PR #671, opened 2021-09-07
make task platform an injectable service

ThomasJuster:


ThomasJuster:
Close in favor of #673

PR #676, opened 2021-09-10
Feature/allow removing sub groups

### LINKED ISSUE #334: Allow removing subgroups
## Motivations
Now we can add subgroup, we should be able to remove subgroups

## Subtasks

In group composition page, for "direct children" and "all descendants", for "sub-groups", "sessions" and "teams":
- (same as in user list) let the user select one, all or several rows 
- (same as in user list) add a "trash" icon + "Remove from group" bottom-right of the table, which is greyed out when there is no selection
- on click on "Remove from group", display a primeng ConfirmPopup with "?" icon and msg "Do you want to also delete the selected group(s)? (will only work if those are empty)"
- on "no": display a primeng ConfirmPopup with "!" icon and msg "By removing "groupA", "groupB" from the group, you may loose manager access to them (if no explicit permission or through other parent group). Are you sure you want to proceed?"
- on "yes": display a primeng ConfirmPopup with "!" icon and msg "Are you sure you want to permanently delete "groupA", "groupB"? This operation cannot be undone."
- depending on the decision, you call "removeGroup" or [groupRemoveChild](https://france-ioi.github.io/algorea-devdoc/api/#operation/groupRemoveChild) (ignore the orphan thing) service on each group. Grey out action and table while processing.
- on success/failure/partial success, display appropriate message as done in pending request component. If the group could not be deleted because it was not empty, display an appropriate message.
- the table needs to be refresh after every is done (same as in user list)

Iloveall:
Fixes #334

### DIFF
--- a/src/app/core/app.component.html
+++ b/src/app/core/app.component.html
@@ -49,7 +49,7 @@
 </div>
 <p-toast></p-toast>
 <p-confirmDialog></p-confirmDialog>
-<p-confirmPopup key="commonPopup"></p-confirmPopup>
+<p-confirmPopup key="commonPopup" styleClass="common-confirm-popup"></p-confirmPopup>
 <p-dialog
   *ngrxLet="fatalError$"
   i18n-header="@@auth-error-modal-title" header="Error"
--- a/src/app/modules/group/components/member-list/member-list.component.ts
+++ b/src/app/modules/group/components/member-list/member-list.component.ts
@@ -263,7 +263,8 @@ export class MemberListComponent implements OnChanges, OnDestroy {
       target: event.target || undefined,
       key: 'commonPopup',
       icon: 'pi pi-exclamation-triangle',
-      message: $localize`Are you sure you want to permanently delete ${getSelectedGroupChildCaptions(this.selection as GroupChild[])}? This operation cannot be undone.`,
+      message: $localize`Are you sure you want to permanently delete ${getSelectedGroupChildCaptions(this.selection as GroupChild[])}?
+       This operation cannot be undone.`,
       acceptLabel: $localize`Yes`,
       acceptIcon: 'fa fa-check',
       rejectLabel: $localize`No`,
@@ -276,7 +277,8 @@ export class MemberListComponent implements OnChanges, OnDestroy {
       target: event.target || undefined,
       key: 'commonPopup',
       icon: 'pi pi-exclamation-triangle',
-      message: $localize`By removing ${getSelectedGroupChildCaptions(this.selection as GroupChild[])} from the group, you may loose manager access to them (if no explicit permission or through other parent group). Are you sure you want to proceed?`,
+      message: $localize`By removing ${getSelectedGroupChildCaptions(this.selection as GroupChild[])} from the group, you may loose
+       manager access to them (if no explicit permission or through other parent group). Are you sure you want to proceed?`,
       acceptLabel: $localize`Yes`,
       acceptIcon: 'fa fa-check',
       rejectLabel: $localize`No`,
--- a/src/app/modules/group/components/member-list/member-list.component.ts
+++ b/src/app/modules/group/components/member-list/member-list.component.ts
@@ -1,5 +1,5 @@
 import { Component, Input, OnChanges, OnDestroy, SimpleChanges, ViewChild } from '@angular/core';
-import { SortEvent } from 'primeng/api';
+import { ConfirmationService, SortEvent } from 'primeng/api';
 import { Table } from 'primeng/table';
 import { Observable, ReplaySubject, Subject } from 'rxjs';
 import { map, switchMap } from 'rxjs/operators';
@@ -13,6 +13,13 @@ import { GroupUsersService, parseResults } from '../../http-services/group-users
 import { GroupData } from '../../services/group-datasource.service';
 import { Filter, GroupCompositionFilterComponent, TypeFilter } from '../group-composition-filter/group-composition-filter.component';
 import { displayResponseToast } from './user-removal-response-handling';
+import { displayGroupResponseToast } from './group-removal-response-handling';
+import { RemoveSubgroupService } from '../../http-services/remove-subgroup.service';
+import { RemoveGroupService } from '../../http-services/remove-group.service';
+
+function getSelectedGroupChildCaptions(selection: GroupChild[]): string {
+  return selection.map(selected => selected.name).join(', ');
+}
 
 interface Column {
   sortable?: boolean,
@@ -71,7 +78,7 @@ export class MemberListComponent implements OnChanges, OnDestroy {
   currentSort: string[] = [];
   currentFilter: Filter = this.defaultFilter;
 
-  selection: Member[] = [];
+  selection: (Member | GroupChild)[] = [];
 
   data: Data = {
     columns: [],
@@ -90,6 +97,9 @@ export class MemberListComponent implements OnChanges, OnDestroy {
     private getGroupDescendantsService: GetGroupDescendantsService,
     private groupUsersService: GroupUsersService,
     private actionFeedbackService: ActionFeedbackService,
+    private removeSubgroupService: RemoveSubgroupService,
+    private confirmationService: ConfirmationService,
+    private removeGroupService: RemoveGroupService,
   ) {
     this.dataFetching.pipe(
       switchMap(params => this.getData(params.route, params.filter, params.sort).pipe(mapToFetchState())),
@@ -97,6 +107,7 @@ export class MemberListComponent implements OnChanges, OnDestroy {
       next: state => {
         this.state = state.tag;
         if (state.isReady) this.data = state.data;
+        this.unselectAll();
       },
       error: _err => {
         this.state = 'error';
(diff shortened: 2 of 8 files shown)


ThomasJuster:
## Acceptance criteria taken from the task #334

In group composition page, for "direct children" and "all descendants", for "sub-groups", "sessions" and "teams":

- [ ] (same as in user list) let the user select one, all or several rows
- [ ] (same as in user list) add a "trash" icon + "Remove from group" bottom-right of the table, which is greyed out when there is no selection
- [ ] on click on "Remove from group", display a primeng ConfirmPopup with "?" icon and msg "Do you want to also delete the selected group(s)? (will only work if those are empty)"
  - [ ] on "no": display a primeng ConfirmPopup with "!" icon and msg "By removing "groupA", "groupB" from the group, you may loose manager access to them (if no explicit permission or through other parent group). Are you sure you want to proceed?"
  - [ ] on "yes": display a primeng ConfirmPopup with "!" icon and msg "Are you sure you want to permanently delete "groupA", "groupB"? This operation cannot be undone."
- depending on the decision, you call "removeGroup" or groupRemoveChild (ignore the orphan thing) service on each group. Grey out action and table while processing.
  - [ ] `removeGroup`
  - [ ] `groupRemoveChild`
- on success/failure/partial success, display appropriate message as done in pending request component.
  - [ ] success
  - [ ] partial success
  - [ ] failure
- [ ] If the group could not be deleted because it was not empty, display an appropriate message.
- [ ] the table needs to be refreshed after every (?) is done (same as in user list)

src/app/core/app.component.html
```diff
@@ -49,7 +49,7 @@
 </div>
```
ThomasJuster, line 49: When chosing to also remove groups that are empty with a mix of empty and populated groups in the selection, the behavior is a weird:
1. If the group is not empty, we still send an xhr request that fails, I doubt we should send it if we know the group is not empty
2. If a group cannot be deleted it won't be removed from the sub groups list. I think the expected behavior is that it should removed from the subgroups list but not deleted

The 2 points are probably correlated, shouldn't we proceed this way for the "remove groups that are empty" scenario:
:arrow_right: if group is empty, call remove group service ; else call "group remove child"
?

What do you think of the current behavior and the proposed one, do you see something else we could/should do ?

ThomasJuster (changes requested):
Big PR, lot of work :+1: 
I sticked to the functional part for now, I saw some small bugs and one weird behavior, I think we can iterate over that later

src/app/core/app.component.html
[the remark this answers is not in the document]
Iloveall: Fixed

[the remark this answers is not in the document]
Iloveall: Fixed

Iloveall: Yes, I guess so lot changes need to approve with @smadbe or maybe create new proposal issue for improve this logic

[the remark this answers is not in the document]
Iloveall: Probably, but I guess so lot changes need to approve with customer or @smadbe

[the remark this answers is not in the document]
ThomasJuster: It seemed pretty straightforward to me but alright, I created the issue

ThomasJuster: Issue created

src/app/modules/group/components/member-list/member-list.component.ts
```diff
@@ -243,4 +258,85 @@ export class MemberListComponent implements OnChanges, OnDestroy {
...
+      },
+      error: _err => {
+        this.removalInProgress$.next(false);
+        this.actionFeedbackService.unexpectedError();
+      }
+    });
+  }
+
+  onRemove(event: Event): void {
+    if (this.selection.length === 0 || !this.groupData) return;
```
ThomasJuster, line 311: An error should be thrown

src/app/modules/group/components/member-list/member-list.component.ts
```diff
@@ -243,4 +258,85 @@ export class MemberListComponent implements OnChanges, OnDestroy {
...
+      key: 'commonPopup',
+      icon: 'pi pi-exclamation-triangle',
+      message: $localize`By removing ${getSelectedGroupChildCaptions(this.selection as GroupChild[])} from the group, you may loose
+       manager access to them (if no explicit permission or through other parent group). Are you sure you want to proceed?`,
+      acceptLabel: $localize`Yes`,
+      acceptIcon: 'fa fa-check',
+      rejectLabel: $localize`No`,
+      accept: () => this.removeGroupsOrSubgroups(selectedIds, groupId),
+    });
+  }
```
ThomasJuster, line 288: To me `selectedIds` induces confusion, you have it directly from the class property `selection`, there's no need to make it a method argument
Or I am missing something :monocle_face:

src/app/modules/group/components/member-list/member-list.component.ts
```diff
@@ -243,4 +258,85 @@ export class MemberListComponent implements OnChanges, OnDestroy {
...
+        this.actionFeedbackService.unexpectedError();
+      }
+    });
+  }
+
+  onRemove(event: Event): void {
+    if (this.selection.length === 0 || !this.groupData) return;
+
+    const groupId = this.groupData.group.id;
+    const selectedIds = this.selection.map(member => member.id);
```
ThomasJuster, line 314: In the mapper function, it's not always a member, unless it is moved into the if statement below, it can also be a group child
I think the various use cases should be explicit, either we are dealing with `memberIds` or with `groupChildIds`, but `selectedIds` is dark

src/app/modules/group/http-services/remove-group.service.ts
```diff
@@ -0,0 +1,36 @@
+import { Injectable } from '@angular/core';
+import { forkJoin, Observable, of } from 'rxjs';
+import { HttpClient } from '@angular/common/http';
+import { appConfig } from '../../../shared/helpers/config';
+import { SimpleActionResponse } from '../../../shared/http-services/action-response';
+import { catchError, map } from 'rxjs/operators';
+
+export function parseResults(data: Map<string, SimpleActionResponse>): { countRequests: number, countSuccess: number, errorText?: string } {
```
ThomasJuster, line 8: `{ countRequests: number, countSuccess: number, errorText?: string }` looks exactly like the type `Result` from group-removal-response-handling.ts, shouldn't it be imported from there ?

src/app/modules/group/http-services/remove-group.service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  remove(id: string): Observable<SimpleActionResponse> {
+    return this.http.delete<SimpleActionResponse>(`${appConfig.apiUrl}/groups/${id}`);
+  }
+
+  removeBatch(ids: string[]): Observable<{ countRequests: number, countSuccess: number, additionalMessage?: string }> {
+    return forkJoin(
+      ids.map(id => this.remove(id).pipe(catchError(({ error }) => of(error))))
+    ).pipe(
+      map(data => parseResults(new Map(Object.entries(data)))),
```
ThomasJuster, line 33: Here `data` is an array right ?
That means a Map is created from an array entries and provided to `parseReults()`, which does a `Array.from(theMap.values())`

Why not passing the array directly ?

ThomasJuster (changes requested):

src/app/modules/group/components/member-list/member-list.component.ts
Iloveall: Fixed

src/app/modules/group/components/member-list/member-list.component.ts
Iloveall: Fixed

src/app/modules/group/components/member-list/member-list.component.ts
Iloveall: Probably it could be covered by my fix for prev comment, please check and let me know what you are think :)

src/app/modules/group/http-services/remove-group.service.ts
Iloveall: Fixed

src/app/modules/group/http-services/remove-group.service.ts
Iloveall: This logic partly based on "as done in pending request component.", and I guess it was done for easiest way to get just values for reduce total count. Do you want I'm improve it?

ThomasJuster: I checked it out:
The `group-users` service returns the map, not the `parseResults()` result. This way it _may_ make sense because the subscriber might use the map in another way.

Here, the service returns directly the `parseResults()` result and not the map, so there's no chance this map is going to be used anywhere in any subscriber, and it's a same in the `remove-subgroups` service

To me it _might_ make sense in the `group-users` service but not here nor in `remove-subgroup` service, I think we should change those `remove-groups` and `remove-subgroups` services only

ThomasJuster (commented):
After the last comment it's good to go :+1:

ThomasJuster (approved):

PR #680, opened 2021-09-14
Adds item deletion logic

### LINKED ISSUE #642: Item deletion
## Overview
Allow the owner of an item to delete it

## Acceptance criteria
In the same way we do for groups:
- [x] in the item edit "advanced parameter" page, on the bottom, add a "Delete this item" button which is greyed out by default.
- [x] if the user is not an owner (see permissions) of the item, grey out the button and display on the left "Only owner can delete items"
- if the user is an owner
   - [x] if the item is a chapter and it has children, display "Only empty chapters can be deleted."
   - [x] if the item is a skill and it has children, display "Only empty skills can be deleted."
   - [x] otherwise, enable the button 
   - [x] on click, display a confirmation (red "yes", no)overlay "Are you sure you want to delete this content?" "Deleting it will also remove permanently all answers and results related with this content."
   - [x] on "yes", call the [dedicated service](https://france-ioi.github.io/algorea-devdoc/api/#operation/itemDelete). 
   - [x] handle success and failure a toast message
   - [x] after a successful deleting, redirect to the root of the platform

Iloveall:
Fixes #642

### DIFF
--- a/src/app/modules/item/components/item-remove-button/item-remove-button.component.ts
+++ b/src/app/modules/item/components/item-remove-button/item-remove-button.component.ts
@@ -1,11 +1,9 @@
 import { Component, Input, OnChanges } from '@angular/core';
-import { ReplaySubject, combineLatest } from 'rxjs';
-import { distinct, map, switchMap, filter } from 'rxjs/operators';
+import { ReplaySubject } from 'rxjs';
+import { distinct, map, switchMap } from 'rxjs/operators';
 import { mapToFetchState } from '../../../../shared/operators/state';
 import { GetItemChildrenService, ItemChild } from '../../http-services/get-item-children.service';
 import { Item } from '../../http-services/get-item-by-id.service';
-import { ItemDataSource } from '../../services/item-datasource.service';
-import { isNotNullOrUndefined } from '../../../../shared/helpers/null-undefined-predicates';
 import { ConfirmationService } from 'primeng/api';
 import { RemoveItemService } from '../../http-services/remove-item.service';
 import { ActionFeedbackService } from '../../../../shared/services/action-feedback.service';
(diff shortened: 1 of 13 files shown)


src/app/modules/item/components/item-remove-button/item-remove-button.component.ts
```diff
@@ -0,0 +1,106 @@
...
+  private readonly id$ = new ReplaySubject<string>(1);
+  readonly state$ = combineLatest([
+    this.id$,
+    this.itemDataSource.state$.pipe(
+      map(state => {
+        if (state.isReady) {
+          return state.data.route.attemptId || state.data.route.parentAttemptId;
+        }
+        return null;
+      }),
```
ThomasJuster, line 31: the item datasource service should only be injected in page components, if you need the route or anything else, you can pass `itemData` as input instead of just an item

ThomasJuster (changes requested):
Sorry for the delay of the review, I had trouble finding test cases.
I have one comment, then it's all good :+1:

src/app/modules/item/components/item-remove-button/item-remove-button.component.ts
Iloveall: Fixed

src/app/modules/item/components/item-remove-button/item-remove-button.component.ts
```diff
@@ -0,0 +1,96 @@
...
+  templateUrl: './item-remove-button.component.html',
+  styleUrls: [ './item-remove-button.component.scss' ],
+})
+export class ItemRemoveButtonComponent implements OnChanges {
+  @Input() item?: Item;
+  @Input() attemptId?: string;
+
+  private readonly params$ = new ReplaySubject<{ id: string, attemptId: string }>(1);
+  readonly state$ = this.params$.pipe(
+    distinct(),
```
ThomasJuster, line 23: Does it have any effect :thinking: ?
If `params$` is a subject of an object, every value value is an object, so they are all different no ?

I wonder if it's not a `distinctUntilChanged((a, b) => a.id === b.id && a.attemptId === b.attemptId)` you're looking for.

If it makes to change, go for it :pray:

src/app/modules/item/pages/item-edit-advanced-parameters/item-edit-advanced-parameters.component.ts
```diff
@@ -58,7 +60,16 @@ export class ItemEditAdvancedParametersComponent implements OnInit {
...
+  attemptId$ = this.itemDataSource.state$.pipe(
+    map(state => {
+      if (state.isReady) {
+        return state.data.route.attemptId || state.data.route.parentAttemptId;
+      }
+      return undefined;
+    })
+  );
+
+  constructor(private itemDataSource: ItemDataSource) { }
```
ThomasJuster, line 72: OK, what I wrote was unclear, sorry about that.
What I meant is: the item datasource service should only be injected into pages (or routes) declared in the `item-routing.module`. Currently there are only 3: `ItemById`, `ItemDetails` and `ItemEdit`

TL;DR: Here it should be `ItemEdit`

That's because currently the datasource services are currently a workaround to share data between `ItemById` and its sub routes `ItemDetails` and `ItemEdit`. At the time, we didn't know there was a way to pass data between a route and its sub routes. Now that we know how to do it, Damien intends to remove those services in a near future.

ThomasJuster (changes requested):

src/app/modules/item/components/item-remove-button/item-remove-button.component.ts
Iloveall: Fixed

ThomasJuster (approved):

PR #687, opened 2021-09-16
Fixes very long title for left menu

### LINKED ISSUE #604: Left menu: handle very long title
## Overview 
Fix the display with parents have very long title.
Screenshot 2021-08-16 at 17 38 06

## Subtasks

- [x] limit to 2 lines max the title, display "..." ellipsis if longer
- [x] Don't let the icon zone on the left squeeze (no more space on the right of the icon). When the title is long: Screenshot 2021-08-16 at 17 45 00 . Should be: Screenshot 2021-08-16 at 17 44 34

Iloveall:
Fixes #604

### DIFF
--- a/src/app/core/components/left-nav-tree/left-nav-tree.component.scss
+++ b/src/app/core/components/left-nav-tree/left-nav-tree.component.scss
@@ -1,4 +1,5 @@
 @import "src/colors.scss";
+@import "src/geometry";
 
 .tree-header {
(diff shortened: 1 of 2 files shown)


ThomasJuster (approved):
Niice ! I heard about line-clamp but I didn't know how to use it exactly and that it was actually supported by a lot of browser :smile: 
I thought since it was webkit it was only supported by chromium browsers and safari…

PR #690, opened 2021-09-21
fix group membership decoder

ThomasJuster:
## Tests

Go to "my group" page:
- On the [production app](https://dev.algorea.org/en/#/groups/mine), the joined group list should be broken.
- On the [branch app](https://dev.algorea.org/branch/fix/joined-groups-list/en/#/groups/mine), it should be fixed.


Iloveall (approved):
I've just add the same fix in another PR :)

PR #699, opened 2021-09-29
Permissions dialog refactoring with small-section component

GabrielVidal1:
I tried to refact the `boolean-section` and `progress-section` components as suggested [here](https://github.com/France-ioi/AlgoreaFrontend/pull/692#issuecomment-925890649) by @ThomasJuster 

I made a `small-section` component that handles the collapsible section (with `ng-content` through [projecting](https://angular.io/guide/content-projection)). However when collapsed, a specific header is used for each component.

* For the `progress-section`, it's the value next to a beaded version of the progress: image

  The coloration of the beads is determined by the index of the selected value (which is an internal value of the component, not accessible from outside), so I don't think it's a good idea to completely encapsulate the logic in a component like `small-section`, if we want to use the `ControlValueAccessor` interface to transform it into a reusable form compatible component.

* For the `boolean-section`, the header contains a simple 'Yes' or 'No' depending on the value, so it could be doable to completely get rid of the component and simply use the `small-section` component with a `switch` inside. However this means duplicating the inside code (the switch and label) to all places where this component is used (4 for now) 

I was able to use the `small-section` component in both the `progress-section` and `boolean-section` component to factorize some of the logic (mainly the collapse behavior) but this didn't quite reduce the quantity of code. 

I suggest keeping both `boolean-section` and `progress-section` and make them implement the `ControlValueAccessor` interface in the next PR:
* `boolean-section` will be a boolean
* `progress-section` will be a union of the types of the different possible values (for the permission dialog box, a string union)

Let me know how you feel about my conclusion :slightly_smiling_face:


src/app/modules/group/components/manager-permission-dialog/manager-permission-dialog.component.html
```diff
@@ -48,11 +46,9 @@
...
           (valueChange)="onValueChange()"
       >
-        <ng-template #label>
-          <span i18n>
-            User(s) can give and revoke members access to some content
-          </span>
-        </ng-template>
+        <span label i18n>
+          User(s) can give and revoke members access to some content
+        </span>
```
ThomasJuster, line 51: boolean-section and progress-section should have the same API as before

src/app/modules/shared-components/components/progress-section/boolean-section/boolean-section.component.html
```diff
@@ -1,38 +1,29 @@
-<alg-progress-section
+<alg-small-section
```
ThomasJuster, line 1: boolean-section should be moved out of progress-section folder

src/app/modules/shared-components/components/small-section/small-section.component.html
```diff
@@ -0,0 +1,18 @@
+<div class="progress-section-container">
```
ThomasJuster, line 1: The name `small-section` does not really reflect what the component does, can you find a more precise name ?
Some ideas to narrow the search: `form-control-section`, `control-section`

src/app/modules/shared-components/components/small-section/small-section.component.html
```diff
@@ -0,0 +1,18 @@
...
+  <div class="progress-section-header" [ngClass]="{disabled: disabled, 'cursor-pointer': collapsed}" (click)="onCollapse()">
```
ThomasJuster, line 2: Beware, a lot of class names contain `progress-section` but the component is not `progress-section` anymore

ThomasJuster (changes requested):
Sorry for the delay of the review, I needed to dive in

GabrielVidal1:
I disliked the fact that values like the section title or icon had to be passed through the `progress-section` component to be used in the `small-section` component.

After some thinking, I found a way to decouple to section part and the logic part as previously suggested [here](https://github.com/France-ioi/AlgoreaFrontend/pull/692#issuecomment-925890649).

The components have been renamed:
* `small-section` is now `collapsible-section`
* `progress-section` is now `progress-form`
* `boolean-section` is now `boolean-form`

Now the `progress-form` and `boolean-form` are used inside a `ng-template` in a `collapsible-section` and the way they are displayed changes depending on the `collapsed` state of the `collapsible-section` (passed through the `ng-template`). When collapsed, the section will render the component in its header (and the component will render its 'header' form). 

This way the inputs of the different components are only relevant to them: the section behavior is handled in the section component, and the logic of the form is handled in the inside component (in the `progress-form` or `boolean-form`).

```html
<alg-collapsible-section title="..." icon="..." collapsible="..." collapsed="...">
  <ng-template #content let-collapsed>
    <alg-progress-form
      [collapsed]="collapsed"
      [(value)]="..."
      [values]="..."
    ></alg-progress-form>
  </ng-template>
</alg-collapsible-section>
```

Because the `progress-form` and `boolean-form` components are meant to be used in a `collapsible-section` component, they have been mode to the `collapsible-section/` folder. These 2 components can still implement the `ControlValueAccessor` as planned.

src/app/modules/shared-components/components/small-section/small-section.component.html
GabrielVidal1: fixed

src/app/modules/shared-components/components/small-section/small-section.component.html
GabrielVidal1: fixed, it has been renamed `collapsible-section`

src/app/modules/shared-components/components/progress-section/boolean-section/boolean-section.component.html
GabrielVidal1: some important changes have been made, see [here](https://github.com/France-ioi/AlgoreaFrontend/pull/699#issuecomment-932177628)

src/app/modules/group/components/manager-permission-dialog/manager-permission-dialog.component.html
GabrielVidal1: some important changes have been made, see [here](https://github.com/France-ioi/AlgoreaFrontend/pull/699#issuecomment-932177628)

src/app/modules/shared-components/components/collapsible-section/collapsible-section.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, ContentChild, Input, TemplateRef } from '@angular/core';
```
ThomasJuster, line 1: Very clean solution :+1:
now progress-~section~ and boolean-~section~ can be used with any collapse widget, even without any widget :+1: :+1:

src/app/modules/shared-components/shared-components.module.ts
```diff
@@ -87,13 +88,14 @@ import { ErrorComponent } from './components/error/error.component';
     SubSectionComponent,
     AddContentComponent,
     FloatingSaveComponent,
-    ProgressSectionComponent,
-    BooleanSectionComponent,
+    ProgressFormComponent,
+    BooleanFormComponent,
```
ThomasJuster, line 92: The `Form` suffix is misleading, those are form controls and not forms, I meaning they are not a group of controls which you can submit

ThomasJuster (commented):
Very neat solution :+1:

src/app/modules/shared-components/shared-components.module.ts
GabrielVidal1: Fixed, I renamed both of them:
* `progress-form` is now `progress-select`
* `boolean-section` is now `switch-select`

ThomasJuster: progress-select is OK :heavy_check_mark: 
switch-select is still weird: it's not a select, it's a switch :thinking:

GabrielVidal1: how about `switch-field`?

ThomasJuster: OK let's try that :+1:

GabrielVidal1: fixed

src/app/modules/group/components/manager-permission-dialog/manager-permission-dialog.component.html
```diff
@@ -25,50 +25,69 @@
         This panel lets you select what you allow the selected user or group of users to do on the current group and its descendants.
```
ThomasJuster, line 25: **Bug** In the group managers page, both progress and switch are collapsible, it should not be the case

**Steps to reproduce**
1. Navigate to [this group managers](https://dev.algorea.org/branch/permissions-dialog-refact/en/#/groups/by-id/672913018859223173;path=52767158366271444/details/managers)
2. For 1st manager, click on the pen icon to edit manager permission
3. Click on the header of a progress or a switch

src/app/modules/group/components/manager-permission-dialog/manager-permission-dialog.component.html
```diff
@@ -25,50 +25,69 @@
...
       </p>
```
ThomasJuster, line 26: **(small) Bug** In the permission edit dialog, when a collapsible section is uncollapsed, the cursor when hovering the header should be a pointer

**Steps to reproduce**
1. Watch [this group](https://dev.algorea.org/branch/permissions-dialog-refact/en/#/groups/by-id/4761034848060385407;path=/details)
2. Click on the chapter link that appears in the overlay
3. Navigate to Progress > Chapter view
4. Click on first score you see (ie annabelle4) and click "Access" to open permission edit dialog

ThomasJuster (changes requested):

src/app/modules/group/components/manager-permission-dialog/manager-permission-dialog.component.html
GabrielVidal1: fixed

ThomasJuster (approved):

PR #700, opened 2021-09-29
Use content-info for displaying children in the left menu

smadbe:
(Easier to review commit per commit)

Fetch item/group children (for navigation) in the item/group-by-id instead of the nav menu so that it can be used globally for navigation.  
The same behavior as before is expected for the menu.

Next steps: (different PRs)
- big cleanup of left-nav-datasource: now there is no children fetching anymore, probably 80% of the complexity may be removed
- clean up of types used in left-nav item/group datasource and left nav tree
- use of a decoder in item nav services
- use of navigation state instead of refetching where navigation is required (for prev/next/up buttons, and for item display?)


src/app/core/components/left-nav/left-nav-group-datasource.ts
```diff
@@ -1,36 +1,61 @@
...
     );
   }
 
+}
+
+function mapChild(child: GroupNavigationChild): NavTreeElement {
+  return {
+    id: child.id,
+    title: child.name,
+    hasChildren: child.type !== 'User',
```
ThomasJuster, line 49: I don't get that condition, a group might have no children, even if it's not a user right ?
I feel like I'm missing something, what does `hasChildren` reflects for a `NavTreeElement` ?

Edit:
I think I get it: it means that the node _might_ have children, right ?

src/app/core/components/left-nav/left-nav-group-datasource.ts
```diff
@@ -14,7 +14,12 @@ export class LeftNavGroupDataSource extends LeftNavDataSource<GroupInfo, NavTree
   }
 
   addDetailsToTreeElement(contentInfo: GroupInfo, treeElement: NavTreeElement): NavTreeElement {
-    return contentInfo.title ? { ...treeElement, title: contentInfo.title } : { ...treeElement };
+    let group = treeElement;
+    if (contentInfo.title) group = { ...treeElement, title: contentInfo.title };
+    if (contentInfo.navData) group = { ...treeElement,
+      children: contentInfo.navData.children.map(mapChild)
+    };
+    return group;
```
ThomasJuster, line 22: I think the local variable `group` is useless here, this is equivalent to:
```ts
if (contentInfo.navData) return { ...treeElement, children: … }
if (contentInfo.title) return { ...treeElement, title: … }
return treeElement
```

src/app/modules/group/pages/group-by-id/group-by-id.component.ts
```diff
@@ -43,8 +43,8 @@ export class GroupByIdComponent implements OnDestroy {
     this.subscriptions.push(
       this.groupDataSource.state$.pipe(
         readyData(),
-        map(({ group, route, breadcrumbs }): GroupInfo => groupInfo({
-          route,
+        map(({ group, route, breadcrumbs, navigation }): GroupInfo => groupInfo({
+          route: groupRoute(group, route.path),
```
ThomasJuster, line 47: What's the use of recreating a group route ? Last time I saw `route`, it was of type `GroupRoute` already

ThomasJuster (commented):
A bit difficult for me to review the code and functional parts. For functional parts, I sticked to navigating on the app and checking the left menu, it worked fine. As for the code, although the PR reduces complexity, it is still hard to follow to me.

I'm not sure it is mentioned in the next steps: There are still 3 calls to `/navigation?child_attempt_id` when navigating to an item

src/app/core/components/left-nav/left-nav-datasource.ts
```diff
@@ -79,35 +78,11 @@ export abstract class LeftNavDataSource<ContentT extends RoutedContentInfo, Menu
...
-              this.fetchChildrenOfElementWithId(
-                prevState.data.subNavMenuData(contentId)
-                  .withUpdatedElement(contentId, el => this.addDetailsToTreeElement(contentInfo, el)),
-                contentId
-              )
-            );
+          //       OR the content is among the displayed items at the root of the tree -> select the right one
+          //       OR the content is a child of one item at the root of the tree -> shift the tree and select it
+          if (prevState.data.hasLevel1Element(contentId) || prevState.data.hasLevel2Element(contentId)) {
+            const select = prevState.data.selectedElementId !== contentId;
```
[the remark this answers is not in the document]
smadbe: I just keep the same logic as before, the arg of the function is "select" which is to "force selection". So we do select only if it was not selected before (so it is the opposite of "isSelected"). It is a bit strange but it was out of scope of this PR to check why it was required.

src/app/shared/models/content/group-info.ts
```diff
@@ -1,9 +1,11 @@
+import { GroupNavigationData } from 'src/app/core/http-services/group-navigation.service';
 import { GroupRoute } from '../../routing/group-route';
 import { ContentInfo, RoutedContentInfo } from './content-info';
 
 export interface GroupInfo extends RoutedContentInfo {
   type: 'group',
-  route: GroupRoute
+  route: GroupRoute;
+  navData?: GroupNavigationData
```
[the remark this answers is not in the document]
smadbe: done. Issue to add the eslint added (https://github.com/France-ioi/AlgoreaFrontend/issues/703)

src/app/core/components/left-nav/left-nav-group-datasource.ts
smadbe: Out of scope of the PR as well (just a move of the previous same logic to here).
Actually, the current group navigation service does not return whether the group has children or not, so I just return if it might have children. But anyway, the template does not do anything with this info for now for group I think.

src/app/modules/group/pages/group-by-id/group-by-id.component.ts
smadbe: Was a mistake. Fixed.

src/app/core/components/left-nav/left-nav-group-datasource.ts
ThomasJuster: if it has both, only the last if populates the group

smadbe: ah of course, that was a bug. Fixed.

ThomasJuster (approved):

PR #712, opened 2021-10-01
Simplify left nav datasource

smadbe:
Goal: Now children fetching has been removed from the left-nav, simply strongly the datasource.

Out of scope:
- more global reorganization of the left nav tree types (require to use decoder version of the item nav first)
- the `$any($any(..` hack of the left nav component. May disappear when the  " left nav tree types " will be fixed

Expected outcome: the left-menu behaves as in the master branch.

### DIFF
--- a/src/app/core/components/left-nav/left-nav-datasource.ts
+++ b/src/app/core/components/left-nav/left-nav-datasource.ts
@@ -1,10 +1,10 @@
-import { concat, EMPTY, Observable, of, ReplaySubject, Subject } from 'rxjs';
+import { EMPTY, Observable, of, ReplaySubject, Subject } from 'rxjs';
 import { map, switchScan } from 'rxjs/operators';
 import { isDefined } from 'src/app/shared/helpers/null-undefined-predicates';
 import { repeatLatestWhen } from 'src/app/shared/helpers/repeatLatestWhen';
 import { fetchingState, FetchState, readyState } from 'src/app/shared/helpers/state';
 import { RoutedContentInfo } from 'src/app/shared/models/content/content-info';
-import { mapErrorToState, mapStateData } from 'src/app/shared/operators/state';
+import { mapStateData, mapToFetchState } from 'src/app/shared/operators/state';
 import { NavTreeData, NavTreeElement } from '../../models/left-nav-loading/nav-tree-data';
 
 export abstract class LeftNavDataSource<ContentT extends RoutedContentInfo, MenuT extends NavTreeElement> {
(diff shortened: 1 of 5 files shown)


src/app/core/components/left-nav/left-nav-datasource.ts
```diff
@@ -1,137 +1,62 @@
...
-
-      mergeMap(([ contentInfo, prevState ]):ObservableInput<DataSourceChange<MenuT>> => {
-
-        if (prevState.isReady) {
-          // CASE: the current content is not an item and the menu has already items displayed
-          if (!contentInfo) {
-            if (prevState.data.selectedElementId !== undefined) return of(dsDeselect<MenuT>());
-            return EMPTY; // no change
-          }
+  state$ = this.contentChanges.pipe(
```
ThomasJuster, line 15: Here there is a accumulation of nested and sibling `if`s, I think/hope there is a way to declare each scenario in a private observable, and `state$` would be a `merge(…)` of them.
It could also be achieved by calling private methods in the `switchScan` callback, the point is to reduce the cognitive load.
What do you think ?

src/app/shared/operators/state.ts
```diff
@@ -49,3 +49,12 @@ export function mapStateData<T, U>(dataMapper: (data: T) => U): OperatorFunction
     }),
   );
 }
+
+/**
+ * Rx operator which convert observable error to "error state". To be used when the state is built "manually".
+ */
+export function mapErrorToState<T>(): OperatorFunction<FetchState<T>,FetchState<T>> {
```
ThomasJuster, line 56: I didn't see any use of this operator, what has it been created for ?

src/app/core/components/left-nav/left-nav-datasource.ts
smadbe: In the current version, the content of the `switchScan` block looks quite good to me to understand the overall logic (14 lines of code, the rest are comments): you see that based on the previous state and the new "content", there is a decision to be made on what to be displayed. And the action based on these case are quite straightforward so reading them directly (details are hidden in external functions already) looks clear to me.
I don't see exactly what is your idea with your merge of observable,  but doing purer Rx here (as often) would not make the logic more readable IMO. I think it would be the opposite even. But I might be wrong :-)

src/app/shared/operators/state.ts
smadbe: It has been used and then removed in a commit later. It is still used in `mapToFetchState` and may be used for other usage later so I prefer to leave it here.

ThomasJuster (approved):

PR #718, opened 2021-10-04
Item-navigation: use decoders

smadbe:
Use decoders for item-navigation services.
The behavior of the left menu should not have changed. Same for activity suggestion box (on [Pixal](https://dev.algorea.org/en/#/groups/by-id/672913018859223173;path=52767158366271444/details) page, click on observation button)

(out-of-scope/next step: clean usage of `NavMenuItem` vs `NavTreeElement`)

### DIFF
--- a/src/app/core/http-services/item-navigation.service.ts
+++ b/src/app/core/http-services/item-navigation.service.ts
@@ -3,116 +3,14 @@ import { HttpClient, HttpParams } from '@angular/common/http';
 import { Observable, of } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { bestAttemptFromResults } from 'src/app/shared/helpers/attempts';
-import { isRouteWithSelfAttempt, FullItemRoute } from 'src/app/shared/routing/item-route';
+import { isRouteWithSelfAttempt, FullItemRoute, ItemRoute } from 'src/app/shared/routing/item-route';
 import { appConfig } from 'src/app/shared/helpers/config';
-import { isASkill, isSkill, ItemType, ItemTypeCategory } from 'src/app/shared/helpers/item-type';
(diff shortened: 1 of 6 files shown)


src/app/core/http-services/item-navigation.service.ts
```diff
@@ -149,52 +68,36 @@ const itemNavigationDataDecoder = D.struct({
...
+const rootSkillsDecoder = D.array(
+  D.struct({
+    groupId: D.string,
+    name: D.string,
+    type: D.literal('Class', 'Team', 'Club', 'Friends', 'Other', 'User', 'Session', 'Base', 'ContestParticipants'),
+    skill: itemNavigationChildDecoderBase
+  })
+);
+
+export type RootSkills = D.TypeOf<typeof rootSkillsDecoder>;
```
ThomasJuster, line 96: Suggestion:
I find it usually easier to declare the non-array type/decoder and then construct the array type or decoder where needed.

It would give 
```ts
const rootSkillDecoder = D.struct(…)
export type RootSkill = D.TypeOf<typeof rootSkillDecoder>
```
and where you need it as array, use `D.array(rootSkillDecoder)` or `RootSkill[]`

Feels cleaner to me, your call

ThomasJuster (approved):

src/app/core/http-services/item-navigation.service.ts
smadbe: I have difficulties to find an actual reason for this. In what case would it be "easier"? (especially as you can always do `MyList[number]` to get the inner type.

Ok, I'll do this, even if not very convinced :-)

ThomasJuster (approved):

PR #731, opened 2021-10-06
Move left nav datasources to independant nav tree services

smadbe:
Motivation: move toward all navigation info fetched through these nav tree services instead of item/groupData.

I recommended reviewing this commit by commit.

At the end it was not so complicated (so far) and even give a few nice improvements (in term of design/quality of code).

The expected behavior is the same as before for the menu (mainly test tab change). 
But a small "bug" has been fixed: when selecting an element and then quickly changing to another tab, the additional fetching requests were making the tab moving back to the previous tab (the one of the selected element).

### DIFF
--- a/src/app/core/components/left-nav/left-nav.component.html
+++ b/src/app/core/components/left-nav/left-nav.component.html
@@ -1,55 +1,59 @@
-<p-tabView
-  styleClass="dark tab-left-nav"
-  (onChange)="onSelectionChangedByIdx($event)"
-  [(activeIndex)]="activeTabIndex"
-  *ngIf="currentUser$|async as currentUser"
->
+<ng-container *ngIf="activeTab$ | async as activeTab;">
 
-  <p-tabPanel>
-    <ng-template pTemplate="header" tabindex="5">
-      <span class="indicator dark"></span>
-      <i class="fa fa-hand-paper"></i>
-      <span class="p-tabview-title" i18n="|Tab name">Activities</span>
-    </ng-template>
-  </p-tabPanel>
+  <p-tabView
+    styleClass="dark tab-left-nav"
+    (onChange)="onSelectionChangedByIdx($event)"
+    [activeIndex]="activeTab.index"
+    *ngIf="currentUser$|async as currentUser"
+  >
 
-  <p-tabPanel>
-    <ng-template pTemplate="header" tabindex="6">
-      <span class="indicator dark"></span>
-      <i class="fa fa-graduation-cap"></i>
-      <span class="p-tabview-title" i18n="|Tab name">Skills</span>
-    </ng-template>
-  </p-tabPanel>
+    <p-tabPanel>
+      <ng-template pTemplate="header" tabindex="5">
+        <span class="indicator dark"></span>
+        <i class="fa fa-hand-paper"></i>
+        <span class="p-tabview-title" i18n="|Tab name">Activities</span>
+      </ng-template>
+    </p-tabPanel>
 
-  <p-tabPanel *ngIf="activeTabIndex === 2 || !currentUser.tempUser">
-    <ng-template pTemplate="header" tabindex="7">
-      <span class="indicator dark"></span>
-      <i class="fa fa-users"></i>
-      <span class="p-tabview-title" i18n="|Tab name">Groups</span>
-    </ng-template>
-  </p-tabPanel>
+    <p-tabPanel>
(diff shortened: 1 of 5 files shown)


src/app/core/components/left-nav/left-nav.component.ts
```diff
@@ -22,79 +20,55 @@ const groupsTabIdx = 2;
...
+  activeTab$ = merge(
+    this.manualTabChange,
+    this.currentContent.content$.pipe(
+      distinctUntilChanged((x,y) => x?.type === y?.type && x?.route?.id === y?.route?.id), // do not re-emit several time for a same content
+      map(content => contentToTabIndex(content)),
+      filter(isDefined),
+    )
+  ).pipe(
+    startWith(0),
+    map(idx => ({ index: idx })), // using object so that Angular ngIf does not ignore the "0" index
```
ThomasJuster, line 41: Why not using `*ngrxLet` then ?

src/app/core/components/left-nav/left-nav.component.html
```diff
@@ -1,55 +1,59 @@
...
-    refreshButtonType="retry"
-    (refresh)="retryError()"
-  ></alg-error>
-
-  <alg-left-nav-tree
-    *ngIf="state.isReady"
-    [data]="state.data"
-    [elementType]="$any(['activity', 'skill', 'group'][activeTabIndex])"
-  ></alg-left-nav-tree>
+<ng-container *ngIf="activeTab$ | async as activeTab;">
```
ThomasJuster, line 1: I might have found a bug (not linked to this PR)

Steps to reproduce:

Case 1:
1. Navigate to [first skill page](https://dev.algorea.org/en/#/skills/by-id/3000;path=;attempId=0/details)
2. Click on `Breadth-First Search (BFS)`
3. Click on `Depth First Search (DFS)`

:arrow_right: In the left nav, we are redirected to the root

Case 2:
1. Navigate to [first skill page](https://dev.algorea.org/en/#/skills/by-id/3000;path=;attempId=0/details)
2. Click on `Breadth-First Search (BFS)`
3. Click on `SASA`
4. Click on `Depth First Search (DFS)`

:arrow_right: In the left nav, we are **not** redirected to the root.

Is that the expected behavior ?

src/app/core/components/left-nav/left-nav.component.ts
```diff
@@ -22,79 +20,55 @@ const groupsTabIdx = 2;
...
+  );
+
+  @Output() themeChange = this.activeTab$.pipe(map(tab => (tab.index === 2 ? 'dark' : null)), delay(0));
```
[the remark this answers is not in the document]
smadbe: Ok changed, I put the `distinctUntilChanged` on the `activeTab$` instead, was triggering a lot for nothing.

src/app/core/components/left-menu/left-menu.component.html
```diff
@@ -1,4 +1,4 @@
 <alg-left-header></alg-left-header>
 <perfect-scrollbar class="left-nav-theme scrollbar" [ngClass]="{'dark': isNavThemeDark}">
-  <alg-left-nav (themeChange)="onNavThemeChange($event)" (selectId)="onSelectId($event)"></alg-left-nav>
+  <alg-left-nav (useDarkTheme)="onNavThemeChange($event)" (selectId)="onSelectId($event)"></alg-left-nav>
```
ThomasJuster, line 3: `useXxx` for an event ? What was wrong with `themeChange` ?

smadbe: I prefer to have a boolean (`useDarkTheme`) than the `string|null` (which was in fact `'dark'|null`) from earlier. 
For the "`useXxx` for an event ", "the component does use dark theme -> yes/no" seems ok to me but I can change it if you prefer. (make proposals)

ThomasJuster: I don't mind the boolean thing, it's more about event naming convention. (Warning, opinionated content below)
For events, as I understand it we use `xx[Change | Update | …]` or `xx[Changed | Updated | …]` convention, but  `useDarkThemeChange`, `darkThemeChange` or `isDarkThemeChange` is not very satisfying from a semantic standpoint.

How about `themeChange(isDark: boolean)` ?

In the end, I think it's not really important, it's just one event :unamused:

smadbe: fixed

PR #735, opened 2021-10-07
implement task `log` platform function

ThomasJuster:
## Tests

Currently, only one task emits log, so the tests are a bit shallow :grimacing: 

1. Navigate to this task [branch](https://dev.algorea.org/branch/feat/task-log/en/#/activities/by-id/8780689846455262402;path=4702,1625159049301502151;attempId=0/details) / [prod](https://dev.algorea.org/en/#/activities/by-id/8780689846455262402;path=4702,1625159049301502151;attempId=0/details)
2. Open your developer console

:arrow_right: you should see logs
:arrow_right: you should not see any decoder error

### DIFF
--- a/src/app/modules/item/services/item-task.service.ts
+++ b/src/app/modules/item/services/item-task.service.ts
@@ -68,8 +68,9 @@ export class ItemTaskService {
       askHint: () => {
         throw new Error('unimplemented method "askHint"');
       },
-      log: () => {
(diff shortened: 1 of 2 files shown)


smadbe (approved):
And what happens if the task does not send an array but something else? (blocking task error, ignored, exception in console, ...)

PR #738, opened 2021-10-07
display task tabs as siblings of item tabs

### LINKED ISSUE #717: Item task: display task tabs as siblings of item tabs

ThomasJuster:
Fixes #717

Test link: https://dev.algorea.org/branch/feat/item-task-tabs/en/#/activities/by-id/8780689846455262402;path=4702,1625159049301502151;attempId=0/details
Test link with multiple view: https://dev.algorea.org/branch/feat/item-task-tabs/en/#/activities/by-id/9105453971757450756;path=4702,1625159049301502151;parentAttempId=0/details

### DIFF
--- a/src/app/modules/item/pages/item-details/item-details.component.html
+++ b/src/app/modules/item/pages/item-details/item-details.component.html
@@ -46,7 +46,7 @@
       </a>
     </nav>
     <div class="bg-white">
-      <alg-item-content *ngIf="contentTab.isActive" [itemData]="itemData" (tabs)="taskTabs = $event"></alg-item-content>
+      <alg-item-content *ngIf="contentTab.isActive" [itemData]="itemData" (taskTabsChange)="setTaskTabs($event)"></alg-item-content>
       <alg-item-progress
         *ngIf="progressTab?.isActive || !!(watchedGroup$ | async) && !contentTab.isActive"
         [itemData]="itemData"
--- a/src/app/modules/item/pages/item-details/item-details.component.html
+++ b/src/app/modules/item/pages/item-details/item-details.component.html
@@ -18,22 +18,35 @@
         [routerLink]="'./'"
         routerLinkActive #contentTab="routerLinkActive"
         [routerLinkActiveOptions]="{exact: true}"
-        [class.active]="contentTab.isActive" i18n>
+        [class.active]="contentTab.isActive" i18n
+        [hidden]="contentTab.isActive && taskTabs.length > 0"
+      >
         Content
       </a>
+      <ng-container *ngIf="contentTab.isActive && taskTabs.length > 0">
+        <a
+          *ngFor="let tab of taskTabs"
+          class="nav-tab-item cursor-pointer"
+          [class.active]="contentTab.isActive && tab.active"
+          (click)="tab.setActive()"
+        >
+          {{ tab.name }}
+        </a>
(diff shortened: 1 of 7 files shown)


src/app/modules/item/pages/item-details/item-details.component.html
```diff
@@ -18,22 +18,35 @@
...
+        [class.active]="progressTab?.isActive"
+      >
         <i class="fa fa-chart-line"></i>
         &nbsp;
         <span i18n>Progress</span>
       </a>
     </nav>
     <div class="bg-white">
-      <alg-item-content *ngIf="contentTab.isActive" [itemData]="itemData"></alg-item-content>
+      <alg-item-content *ngIf="contentTab.isActive" [itemData]="itemData" (tabs)="taskTabs = $event"></alg-item-content>
```
smadbe, line 49: * I prefer using a intermediary "callback" function instead of changing variables directly in the template (may be harder to debug, to follow and (not here) we may lose type checking)
* Outputs generally have, as name, the event which happened, considering the value as the argument of this change. "tabs" is a bit short/unclear for an event. Suggestion: `taskTabsChange`

ThomasJuster:
> It is a bit confusing to have the tabs ressetted when clicking on "progress". Couldn't you keep them ? And grey them out (all but the first) ... ok I can ask what is the expected behavior...

Fixed. What I did is:
- When multiple views are available, hide content tabs and display task views in tabs instead. (Progress tab remains)
- When one (or less) view is available, leave tabs as is.

src/app/modules/item/pages/item-details/item-details.component.html
```diff
@@ -18,15 +18,17 @@
...
-        [class.active]="contentTab.isActive" i18n
-        [hidden]="contentTab.isActive && taskTabs.length > 0"
+        [class.active]="contentTab.isActive"
+        [hidden]="taskTabs.length > 0"
+        i18n
       >
         Content
       </a>
-      <ng-container *ngIf="contentTab.isActive && taskTabs.length > 0">
+      <ng-container *ngIf="taskTabs.length > 0">
```
smadbe: condition looks useless, if `taskTabs.length===0`, all should go well too

src/app/modules/item/pages/item-details/item-details.component.ts
```diff
@@ -21,19 +21,20 @@ export class ItemDetailsComponent implements OnDestroy {
...
   taskTabs: TaskTab[] = [];
+  defaultTaskView?: TaskTab['view'];
 
   fullFrameContent$ = this.layoutService.fullFrameContent$;
   readonly watchedGroup$ = this.userService.watchedGroup$;
 
-  private subscription = this.router.events.subscribe(event => {
-    if (event instanceof NavigationEnd) this.taskTabs = [];
+  private subscription = this.activatedRoute.params.subscribe(() => {
+    this.taskTabs = []; // reset task tabs when item changes.
```
smadbe, line 30: shouldn't you follow `this.itemDataSource.state$` instead? It looks like you following input changes in two different ways in the same component.

src/app/modules/item/pages/item-details/item-details.component.html
ThomasJuster: Indeed, I removed it a bit after your comment :smile:

PR #745, opened 2021-10-10
Feature/convert all service call to decoder

### LINKED ISSUE #638: Convert all service call to decoder
## Overview
Use decoders in all services which call the backend.

## Acceptance criteria
- [ ] All calls to the backend not yet converted to decoder have been converted to decoder.
- [ ] All types are checked carefully against the API specs (especially optional and nullable!)
- [ ] When needed, for repeating types or subtuples, re-usable specific decoder are created.

Iloveall:
Fixes #638


src/app/core/http-services/managed-groups.service.ts
```diff
@@ -1,29 +1,29 @@
 import { Injectable } from '@angular/core';
```
ThomasJuster, line 1: I think there's a bug [here (group composition user list)](https://dev.algorea.org/branch/feature/convert-all-service-call-to-decoder/en/#/groups/by-id/672913018859223173;path=52767158366271444/details/members)

Iloveall: Also need to add Session type to doc https://france-ioi.github.io/algorea-devdoc/api/#operation/invitationsView

ThomasJuster: > Please, create BE issue for doc https://france-ioi.github.io/algorea-devdoc/api/#operation/groupUserRequestsView
Currently API returns grade as nullable

I did not understand this comment, there's no `grade` property in the response schema, is it `group` ?
And it appears required but is actually nullable ?

src/app/core/http-services/managed-groups.service.ts
```diff
@@ -1,29 +1,29 @@
...
-  can_grant_group_access: boolean,
-}
+import { pipe } from 'fp-ts/function';
+import * as D from 'io-ts/Decoder';
+import { decodeSnakeCase } from '../../shared/operators/decode';
+
+const typeDecoder = D.literal('Class', 'Team', 'Club', 'Friends', 'Other', 'Session', 'Base');
+export const manageTypeDecoder = D.literal('none', 'memberships', 'memberships_and_group');
+
+const groupDecoder = pipe(
```
ThomasJuster, line 12: The `pipe` seems useless

src/app/modules/group/http-services/get-group-members.service.ts
```diff
@@ -2,34 +2,35 @@ import { HttpClient, HttpParams } from '@angular/common/http';
...
-  action: 'invitation_accepted'|'join_request_accepted'|'joined_by_code'|'added_directly',
-  user: null|{
-    login: string,
-    firstName: string|null,
-    lastName: string|null,
-    grade: number|null,
-    groupId: string,
-  },
-}
+const memberDecoder = pipe(
```
ThomasJuster, line 24: The `pipe` looks useless here as well

src/app/modules/group/http-services/get-requests.service.ts
```diff
@@ -3,10 +3,60 @@ import { HttpClient, HttpParams } from '@angular/common/http';
...
+  D.intersect(
+    D.partial({
+      firstName: D.nullable(D.string),
+      lastName: D.nullable(D.string),
+      grade: D.nullable(D.number),
+    }),
+  ),
+);
+
+const groupPendingRequestDecoder = pipe(
```
ThomasJuster, line 25: `pipe` looks useless here as well, same below

src/app/modules/group/http-services/get-requests.service.ts
```diff
@@ -3,10 +3,60 @@ import { HttpClient, HttpParams } from '@angular/common/http';
...
+    invitingUser: D.nullable(D.struct({
+      id: D.string,
+      firstName: D.string,
+      lastName: D.string,
+      login: D.string,
+    })),
+  })
+);
 
 export interface PendingRequest {
```
ThomasJuster, line 57: I think some of the decoders can be reused as types in the interfaces below, there is a lot of type-related code duplicate here

src/app/modules/item/http-services/get-breadcrumb.service.ts
```diff
@@ -6,18 +6,27 @@ import { isRouteWithSelfAttempt, FullItemRoute } from 'src/app/shared/routing/it
...
+  D.struct({
+    itemId: D.string,
+    languageTag: D.string,
+    title: D.string,
+    type: D.literal('Chapter', 'Task', 'Course', 'Skill'),
+  }),
+  D.intersect(
+    D.partial({
+      attemptId: D.string,
+      attemptNumber: D.string,
```
ThomasJuster, line 26: Can you restore the comments :pray: ?

src/app/modules/shared-components/components/access-code-view/access-code-view.component.ts
```diff
@@ -31,11 +31,15 @@ export class AccessCodeViewComponent {
     this.joinByCodeService.checkCodeValidity(this.code).subscribe(response => {
       this.state = 'ready';
 
-      if (!response.valid) {
+      if (!response.valid && response.reason) {
```
ThomasJuster, line 34: Why is this added condition `response.reason` needed ?

src/app/shared/http-services/get-group-progress.service.ts
```diff
@@ -1,50 +1,40 @@
...
-interface RawGroupProgress {
-  group_id: string,
-  item_id: string,
-  average_score: number,
-  avg_hints_requested: number,
-  avg_submissions: number,
-  avg_time_spent: number,
-  validation_rate: number,
-}
+const groupProgressDecoder = pipe(
```
ThomasJuster, line 10: `pipe` looks useless, same below

src/app/shared/http-services/group-permissions.service.ts
```diff
@@ -4,24 +4,35 @@ import { Observable } from 'rxjs';
...
 
-export interface Permissions {
-  can_view: 'none'|'info'|'content'|'content_with_descendants'|'solution',
-  can_grant_view: 'none'|'enter'|'content'|'content_with_descendants'|'solution'|'solution_with_grant',
-  can_watch: 'none'|'result'|'answer'|'answer_with_grant',
-  can_edit: 'none'|'children'|'all'|'all_with_grant',
-  can_make_session_official: boolean,
-  is_owner: boolean,
-}
+const permissionsDecoder = pipe(
```
ThomasJuster, line 11: `pipe` seems useless, same below

src/app/core/http-services/managed-groups.service.ts
Iloveall: In doc - user.grade described as optional, type number. But actual API returned null value. So I can assume or API returns wrong value or doc should be correct like user.grade could be Nullable.

src/app/modules/shared-components/components/access-code-view/access-code-view.component.ts
Iloveall: By doc reason is optional https://france-ioi.github.io/algorea-devdoc/api/#operation/groupsCodeCheck

src/app/modules/group/http-services/get-requests.service.ts
Iloveall: Actually I tried, but realised it requires a lot of changes including group-manager-list.component.spec.ts 

Maybe it could be done in separated issue?

src/app/modules/group/http-services/get-requests.service.ts
Iloveall: Fixed

src/app/modules/group/http-services/get-requests.service.ts
ThomasJuster: I think it's fine as is, I think it won't be worth the effort. Let's leave it like this.

ThomasJuster (commented):
One last small comment and we're good to go :+1:

src/app/modules/shared-components/components/access-code-view/access-code-view.component.ts
ThomasJuster: I think it should be handled at the same place, that way we can clearly see at first glance that no matter what, when the response is invalid, we display an error message, ie:
```ts
if (!response.valid) {
  this.actionFeedbackService.error(response.reason ? … : fallback value)
}
```

For the error message itself, since it's displayed to the end user, it should not contain tech-related information.
Here, we check the code validity, so if the response is invalid, I guess it means that the code is invalid. The message could be something like `"The provided code is invalid"` ?

Iloveall: Fixed

PR #748, opened 2021-10-11
Reorganisation of nav tree fetching

### LINKED ISSUE #726: Left menu bug when the same item appears several time
If the the same item appears several time in the left menu (for instance on `/skills/by-id/3002;path=3000;attempId=0/details`), "DFS" is both a sibling and a child of "BFS", the menu always select the sibling on navigation, even it was the child which was selected. In such a case, the left nav does not correspond to the page path.

smadbe:
Reorganisation (improvements?) of code related to nav tree fetching.

* Should fix #726
* Should fix: because of the switchMap reception of additional info (attempt, ...) was cancelling the previous request and so some requests were possibly repeated because of that.
* Should make the core computation of nav tree (the content of the mergeScan) a bit cleaner

### DIFF
--- a/src/app/core/models/left-nav-loading/nav-tree-data.ts
+++ b/src/app/core/models/left-nav-loading/nav-tree-data.ts
@@ -1,4 +1,5 @@
 import { ensureDefined } from 'src/app/shared/helpers/null-undefined-predicates';
+import { ContentRoute, pathEquals } from 'src/app/shared/routing/content-route';
 
 type Id = string;
 export enum GroupManagership { False = 'false', True = 'true', Descendant = 'descendant' }
@@ -48,8 +49,9 @@ export class NavTreeData {
    * Return this with the element from `elements` (identified by its id) replaced by the given one
    * If the element is not found, return this unchanged.
    */
-  withUpdatedElement(id: Id, update: (el:NavTreeElement)=>NavTreeElement): NavTreeData {
-    const idx = this.elements.findIndex(i => i.id === id);
+  withUpdatedElement(route: ContentRoute, update: (el:NavTreeElement)=>NavTreeElement): NavTreeData {
+    if (!pathEquals(route.path, this.pathToElements)) return this; // the element in not in tree as their paths do not match
+    const idx = this.elements.findIndex(i => i.id === route.id);
     if (idx === -1) return this;
     const elements = [ ...this.elements ];
     elements[idx] = update(ensureDefined(elements[idx]));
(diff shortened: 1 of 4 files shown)


src/app/shared/routing/content-route.ts
```diff
@@ -21,3 +21,8 @@ export function pathAsParameter(value: string[]): UrlCommandParameters {
   params[pathParamName] = value;
   return params;
 }
+
+export function pathEquals(path1: Id[], path2: Id[]): boolean {
+  if (path1.length !== path2.length) return false;
+  return path1.every((id, idx) => id === path2[idx]);
+}
```
ThomasJuster, line 28: It looks like a regular `arrayEquals`, it does not look product-related :thinking: 
Shouldn't it be an array helper or something ?

src/app/core/models/left-nav-loading/nav-tree-data.ts
```diff
@@ -1,4 +1,5 @@
 import { ensureDefined } from 'src/app/shared/helpers/null-undefined-predicates';
```
ThomasJuster, line 1: I think I found a new bug not related to this PR, but probably to the whole left-nav improvement thingy:
1. Go to [My groups](https://dev.algorea.org/en/#/groups/mine) directly (by pasting url, not by navigating inside the app)

:arrow_right: The left nav displays activities. Shouldn't it display `My groups` in the `Groups` tab ?

src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -12,49 +12,46 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo, Childre
...
-        const contentInMenuLev1 = prevData.hasLevel1Element(contentId);
-        const contentInMenuLev2 = prevData.hasLevel2Element(contentId);
-        if (contentInMenuLev1 || contentInMenuLev2) {
-          let data = contentInMenuLev1 ? prevData.withSelection(contentId) : prevData.subNavMenuData(contentId);
-          data = data.withUpdatedElement(contentId, el => this.addDetailsToTreeElement(el, contentInfo, children));
+  private children$ = this.content$.pipe(this.childrenNavigation());
+  state$ = combineLatest([ this.content$, this.children$ ]).pipe(
+    mergeScan((prevState: FetchState<NavTreeData>, [ content, children ]) => {
+
+      // CASE 1: the content is not of type of this menu
```
ThomasJuster, line 26: What does it mean `is not of type of this menu` ? (Same for `CASE 2`)

src/app/core/models/left-nav-loading/nav-tree-data.ts
smadbe: Yes, I saw this yesterday as well. Unrelated with this PR, but we might want to fix it.

smadbe: Issue created. #749

src/app/core/services/navigation/nav-tree.service.ts
smadbe: nav menu can be for activities, skills or groups. How do you want me to say that without being too long?

src/app/shared/routing/content-route.ts
smadbe: fixed

PR #755, opened 2021-10-13
refresh score after saving a grade

### LINKED ISSUE #714: Item: refresh score in left nav, top bar and title
## Description

After submitting an answer, the score is not refreshed in the left nav, in the top bar and in the title.

## Steps to reproduce

1. As an anonymous user (use a private window), navigate to [this item](https://dev.algorea.org/en/#/activities/by-id/1646068694606368976;path=4702,1352246428241737349,1070539248445908636,61555878157023179;attempId=0/details)
2. Solve the first exercise
3. Validate your answer

:arrow_right: Acutal: The score remains `0` in the header and in the left nav bar
:arrow_right: Expected: The score should be refreshed

ThomasJuster:
## Description

Fixes #714

## Test cases

- [ ] Case 1
  1. Given I am anonymous (it also works for usual user but steps are harder to reproduce)
  2. When I navigate to [this item](https://dev.algorea.org/branch/feat/item-refresh-score/en/#/activities/by-id/784078458149384669;path=4702,1352246428241737349,314613032161178344,34689573454313988;parentAttempId=0/details)
  3. And I complete first level
  4. The score should be refreshed in 3 places:
    - In the item title (:bulb: scroll to top to see it)
    - In the item header (:bulb: scroll down a bit to see it appear at the top of the screen)
    - In the left menu

### DIFF
--- a/src/app/modules/item/components/item-header/item-header.component.ts
+++ b/src/app/modules/item/components/item-header/item-header.component.ts
@@ -1,6 +1,9 @@
 import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
+import { filter } from 'rxjs/operators';
 import { ActivityNavTreeService, SkillNavTreeService } from 'src/app/core/services/navigation/item-nav-tree.service';
 import { isASkill } from 'src/app/shared/helpers/item-type';
+import { isItemInfo } from 'src/app/shared/models/content/item-info';
+import { CurrentContentService } from 'src/app/shared/services/current-content.service';
 import { ModeAction, ModeService } from 'src/app/shared/services/mode.service';
 import { ItemData } from '../../services/item-datasource.service';
 
@@ -16,10 +19,13 @@ export class ItemHeaderComponent implements OnChanges {
   private skillNavigationNeighbors$ = this.skillNavTreeService.navigationNeighbors$;
   navigationNeighbors$ = this.activityNavigationNeighbors$;
 
+  content$ = this.currentContentService.content$.pipe(filter(isItemInfo));
+
   constructor(
     private modeService: ModeService,
     private activitiyNavTreeService: ActivityNavTreeService,
     private skillNavTreeService: SkillNavTreeService,
+    private currentContentService: CurrentContentService,
   ) {}
 
   ngOnChanges(_changes: SimpleChanges): void {
--- a/src/app/modules/item/components/item-header/item-header.component.ts
+++ b/src/app/modules/item/components/item-header/item-header.component.ts
@@ -1,9 +1,7 @@
 import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
-import { filter } from 'rxjs/operators';
 import { ActivityNavTreeService, SkillNavTreeService } from 'src/app/core/services/navigation/item-nav-tree.service';
 import { isASkill } from 'src/app/shared/helpers/item-type';
-import { isItemInfo } from 'src/app/shared/models/content/item-info';
-import { CurrentContentService } from 'src/app/shared/services/current-content.service';
+import { ItemInfo } from 'src/app/shared/models/content/item-info';
 import { ModeAction, ModeService } from 'src/app/shared/services/mode.service';
 import { ItemData } from '../../services/item-datasource.service';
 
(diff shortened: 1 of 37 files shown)


src/app/modules/item/components/item-header/item-header.component.ts
```diff
@@ -16,10 +19,13 @@ export class ItemHeaderComponent implements OnChanges {
   private skillNavigationNeighbors$ = this.skillNavTreeService.navigationNeighbors$;
   navigationNeighbors$ = this.activityNavigationNeighbors$;
 
+  content$ = this.currentContentService.content$.pipe(filter(isItemInfo));
```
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/modules/item/pages/item-details/item-details.component.ts
```diff
@@ -23,8 +26,9 @@ export class ItemDetailsComponent implements OnDestroy {
   taskTabs: TaskTab[] = [];
   taskView?: TaskTab['view'];
 
-  fullFrameContent$ = this.layoutService.fullFrameContent$;
+  readonly fullFrameContent$ = this.layoutService.fullFrameContent$;
   readonly watchedGroup$ = this.userService.watchedGroup$;
+  readonly content$ = this.currentContentService.content$.pipe(filter(isItemInfo));
```
smadbe, line 31: But that's the same here. Items push info into the current content, components on the right side (excluding the top bar) of the screen are part of the content so should not use current content.

src/app/modules/item/services/item-task-answer.service.ts
```diff
@@ -102,18 +105,46 @@ export class ItemTaskAnswerService implements OnDestroy {
...
 
   clearAnswer(): Observable<unknown> {
     return this.task$.pipe(take(1), switchMap(task => task.reloadAnswer('')));
   }
+
+  private updateCurrentContentScore(currentScore: number): void {
+    const content = this.currentContent.current();
+    if (!content) throw new Error('unexpected: no content to update');
+    if (!isItemInfo(content)) throw new Error('unexpected: content is not item info');
+
```
smadbe, line 132: Mmh no.
For the same reason which was described above.
Ok this is conceptual. But in practice, that two different parts push on the same channel at the same time... item-by-id needs to know about the score! If item-by-id repush something for some reason (we want to re-add navigation in item-by-id for instance ;-)), it shouldn't break this score update for instance.

smadbe (changes requested):
Yesterday I stopped after the first remark but it was just the tip hiding the iceberg. I don't think that's the right way to do. The info should climb back to the item-by-id which is the only item-* component in charge of updating the current-content.
(we can 📞 if you want)

src/app/modules/item/services/item-datasource.service.ts
```diff
@@ -38,6 +38,7 @@ export class ItemDataSource implements OnDestroy {
     // maxScorePatch is resetted to undefined at each refresh OR item change (see subscriptions below)
     scan<number | undefined, number | undefined>((max, score) => (score ? Math.max(score, max ?? 0) : undefined), undefined),
```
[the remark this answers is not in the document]
ThomasJuster: fixed

src/app/modules/item/services/item-datasource.service.ts
```diff
@@ -34,8 +34,8 @@ export class ItemDataSource implements OnDestroy {
   private readonly refresh$ = new Subject<void>();
   private readonly scorePatch$ = new Subject<number | undefined>();
   private readonly maxScorePatch$ = this.scorePatch$.pipe(
-    // Keep max score of all emitted scores **for current item**
-    // maxScorePatch is resetted to undefined at each refresh (see subscriptions below)
+    // Keep max score of all emitted scores. NB: adding operators to a subject makes it COLD.
+    // Since it is cold, mean of max scores is ONLY computed for values emitted during the lifetime of the subscription
```
smadbe, line 38: why "mean"?

smadbe (approved):
Looks good (I haven't rechecked everything).

src/app/modules/item/services/item-datasource.service.ts
ThomasJuster: :thinking: indeed, it's max. I'll change it

PR #763, opened 2021-10-15
Let neighbor navigation use the nav tree service to get the info

smadbe:
Let the neighbor navigation use the new nav tree service to get the info.

2 cases to be tested:
- parent-previous-next navigation on item pages (activities and skills): free play as testing
- "go to next task" in item task: **untested**. @ThomasJuster  could you test that properly?

Checked: there is no useless calls to the navigation services.

### DIFF
--- a/src/app/core/http-services/item-navigation.service.ts
+++ b/src/app/core/http-services/item-navigation.service.ts
@@ -1,11 +1,10 @@
 import { Injectable } from '@angular/core';
 import { HttpClient, HttpParams } from '@angular/common/http';
-import { Observable, of } from 'rxjs';
+import { Observable } from 'rxjs';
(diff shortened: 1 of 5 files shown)


ThomasJuster:
FYI: I tested "Go to next task" [here](https://dev.algorea.org/branch/neighbor-nav-using-nav-tree-srv/en/#/activities/by-id/784078458149384669;path=4702,1352246428241737349,314613032161178344,34689573454313988;parentAttempId=0/details) as anonymous user, went to 4 stars exercise, and repeated 360 times "move forward 1 step" and "turn left 1°" :smile: 
A dialog will appear saying "Go to next task ?", click "yes"

It works :heavy_check_mark:

ThomasJuster (approved):
:+1:

PR #764, opened 2021-10-15
Fix unit test unforce

### LINKED ISSUE #711: Remove the force on url test
Currently, unit test are not run as there is a "force" on two specific tests which makes the tests run only on these.
(src/app/shared/helpers/url.spec.ts)

smadbe:
Two tests were forced (my bad), so only these tests were run when launching tests.

So:
- remove the force
- add an eslint rule to prevent committing such force-tests
- fix other tests which were broken but not visible as not run

Fixes #711


.eslintrc.js
```diff
@@ -7,9 +7,9 @@ module.exports = {
     'eslint:recommended',
     'plugin:@typescript-eslint/eslint-recommended',
     'plugin:@typescript-eslint/recommended-requiring-type-checking',
-    'plugin:@angular-eslint/recommended'
+    'plugin:@angular-eslint/recommended',
   ],
-  plugins: ['rxjs'],
+  plugins: ['rxjs', 'jest'],
```
ThomasJuster, line 12: Is it normal to add a Jest ESLint plugin when we actually don't use Jest … ?

ThomasJuster: I did some digging, I think we are using jasmine, there's an ESLint jasmine plugin with a rule that looks to do the trick:
https://github.com/tlvince/eslint-plugin-jasmine/blob/master/docs/rules/no-focused-tests.md

PR #765, opened 2021-10-15
fix editing while observing

### LINKED ISSUE #554: Bug: switching from observation to editing mode preserve observation without showing it
Scenario:
- start watching a group
- go on an item page
- enter into editing mode
- leave editing mode
- the ui shows no mode while the pages consider that are still watching the initial group

smadbe:
Fixes #554 

Scenario:
- go to group 4LTDI1819/Pixal
- start observing
- click directly to visit "TestTask"
- in the progress tab, I see "Situation of Pixal"
- press the "start editing" icon
- press the "stop editing" icon
- in progress tab, I see "Your situation" (was "Situation of Pixal" on master)

Non-goals: the way to manage edit/observation modes is a bit strange and weird. The goal on this PR is just to fix the bug, not to improve the code. In the same way, probably there may be a better way to ensure that the observation mode is `off` than stopping it even if not already `on`, but that works and doing better would have require more code rewriting.

### DIFF
--- a/src/app/modules/group/pages/group-edit/group-edit.component.ts
+++ b/src/app/modules/group/pages/group-edit/group-edit.component.ts
@@ -1,7 +1,7 @@
 import { Component, OnDestroy } from '@angular/core';
 import { FormBuilder, Validators } from '@angular/forms';
 import { mapStateData, readyData } from 'src/app/shared/operators/state';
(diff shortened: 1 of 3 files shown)


ThomasJuster (approved):
:+1: 
Thanks for the testing details

PR #767, opened 2021-10-16
Feature/manager page handle partial error

### LINKED ISSUE #740: Manager page: handle partial error
## Overview
Handle partial success in  the manager page.

## Acceptance criteria
- [ ] As you delete in parallel, some may fail and some may succeed. You need an intermediate (yellow) message (when not all have succeeded and not all have failed) such as "An error has occurred while removing some managers, check the new manager list".

Iloveall:
Fixes #740

I've kept the same structure of messages as for members

### DIFF
--- a/src/app/modules/group/components/group-manager-list/group-manager-list.component.ts
+++ b/src/app/modules/group/components/group-manager-list/group-manager-list.component.ts
@@ -3,11 +3,12 @@ import { GetGroupManagersService, Manager } from '../../http-services/get-group-
 import { GroupData, GroupDataSource } from '../../services/group-datasource.service';
 import { RemoveGroupManagerService } from '../../http-services/remove-group-manager.service';
 import { ActionFeedbackService } from '../../../../shared/services/action-feedback.service';
-import { concat, ReplaySubject, Subject } from 'rxjs';
+import { ReplaySubject, Subject } from 'rxjs';
 import { switchMap, map, distinctUntilChanged } from 'rxjs/operators';
 import { mapToFetchState } from '../../../../shared/operators/state';
 import { UserSessionService } from '../../../../shared/services/user-session.service';
 import { ConfirmationService } from 'primeng/api';
+import { displayGroupManagerRemovalResponseToast } from './group-manager-removal-response-handling';
 
 @Component({
   selector: 'alg-group-manager-list',
@@ -40,7 +41,6 @@ export class GroupManagerListComponent implements OnChanges, OnDestroy {
     private getGroupManagersService: GetGroupManagersService,
     private removeGroupManagerService: RemoveGroupManagerService,
     private actionFeedbackService: ActionFeedbackService,
-    private feedbackService: ActionFeedbackService,
     private groupDataSource: GroupDataSource,
     private userService: UserSessionService,
     private confirmationService: ConfirmationService,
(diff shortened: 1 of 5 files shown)


src/app/modules/group/components/group-manager-list/group-manager-list.component.ts
```diff
@@ -3,11 +3,12 @@ import { GetGroupManagersService, Manager } from '../../http-services/get-group-
 import { GroupData } from '../../services/group-datasource.service';
```
ThomasJuster, line 3: When all requests fail, the list is reloaded, I think it should not be the case

ThomasJuster, line 3: I tried removing 3 managers with 1 succeeding request and 2 failing requests. I get the message `"1 manager(s) have been removed, 1 could not be removed."`

Iloveall: Probably because requests was sequential. Fixed

src/app/modules/group/http-services/remove-group-manager.service.ts
```diff
@@ -14,4 +25,22 @@ export class RemoveGroupManagerService {
   remove(groupId: string, managerId: string): Observable<SimpleActionResponse> {
     return this.http.delete<SimpleActionResponse>(`${appConfig.apiUrl}/groups/${groupId}/managers/${managerId}`);
   }
+
+  removeBatch(parentGroupId: string, ids: string[], ownManagerId?: string): Observable<Result> {
+    return merge(...ids.map(id => this.remove(parentGroupId, id))).pipe(
+      catchError(() => of(failedState)),
```
ThomasJuster, line 31: Here, when a request fails, all the pending ones are cancelled because an error is thrown at the global (merged) observable level. That error causes the whole observable to stop, which cancels treatment for other pending requests

I think the catchError should be applied to each "remove" observable (to be tested):
```ts
return merge(...ids.map(id => this.remove(parentGroupId, id).pipe(catchError(…)).pipe(
  reduce(…)
)
```
PS: I know I told you recently to keep observables as flat as possible, but sometimes it's not possible :confused:

src/app/modules/group/components/group-manager-list/group-manager-list.component.ts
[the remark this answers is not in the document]
ThomasJuster: I tried removing 4 managers including self with 2 fails. Having 2 fails induces here a 3rd: we don't remove self.
In this particular case, the partial message shows information about 3 managers but I tried removing 4 of them.

I think self should also be among failed requests count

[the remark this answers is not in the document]
Iloveall: Fixed

src/app/modules/group/http-services/remove-group-manager.service.ts
```diff
@@ -2,43 +2,49 @@ import { Injectable } from '@angular/core';
 import { merge, Observable, of } from 'rxjs';
 import { HttpClient } from '@angular/common/http';
 import { appConfig } from '../../../shared/helpers/config';
-import { SimpleActionResponse } from '../../../shared/http-services/action-response';
+import { assertSuccess, SimpleActionResponse } from '../../../shared/http-services/action-response';
 import { reduce, map, switchMap, catchError } from 'rxjs/operators';
-import { Result } from '../components/member-list/group-removal-response-handling';
+import { Result } from '../components/group-manager-list/group-manager-removal-response-handling';
```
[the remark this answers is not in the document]
Iloveall: Fixed

smadbe (approved):

ThomasJuster (approved):

PR #799, opened 2021-10-28
populate task random seed with attempt id

ThomasJuster:
## Description

Fixes #...

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this task](https://dev.algorea.org/branch/feat/task-random-seed/en/#/activities/by-id/1655405614123744253;path=4702,1352246428241737349,1070539248445908636,35415640138424232;parentAttempId=0/details)
  3. The task should be displayed normally
- [ ] Case 2:
  2. When I go to [this task](https://dev.algorea.org/branch/feat/task-random-seed/en/#/activities/by-id/1655405614123744253;path=4702,1352246428241737349,1070539248445908636,35415640138424232;parentAttempId=0/details)
  3. The task should be displayed normally

### DIFF
--- a/src/app/modules/item/services/item-task.service.ts
+++ b/src/app/modules/item/services/item-task.service.ts
@@ -70,7 +70,9 @@ export class ItemTaskService {
   }
 
   private bindPlatform(task: Task): void {
-    if (!this.attemptId) throw new Error('attemptId must be defined');
+    if (!this.attemptId) throw new Error('attemptId must be defined. The "configure" method has probably not been called as expected');
+    // attempt id can be used as a seed as these are currently assigned incrementally by participant id
+    // If this changes, this needs to be adapted.
     const randomSeed = Number(this.attemptId);
     if (Number.isNaN(randomSeed)) throw new Error('random seed must be a number');
 
--- a/src/app/modules/item/services/item-task.service.ts
+++ b/src/app/modules/item/services/item-task.service.ts
@@ -43,6 +43,7 @@ export class ItemTaskService {
   );
 
   private readOnly = false;
+  private attemptId?: string;
 
   constructor(
     private initService: ItemTaskInitService,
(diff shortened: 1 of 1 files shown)


src/app/modules/item/services/item-task.service.ts
```diff
@@ -68,12 +70,16 @@ export class ItemTaskService {
   }
 
   private bindPlatform(task: Task): void {
+    if (!this.attemptId) throw new Error('attemptId must be defined');
+    const randomSeed = Number(this.attemptId);
```
smadbe, line 74: The reason ids are string is that they are big int, so they are up to 2^63-1 while JS numbers loose the precision when >2^52. So mapping an id from the backend to JS may end up with many collisions in the high numbers, which is pretty annoying for a random seed (doing +1 would give the same seed).
But at the same time, currently attempt ids are incremental (they are the ones) and in such a case, it becomes a problem only we have >2^52 attempts... which is pretty safe probably.
Possibly, it is worth adding a comment such as `// attempt id can be used as a seed as these are currently assigned incrementally by participant id. If this changes, this needs to be adapted. `

src/app/modules/item/services/item-task.service.ts
```diff
@@ -68,12 +70,16 @@ export class ItemTaskService {
   }
 
   private bindPlatform(task: Task): void {
+    if (!this.attemptId) throw new Error('attemptId must be defined');
```
smadbe, line 73: For easier debugging, add  "... The "configure" has probably not been called as expected."

smadbe (commented):
👍  Good.
It makes me realize you pass both the attempt id and the route between service/component, which is not great, it is either redundant or possibly inconsistent. It would preferable to pass a `ItemRouteWithSelfAttempt` (renamed to a better name if needed). Do you want me to fix it in the PR, in a separate one?

ThomasJuster:
> Do you want me to fix it in the PR, in a separate one?

A separate one :pray:, this one is supposed to be trivial

src/app/modules/item/services/item-task.service.ts
ThomasJuster: fixed

smadbe (approved):

PR #809, opened 2021-11-09
save task state & answer at regular interval

### LINKED ISSUE #776: item task: save task state+answer at regular interval
## Motivation

The end user should be able to resume the task wherever s⋅he left off. For that purpose, we need to auto save the task at right times.

## Acceptance criteria

- [x] Checks every 5 seconds whether the task state/answer has changed since the latest successful save
- [x] If so, save immediately.
- [x] The 5-second timer is stopped-resetted when a save request is sent. So, the countdown starts once we know if the request was a success or a failure.
- [x] By design, there should not be parallel save.

ThomasJuster:
## Description

Fixes #776

## Notes

...

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. When I navigate to [this task](https://dev.algorea.org/branch/feat/save-task-interval/en/#/activities/by-id/4080996542681828;path=4702,1352246428241737349,314613032161178344,310572474623192846;attempId=0/details)
  3. And I open developers tools on network tab
  4. And I throttle the network connection to slow it down (:bulb: to make the request exceed 5s, I created a profile with `download: 3kbit/s` and `latency: 1000ms`)
  5.  And I type code in the editor
  6. Then I must see requests every 5 seconds _after_ previous request ended, and requests lasting more than 5 seconds should not be cancelled
  :warning: Request non-cancellation cannot be tested on branch app because there is a timeout cancelling xhr requests after 3s. It can be tested locally: change `DEFAULT_TIMEOUT` value to >5s (ie: 10s) in app module providers OR lower the value of answer and state interval in item-task-answer service

### DIFF
--- a/src/app/modules/item/services/item-task-answer.service.ts
+++ b/src/app/modules/item/services/item-task-answer.service.ts
@@ -60,7 +60,7 @@ export class ItemTaskAnswerService implements OnDestroy {
   readonly saveAnswerAndStateInterval$ = this.task$.pipe(
     delayWhen(() => combineLatest([ this.initializedTaskState$, this.initializedTaskAnswer$ ])),
     switchMap(task => forkJoin({ answer: task.getAnswer(), state: task.getState() })),
-    switchMap(({ answer, state }) => this.saveAnswerAndStateInterval(answer, state)),
+    switchMap(saved => this.saveAnswerAndStateInterval(saved.answer, saved.state)),
     shareReplay(1),
   );
 
@@ -121,7 +121,7 @@ export class ItemTaskAnswerService implements OnDestroy {
       switchMap(task => forkJoin({ answer: task.getAnswer(), state: task.getState() })),
       startWith({ answer: savedAnswer, state: savedState }), // start with saved answer & state to keep only changed answer & state
       distinctUntilChanged((a, b) => a.answer === b.answer && a.state === b.state),
-      skip(1), // skip currently saved answer & state ...
+      skip(1), // skip currently saved answer & state tuple ...
       take(1), // ... and take next unsaved one
       withLatestFrom(this.config$),
       switchMap(([{ answer, state }, { route, attemptId }]) =>
--- a/src/app/modules/item/services/item-task-answer.service.ts
+++ b/src/app/modules/item/services/item-task-answer.service.ts
@@ -13,6 +13,7 @@ import {
   switchMapTo,
   take,
   takeUntil,
+  tap,
   withLatestFrom,
 } from 'rxjs/operators';
 import { SECONDS } from 'src/app/shared/helpers/duration';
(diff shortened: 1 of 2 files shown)


src/app/modules/item/services/item-task-answer.service.ts
```diff
@@ -121,4 +114,26 @@ export class ItemTaskAnswerService implements OnDestroy {
...
+      skip(1), // skip currently saved answer & state tuple ...
+      take(1), // ... and take next unsaved one
+      withLatestFrom(this.config$),
+      switchMap(([{ answer, state }, { route, attemptId }]) =>
+        this.currentAnswerService.update(route.id, attemptId, { answer, state }).pipe(
+          mapTo({ success: true }),
+          catchError(() => of({ success: false })),
+          switchMap(result => concat(
+            of(result), // emit success state
+            this.saveAnswerAndStateInterval(answer, state)), // restart treatment
```
smadbe, line 133: I cannot convince myself that it is a good thing to do this in Rx, but I acknowledge it is not an easy case.
Wasn't it achievable using `repeatWhen`?

smadbe (approved):

src/app/modules/item/services/item-task-answer.service.ts
ThomasJuster: I tested behavior of `repeat` operator (using [rxviz](https://rxviz.com/)), it appears to be exactly what we need here :ok_hand: 
I'll also retest the app

src/app/modules/item/services/item-task-answer.service.ts
```diff
@@ -121,4 +114,26 @@ export class ItemTaskAnswerService implements OnDestroy {
   clearAnswer(): Observable<unknown> {
     return this.task$.pipe(take(1), switchMap(task => task.reloadAnswer('')));
   }
+
+  private saveAnswerAndStateInterval(savedAnswer: string, savedState: string): Observable<{ success: boolean }> {
+    return interval(answerAndStateSaveInterval).pipe(
+      switchMap(() => this.task$),
```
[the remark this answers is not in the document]
ThomasJuster: fixed

src/app/modules/item/services/item-task-answer.service.ts
ThomasJuster: It works, the only problem I had was :
I can't repeat as is the observable because I need to take last _saved_ values at each repeat. I opted for an internal thunk `saved` holding the saved values, and I use this object to filter from the task only unsaved values

It looks more rx, but I find not completely satisfying. I had other methods in mind like using a Su bject, but technically it's same result: when receiving a result, populate a variable external to the observable with saved values and re-read it in observable next value to apply a filter operation

src/app/modules/item/services/item-task-answer.service.ts
```diff
@@ -145,4 +139,26 @@ export class ItemTaskAnswerService implements OnDestroy {
...
+      switchMapTo(this.task$),
+      switchMap(task => forkJoin({ answer: task.getAnswer(), state: task.getState() })),
+      distinctUntilChanged((a, b) => a.answer === b.answer && a.state === b.state),
+      filter(({ answer, state }) => answer !== saved.answer || state !== saved.state),
+      take(1),
+      withLatestFrom(this.config$),
+      switchMap(([{ answer, state }, { route, attemptId }]) =>
+        this.currentAnswerService.update(route.id, attemptId, { answer, state }).pipe(
+          map(() => {
+            Object.assign(saved, { answer, state });
```
smadbe, line 155: this is a side effect, it should be done in a ` tap` so that it is clean that it is a dirty side effect

smadbe (approved):
This side effect is not great. 
That looks like a case where using a `scan` (or `switchScan` probably) should help.
But no offense if you merge it as it is :-)

PR #819, opened 2021-11-24
fix language race condition when loading item

### LINKED ISSUE #797: Race condition when changing language
**Describe the bug**
At app startup on item page for an anonymous user, two operations are triggered simultaneously:
1. If the current user language does not match app language, update current user language
2. Fetch item content (item, breadcrumbs, …, see item data source)

The first case happens only when an anonymous user changes the language using the app language picker

Check if the race condition also occurs when starting app on a group page

**To Reproduce**
1. As an anonymous user
2. Go to [the app](https://dev.algorea.org/)
3. Open the devtools to check xhr requests
4. Change language using the app language picker
5. The app is reloaded
6. **Sometimes** `PUT /current-user` occurs after `GET /item` and/or `GET /breadcrumbs`

**Expected behavior**

- [ ] The item content (and more generically page content) fetching must be triggered _after_ the current user language is updated.
- [ ] If the current user language is **not** updated, page content should be loaded straight away.

ThomasJuster:
## Description

Fixes #797

Groups are not concerned by the issue since they don't have localized info.

However since the aim is to avoid having server response badly localized, I think the authenticated users should be handled as well. The issue does not mention anything about it, so I made an exhaustive list of use cases:
1. No language mismatch :arrow_right: can start fetching localized content
2. Language mismatch
    1. User is anonymous/temp :arrow_right: update user language :arrow_right: can start fetching localized content
    2. User is logged (aka **not** anonymous)
        1. User clicks on "Visit platform in [his/her] language" :arrow_right: No-op since user left the app
        2. User clicks on "Update my language" :arrow_right: update user language :arrow_right: can start fetching localized content

## Test cases

:bulb: for each test case, throttling the network helps viewing in detail what's happening.

- [ ] Case 1:
  2. When I navigate to
      - [this activity in "en"](https://dev.algorea.org/branch/fix/language-race-condition/en/#/activities/by-id/128139237432513103;path=4702,4102;parentAttempId=0/details)
      - [this skill in "en"](https://dev.algorea.org/branch/fix/language-race-condition/en/#/skills/by-id/1133373964112078654;path=3000,3002;parentAttempId=0/details)
  3. Then the app starts normally and the title should always be displayed in English

- [ ] Case 2:
  2. And I open developer tools to network panel and throttle the network to bad connection
  3. When I navigate to
      - [this activity in "fr"](https://dev.algorea.org/branch/fix/language-race-condition/fr/#/activities/by-id/128139237432513103;path=4702,4102;parentAttempId=0/details)
      - [this skill in "fr"](https://dev.algorea.org/branch/fix/language-race-condition/fr/#/skills/by-id/1133373964112078654;path=3000,3002;parentAttempId=0/details)
  4. Then I must see that item requests are hold until the current user language is updated, and the title should be displayed in French

- [ ] Case 3:
  1. Given I am a logged user with "en" as default language
  2. When I navigate to
      - [this activity in "en"](https://dev.algorea.org/branch/fix/language-race-condition/en/#/activities/by-id/128139237432513103;path=4702,4102;parentAttempId=0/details)
      - [this skill in "en"](https://dev.algorea.org/branch/fix/language-race-condition/en/#/skills/by-id/1133373964112078654;path=3000,3002;parentAttempId=0/details)
  3. Then app starts normally and the title should be displayed in English

- [ ] Case 4:
  1. Given I am a logged user with "en" as default language
  2. And I open developer tools to network panel and throttle the network to bad connection
  3. When I navigate to
      - [this activity in "fr"](https://dev.algorea.org/branch/fix/language-race-condition/fr/#/activities/by-id/128139237432513103;path=4702,4102;parentAttempId=0/details)
      - [this skill in "fr"](https://dev.algorea.org/branch/fix/language-race-condition/fr/#/skills/by-id/1133373964112078654;path=3000,3002;parentAttempId=0/details)
  4. And I click on "update my language"
  5. Then I must see that item requests are hold until the current user language is updated and the title must be displayed in French

### DIFF
--- a/src/app/core/components/language-mismatch/language-mismatch.component.ts
+++ b/src/app/core/components/language-mismatch/language-mismatch.component.ts
@@ -1,6 +1,6 @@
 import { Component, OnDestroy } from '@angular/core';
-import { EMPTY } from 'rxjs';
-import { catchError, filter, map, retry, switchMap, takeUntil } from 'rxjs/operators';
+import { EMPTY, merge, Subject } from 'rxjs';
+import { catchError, defaultIfEmpty, filter, map, retry, shareReplay, switchMap, take, takeUntil } from 'rxjs/operators';
 import { mapPending } from 'src/app/shared/operators/map-pending';
 import { UserSessionService } from 'src/app/shared/services/user-session.service';
 import { LocaleService } from '../../services/localeService';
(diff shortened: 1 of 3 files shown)


smadbe (approved):
Looks like a way to solve the problem.

PR #825, opened 2021-11-29
Feature/fix score rings

### LINKED ISSUE #804: Fix score rings
For instance, for chapter: https://dev.algorea.org/en/#/activities/by-id/34689573454313988;path=4702,1352246428241737349,314613032161178344;attempId=0/details

# Big score rings
Actual: 
Screenshot 2021-11-01 at 12 20 04

Expected
Screenshot 2021-11-01 at 12 20 13

- [ ] The digits should be larger, possibly lighter weight. The text size should be always proportional to the circle size.

# Success rings 

Actual
Screenshot 2021-11-01 at 12 21 42
Screenshot 2021-11-01 at 12 41 30

Expected:
- [ ] When on a **light** background, use `rgb(105, 206, 77)` as green color so that it is more visible. 
- [ ] On the chapter list, the "tick" should be visible on validated/full score

Iloveall:
## Description

Fixes #804

I've remove Math.floor() as it was reason disappearing successful icon, font-size was 0 rem

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/fix-score-rings/en/#/activities/by-id/34689573454313988;path=4702,1352246428241737349,314613032161178344;attempId=0/details)
  3. I can see font size proportional to score ring 
  4. I can see successful icon in list under content tab
 
- [x] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/fix-score-rings/en/#/activities/by-id/784078458149384669;path=4702,1352246428241737349,314613032161178344,34689573454313988;attempId=0/details)
  3. I can see changed green color for score ring successful icon

### DIFF
--- a/src/app/modules/shared-components/components/score-ring/score-ring.component.html
+++ b/src/app/modules/shared-components/components/score-ring/score-ring.component.html
@@ -32,8 +32,10 @@
       fill-opacity="0.4"
     />
   </svg>
-  <span class="ring-icon" [ngStyle]="{color: icon ? iconFill : textFill}">
+  <span class="ring-data" [ngStyle]="{color: icon ? iconFill : textFill}">
     <i *ngIf="icon" class="fa fa-{{icon}}" [ngStyle]="{'font-size.rem': fontSize}"></i>
-    <ng-container *ngIf="!icon">{{currentScore | number:'1.0-0'}}</ng-container>
+    <ng-container *ngIf="!icon">
+      <span class="ring-digits" [ngStyle]="{'font-size.rem': fontSize}">{{currentScore | number:'1.0-0'}}</span>
+    </ng-container>
   </span>
 </div>
--- a/src/app/modules/shared-components/components/score-ring/score-ring.component.html
+++ b/src/app/modules/shared-components/components/score-ring/score-ring.component.html
@@ -1,6 +1,6 @@
 <div class="ring-container" [ngStyle]="{width: forTree ? '100%' : diameter / 12 + 'rem', height: forTree ? '100%' : diameter / 12 + 'rem'}">
   <svg
-    viewBox="0 0 64 64"
+    viewBox="0 0 65 65"
   >
     <path
       stroke="black"
--- a/src/app/modules/shared-components/components/score-ring/score-ring.component.ts
+++ b/src/app/modules/shared-components/components/score-ring/score-ring.component.ts
@@ -1,7 +1,7 @@
 import { Component, Input, OnChanges, SimpleChanges, ViewChild, ElementRef } from '@angular/core';
 
 enum ScoreRingColor {
-  Success = '#B8E986',
+  Success = 'rgb(105, 206, 77)',
   Initial = '#F90223',
   DefaultText = '#4A4A4A',
   DarkText = '#FFF',
@@ -86,7 +86,6 @@ export class ScoreRingComponent implements OnChanges {
       this.icon = undefined;
     }
 
-
-    this.fontSize = this.forTree ? 1 : Math.floor((2 * this.diameter) / 64);
+    this.fontSize = (2 * this.diameter) / 56;
   }
 }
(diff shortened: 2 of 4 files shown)


src/app/modules/shared-components/components/score-ring/score-ring.component.html
```diff
@@ -32,8 +32,10 @@
       fill-opacity="0.4"
```
ThomasJuster, line 32: On [this page] as usual user, score rings are a bit truncated on firefox:
image

src/app/modules/shared-components/components/score-ring/score-ring.component.html
```diff
@@ -32,8 +32,10 @@
...
     />
```
ThomasJuster, line 33: On same screenshot, the inner horizontal alignment is _slightly_ imperfect, for me it's particularly visible on `0` (I have it for both firefox and chrome)
screenshot

src/app/modules/shared-components/components/score-ring/score-ring.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, Input, OnChanges, SimpleChanges, ViewChild, ElementRef } from '@angular/core';
 
 enum ScoreRingColor {
-  Success = '#B8E986',
+  Success = 'rgb(105, 206, 77)',
```
ThomasJuster, line 4: I wonder if we should use the equivalent hex notation to keep consistency with the other colors below, your call

src/app/modules/shared-components/components/score-ring/score-ring.component.html
Iloveall: It depends on font settings, 

image

image

As you can see 100 has some padding on left, 86 has another. Horizontal alignment now does with flex, so it better then before. Also different browsers works with fonts with different way.

src/app/modules/shared-components/components/score-ring/score-ring.component.ts
Iloveall: Hex Fixed

src/app/modules/shared-components/components/score-ring/score-ring.component.html
Iloveall: I've add some fix, could you test pls, because I have another behaviour on my Firefox

ThomasJuster: Works fine :+1:

src/app/modules/shared-components/components/score-ring/score-ring.component.ts
ThomasJuster: Yes as component input is ok, I think by default it should be adapted to a light background because it's the most common use case

ThomasJuster (commented):
I responded to a comment [here](https://github.com/France-ioi/AlgoreaFrontend/pull/825#discussion_r760891607), I'll re-review after

src/app/modules/shared-components/components/score-ring/score-ring.component.ts
Iloveall: Fixed

src/app/modules/shared-components/components/score-ring/score-ring.component.ts
```diff
@@ -22,6 +23,7 @@ export class ScoreRingComponent implements OnChanges {
   @Input() icon?: string; // a font-awesome icon identifier
   @Input() scoreFillColor?: string;
   @Input() forTree = false;
+  @Input() isDarkBg = false;
```
ThomasJuster, line 26: I think the name `isDarkBg` is misleading by indicating a dark background inside the score-ring component, like `isDark` changes the _inner_ text color
I think something like `forDarkBg` would be more suited, not sure though.
What do you think ?

Iloveall: Perfect, fixed

ThomasJuster (commented):
I found another spot where the green should be for dark bg [here](https://dev.algorea.org/branch/feature/fix-score-rings/en/#/activities/by-id/784078458149384669;path=4702,1352246428241737349,314613032161178344,34689573454313988;attempId=0/details/progress/chapter) (as usual user)
image

ThomasJuster (approved):
> Hi Alex. For the green on the orange icon, leave them as before ... we'll fix the orange first.

All good then :+1:

PR #842, opened 2021-12-06
Provide a "bare" version of the plateform

### LINKED ISSUE #837: (LTI) Provide a "bare" version of the plateform
## Overview

We need to be able to display a bare version of the platform for using it a LTI provider.

## Acceptance criteria
- [ ] when a specific argument (query string?) is given to the platform: 
  - [ ] do not display the left (same as full frame but with no way to show the menu)
  - [ ] do not display the black top-right user menu

ThomasJuster:
## Description

Fixes #837 

I'm unsure about how to determine when the platform is run as LTI provider, this specific part will − I think − require some discussion, especially if we need to _persist_ the "run as LTI provider" indicator in the URL.

For information: currently, the rule to determine if the platform is run as LTI provider is only `location.hash.includes('asLTIProvider')`, this check is performed _only_ at app init.

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. When I go to [this item page](https://dev.algorea.org/branch/feat/bare-platform/en/#/activities/by-id/9105453971757450756;path=4702,1625159049301502151;attempId=0;asLTIProvider=/details) (note that `'asLTIProvider'` is passed as path parameter)
  3. Then I see the left menu is hidden with no possibility to expand it
  4. Then I also see the top right controls are **not** displayed
  5. And I navigate to prev sibling using navigation
  6. Then I see the content remains collapsed although it's a chapter

## No regression tests (due to `layoutService.fullFrame` structure change)

- [ ] Group page:
  1. Given I am usual user
  2. When I navigate to [my groups](https://dev.algorea.org/branch/feat/bare-platform/en/#/groups/by-id/52767158366271444;path=/details)
  3. Then I should see the group name under the top bar, here "4LTDI1819"
  4. And I collapse the left menu
  5. Then I should not see the group name under the top bar anymore
 
- [ ] My group page:
  1. Given I am usual user
  2. When I navigate to [this group](https://dev.algorea.org/branch/feat/bare-platform/en/#/groups/mine)
  3. Then I should see the title "My groups" under the top bar
  4. And I collapse the left menu
  5. Then I should not see the title "My groups" under the top bar anymore
 

- [ ] User page:
  1. Given I am usual user
  2. When I navigate to [self user profile page](https://dev.algorea.org/branch/feat/bare-platform/en/#/groups/mine)
  3. Then I should see the my name "Armelle …" under the top bar
  4. And I collapse the left menu
  5. Then I should not see my name "Armelle …" under the top bar anymore

- [ ] Item page:
  1. Given I am any user
  2. When I navigate to [this item](https://dev.algorea.org/branch/feat/bare-platform/en/#/activities/by-id/5207993233955027100;path=4702,1625159049301502151;attempId=0/details) (Thomas tasks)
  3. And I click on first item **in item content, not left menu**
  4. Then left menu is collapsed
  5. And I navigate to parent using navigation arrows in header
  6. Then left menu expanded, I see the chapter name "Thomas tasks" under the top bar
  7. And I collapse the left menu
  8. Then I don't see the chapter name "Thomas tasks" under the top bar anymore

### DIFF
--- a/src/app/shared/services/layout.service.ts
+++ b/src/app/shared/services/layout.service.ts
@@ -5,7 +5,6 @@ import { delay } from 'rxjs/operators';
 export interface FullFrameContent {
   expanded: boolean,
   canToggle: boolean,
-  showTopRightControls: boolean,
 }
 
 @Injectable({
@@ -17,11 +16,13 @@ export class LayoutService {
 
   private fullFrameContent = new BehaviorSubject<FullFrameContent>(
     this.platformAsLTIProvider
-      ? { expanded: true, canToggle: false, showTopRightControls: false }
-      : { expanded: false, canToggle: true, showTopRightControls: true }
+      ? { expanded: true, canToggle: false }
+      : { expanded: false, canToggle: true }
   );
   /** Expands the content by hiding the left menu and select headers */
   fullFrameContent$ = this.fullFrameContent.asObservable().pipe(delay(0));
+
+  readonly showTopRightControls = !this.platformAsLTIProvider;
 
   private contentFooter = new BehaviorSubject<boolean>(true);
   /**
--- a/src/app/core/app.component.html
+++ b/src/app/core/app.component.html
@@ -10,6 +10,7 @@
       [scrolled]="scrolled"
       [currentMode]="currentMode"
       [fullFrameContent]="fullFrameContent"
+      [showTopRightControls]="showTopRightControls"
     ></alg-top-bar>
     <div
       class="right"
--- a/src/app/core/app.component.ts
+++ b/src/app/core/app.component.ts
@@ -26,6 +26,7 @@ export class AppComponent implements OnInit, OnDestroy {
   );
 
   fullFrameContent$ = this.layoutService.fullFrameContent$;
+  showTopRightControls = this.layoutService.showTopRightControls;
   contentFooter$ = this.layoutService.contentFooter$;
   scrolled = false;
 
--- a/src/app/core/components/top-bar/top-bar.component.html
+++ b/src/app/core/components/top-bar/top-bar.component.html
@@ -5,5 +5,5 @@
     [fullFrameContent]="fullFrameContent"
     [scrolled]="scrolled"
   ></alg-content-top-bar>
-  <alg-top-right-controls *ngIf="fullFrameContent?.showTopRightControls" [drawLeftBorder]="currentMode === 'normal'"></alg-top-right-controls>
+  <alg-top-right-controls *ngIf="showTopRightControls" [drawLeftBorder]="currentMode === 'normal'"></alg-top-right-controls>
 </ng-container>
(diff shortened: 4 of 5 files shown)


src/app/shared/services/layout.service.ts
```diff
@@ -1,16 +1,23 @@
...
 export class LayoutService {
   // Service allowing modifications of the layout
+  private platformAsLTIProvider = location.hash.includes('asLTIProvider');
+  readonly hideLeftMenu = this.platformAsLTIProvider;
+  readonly hideTopRightControls = this.platformAsLTIProvider;
 
   private fullFrameContent = new BehaviorSubject<boolean>(false);
   /** Expands the content by hiding the left menu and select headers */
-  fullFrameContent$ = this.fullFrameContent.asObservable().pipe(delay(0));
+  fullFrameContent$ = this.fullFrameContent.asObservable().pipe(
```
smadbe, line 16: I think it may be misleading to have  a `$` var which does not have exactly the same meaning as the non$ one.

smadbe (changes requested):

src/app/shared/services/layout.service.ts
ThomasJuster: Fixed

src/app/shared/services/layout.service.ts
```diff
@@ -7,8 +7,11 @@ import { delay } from 'rxjs/operators';
 })
 export class LayoutService {
   // Service allowing modifications of the layout
+  private platformAsLTIProvider = location.hash.includes('asLTIProvider');
+  readonly hideLeftMenu = this.platformAsLTIProvider;
+  readonly hideTopRightControls = this.platformAsLTIProvider;
 
-  private fullFrameContent = new BehaviorSubject<boolean>(false);
+  private fullFrameContent = new BehaviorSubject<boolean>(this.hideLeftMenu);
```
smadbe, line 14: Hum it seems I forgot to submit the main remark in the previous review 😔
So it was:
My bad, the spec was not very clear but the bare state is: everything like full frame content, but with top-right hidden as well and impossibility to leave the full frame mode.
So I don't like the way there is are two independent (but not independent at all) states: `fullFrameContent` and `hideLeftMenu`. 
As I always try to avoid as much as possible non-minimal state and dependency between "instance" variables, I would say that "forbid leaving full frame" and "hide top right" should just be an option of the full frame content (so that "hide top right" does not make sense with the menu, which prevent having to handle this case properly in the design)

smadbe (changes requested):

smadbe (approved):
👍

smadbe (changes requested):
Approval was a mistake.

src/app/shared/services/layout.service.ts
ThomasJuster: While "forbid leaving full frame" as a `fullFrameContent` option makes sense, I doubt "hide top right" is meaningful.
I don't understand why hiding those top right controls has anything to do with full frame mode, as I understand it here we want to hide top right controls because the platform is run as LTI provider, and as it happens we want to force full frame for the same reason, but I don't think it makes the 2 states/variables dependants on each other

PR #843, opened 2021-12-07
LTI content parameter support

ThomasJuster:
## Description

Fixes #...

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  3. And I click on ...
  4. Then I see ...
  5. And

- [ ] Case 2:
  1. Given I am the usual user
  2. When ...


src/app/core/services/lti.service.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@angular/core';
```
smadbe, line 1: If you do an LTI module, I would expect more the service to be in it.
This would require to work differently, instead of having the layout service depending on it, the LTIService would require to "push" what's required on the core services... which is also cleaner as the core services would not know about LTI or any optional additions in the future.
I haven't read the code yet, so I don't know if that's possible, let me know your thoughts about that.

ThomasJuster: I intended to put it in the LTI module at first, in the end I didn't because the layout service relies on it, and the layout service is declared (logically) at app-level.
I'll try reversing the dependency order, thanks to your comment I have another idea that will do well :wink:

ThomasJuster: Fixed
Architecture is now much cleaner, however now there's a blink at startup in LTI mode: the left nav is displayed until the LTI module is being lazily fetched because the LTI component now is the one changing the app layout in its constructor.
Is that way ok or should I change it back ?

smadbe:
I am surprised by case 4, it looks like a misunderstanding of how the platform (in non-LTI) work. Users are often new when using LTI, they may not have prior "result" but that's not a problem. As any item in non-LTI, if there are no result, a result is auto-created if the permissions allows it and if not explicit-entry.

smadbe (changes requested):
If I get an error page (for instance https://dev.algorea.org/branch/feat/lti-content-parameter-support/en/#/lti?content_id=9105453971757450756) and then I change the url to a valid page (for instance https://dev.algorea.org/branch/feat/lti-content-parameter-support/en/#/lti?content_id=1625159049301502151), the error stays.

ThomasJuster:
> If I get an error page (for instance https://dev.algorea.org/branch/feat/lti-content-parameter-support/en/#/lti?content_id=9105453971757450756) and then I change the url to a valid page (for instance https://dev.algorea.org/branch/feat/lti-content-parameter-support/en/#/lti?content_id=1625159049301502151), the error stays.

Fixed

ThomasJuster:
> I am surprised by case 4, it looks like a misunderstanding of how the platform (in non-LTI) work. Users are often new when using LTI, they may not have prior "result" but that's not a problem. As any item in non-LTI, if there are no result, a result is auto-created if the permissions allows it and if not explicit-entry.

Fixed, I changed case 4

smadbe (changes requested):
Case 1: 
- I visit the given link.
- the content is not ""Bienvenue sur Algoréa"" as written
- the navigation on top of the task offers to navigate to the previous item, which is strange as I am supposed to be on a first child

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -0,0 +1,105 @@
...
+      next: state => {
+        if (!state.isReady) return;
+        const { firstChild, path, result } = state.data;
+        const itemRoute = fullItemRoute('activity', firstChild.id, [ ...path, state.data.itemId ], { attemptId: result.attemptId });
+        this.itemRouter.navigateTo(itemRoute, { navExtras: { replaceUrl: true } });
+      },
+    });
+  }
+
+  private getNavigationData(itemId: string): Observable<{ itemId: string, firstChild: ItemChild, path: string[], result: Result }> {
```
smadbe, line 72: it seems to do all the call 3x 
Screenshot 2021-12-15 at 18 15 54

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -0,0 +1,105 @@
...
+    const item$ = this.getItemByIdService.get(itemId).pipe(shareReplay(1));
+    const path$ = this.getItemPathService.getItemPath(itemId).pipe(shareReplay(1));
+    const results$ = this.getResultsService.get(itemId, { attempt_id: '0' });
+
+    const bestResult$ = forkJoin([ item$, results$ ]).pipe(
+      switchMap(([ item, results ]) => {
+        if (results.length === 0 && !implicitResultStart(item)) throw explicitEntryWithNoResultError;
+        return path$.pipe(switchMap(path => this.getBestResultOrStartOne(item, results, path)));
+      }),
+      shareReplay(1),
```
smadbe, line 82: You are taking a wrong direction on the results/attempts. You assume that your can rely on the parent attempt "0" but on the results you search for the best result. Let's make it simpler.
As you don't know you parent attempt, no need for all this, just `startWithoutAttempt`  when you know you item path. You should get an attempt that you can use.
Also, you probably don't need to fetch the item info, if "startWithoutAttempt" fails, that's probably because you or one of your ancestors is implicit entry, no need for extra checks.
It is not clear to me at this state (I have read the rest of the code yet) what you need ... but I assume that with these two calls, you'll get the item path and the attempt id, that's enough to fetch the children.
That should simplify a lot your code.

smadbe (changes requested):

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: Fixed
path-from-root fails before start-without-attempt, I'll handle it this way

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: Fixed

src/app/modules/item/http-services/get-results.service.ts
```diff
@@ -55,10 +55,13 @@ export class GetResultsService {
 
   constructor(private http: HttpClient) {}
 
-  getResults(item: FullItemRoute): Observable<Result[]> {
+  get(
+    itemId: string,
+    params?: { attempt_id?: string, parent_attempt_id?: string, limit?: number, sort?: Array<'id' | '-id'> },
```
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -0,0 +1,105 @@
...
+import { ResultActionsService } from 'src/app/shared/http-services/result-actions.service';
+import { mapToFetchState } from 'src/app/shared/operators/state';
+import { fullItemRoute } from 'src/app/shared/routing/item-route';
+import { ItemRouter } from 'src/app/shared/routing/item-router';
+import { LayoutService } from 'src/app/shared/services/layout.service';
+
+const explicitEntryWithNoResultError = new Error('no result');
+const noChildError = new Error('no first child');
+const noContentIdError = new Error('no content id');
+type LTIError = 'no content id' | 'fetch error' | 'explicit entry with no result' | 'no child';
```
smadbe, line 20: Remark (not to be changed necessarily): usually, in this project, string union types are done in snake case... the risk I see with plain text is to have later a developer use them in strings. Actually we just use union types instead of enums is that enums are boring to use in templates... btw, my latest way to handle this was to do enum with string values (for instance `TypeFilter` or `GroupManagership` or `Mode`).
Also, there is a similarity/duplication between `'no first child'` of the error and `'no child'` ... it may be misleading...  as `'no child'`  is in both the enum and the error.

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -0,0 +1,105 @@
...
+    private getItemPathService: GetItemPathService,
+    private resultActionsService: ResultActionsService,
+    private getItemChildrenService: GetItemChildrenService,
+  ) {
+    this.layoutService.toggleTopRightControls(false);
+    this.layoutService.toggleFullFrameContent(true, false);
+
+    this.navigationData$.subscribe({
+      next: state => {
+        if (!state.isReady) return;
```
smadbe, line 63: maybe filter only the ready state in a pipe, it might be clearer.

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -0,0 +1,105 @@
...
+  templateUrl: './lti.component.html',
+  styleUrls: [ './lti.component.scss' ],
+})
+export class LTIComponent {
+
+  readonly navigationData$ = this.activatedRoute.queryParamMap.pipe(
+    switchMap(queryParams => {
+      const contentId = queryParams.get('content_id');
+      if (!contentId) return of(errorState(noContentIdError));
+      return this.getNavigationData(contentId).pipe(mapToFetchState());
```
smadbe, line 33: the only reason to use `mapToFetchState` is to avoid fetching errors failing the whole subscription to route?

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -0,0 +1,105 @@
...
+      catchError(err => {
+        // If error is http forbidden, it PROBABLY means the item requires explicit entry.
+        throw errorIsHTTPForbidden(err) ? explicitEntryWithNoResultError : err;
+      }),
+      switchMap(path => this.getResultsService.getResults(fullItemRoute('activity', itemId, path, { attemptId: defaultAttemptId }))),
+      map(results => {
+        const result = bestAttemptFromResults(results);
+        if (!result) throw explicitEntryWithNoResultError;
+        return result;
+      }),
```
smadbe, line 85: I still don't get it. Why don't you use the attempt returned by `startWithoutAttempt` instead of the getting a results assuming it has `defaultAttemptId` (if we assume it has `defaultAttemptId`, why not using it directly?)

smadbe (changes requested):

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: fixed

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: yes

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: fixed

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: fixed

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -0,0 +1,105 @@
...
+export class LTIComponent {
+
+  readonly navigationData$ = this.activatedRoute.queryParamMap.pipe(
+    switchMap(queryParams => {
+      const contentId = queryParams.get('content_id');
+      if (!contentId) return of(errorState(noContentIdError));
+      return this.getNavigationData(contentId).pipe(mapToFetchState());
+    }),
+    shareReplay(1),
+  );
```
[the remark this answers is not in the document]
ThomasJuster: fixed (It was very quick to change)

src/app/modules/lti/pages/lti/lti.component.ts
smadbe: ah I didn't see you use the loading state 👍

smadbe (approved):
👍

PR #844, opened 2021-12-07
lti: restrict navigaton to given content

### LINKED ISSUE #841: (LTI) Only allow navigation under the given content
## Overview

In order to integrate with LTI as a provider, we need to only allow navigation under the given content.

## Acceptance criteria
- [ ] the item navigation (prev/next/top widget) should only allow navigating under (not including itself and but including its descendants) the content given as arg (content_id). In other words, the navigation should as usual but should be disabled on the "content_id" itself. (use the path for that)

ThomasJuster:
## Description

Fixes #841 

I tried to implement navigation restriction respecting dependency directions, it felt appropriate to place the logic in nav tree service since it's the one already holding the `navigationNeighbours$` observable.

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. When I go to [this chapter](https://dev.algorea.org/branch/feat/lti-restrict-navigation/en/#/lti/1625159049301502151)
  3. ~And I navigate to the parent using navigation arrows~ Then I cannot navigate to parent because the parent button is disabled
  4. And I navigate to any child at any depth
  5. Then I see the navigation works as usual

### DIFF
--- a/src/app/core/services/navigation/nav-tree.service.ts
+++ b/src/app/core/services/navigation/nav-tree.service.ts
@@ -1,4 +1,4 @@
-import { BehaviorSubject, combineLatest, merge, Observable, of, OperatorFunction, Subject } from 'rxjs';
+import { combineLatest, merge, Observable, of, OperatorFunction, Subject } from 'rxjs';
 import { delay, distinctUntilChanged, map, mapTo, mergeScan, shareReplay, startWith } from 'rxjs/operators';
 import { isDefined } from 'src/app/shared/helpers/null-undefined-predicates';
 import { repeatLatestWhen } from 'src/app/shared/helpers/repeatLatestWhen';
@@ -76,12 +76,10 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
     shareReplay(1),
   );
 
-  private navigationRootElementId$ = new BehaviorSubject<string | undefined>(undefined);
+  navigationNeighborsRestrictedToDescendantOfElementId: string | undefined = undefined;
   navigationNeighbors$: Observable<FetchState<NavigationNeighbors|undefined>> = this.state$.pipe(
-    repeatLatestWhen(this.navigationRootElementId$), // refresh when navigation root element changes
     mapStateData(navData => {
-      const navigationRootElementId = this.navigationRootElementId$.value;
-      if (!navData.selectedElementId || navData.selectedElementId === navigationRootElementId) return undefined;
+      if (!navData.selectedElementId) return undefined;
       const idx = navData.elements.findIndex(e => e.id === navData.selectedElementId);
       if (idx < 0) return undefined;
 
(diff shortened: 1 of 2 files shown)


src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -76,9 +76,12 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
     shareReplay(1),
   );
 
+  private navigationRootElementId$ = new BehaviorSubject<string | undefined>(undefined);
```
smadbe, line 79: From what I understand, `navigationRootElementId$` is only used in the context of restricting the navigation (stay "empty" in the regular case). 
If so:
- this (variable + functions) requires good comments in the code what it is for. As it is, it adds a lot of confusion in a service which is already one of the more complex of the app. You also need to explain what `undefined` means for this var.
- not sure `rootElement` is a good name for that. Ok I understand now that you have probably added the `navigation` prefix to show it applies to `navigationNeighbors` only but as everything is in this service in related to navigation, it does not work. I think we need something very explicit here... for instance... `restrictNeighborNavigationToDescendantOfElementId `

src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -76,9 +76,12 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
...
   navigationNeighbors$: Observable<FetchState<NavigationNeighbors|undefined>> = this.state$.pipe(
+    repeatLatestWhen(this.navigationRootElementId$), // refresh when navigation root element changes
     mapStateData(navData => {
-      if (!navData.selectedElementId) return undefined;
+      const navigationRootElementId = this.navigationRootElementId$.value;
```
smadbe, line 83: `Subject.value` should only be used in specific case when you exit Rx.
Here, you have cleaner way to combine the subject.

(alternative: you use regular variable and not a subject... anyway it is not supposed to change once the app is loaded)

src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -76,9 +76,12 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
...
+      if (!navData.selectedElementId || navData.selectedElementId === navigationRootElementId) return undefined;
```
smadbe, line 84: That's not the spec " should only allow navigating under (**not including itself** and but including its descendants)"

smadbe (changes requested):

src/app/core/services/navigation/nav-tree.service.ts
ThomasJuster: fixed

src/app/core/services/navigation/nav-tree.service.ts
ThomasJuster: fixed

src/app/core/services/navigation/nav-tree.service.ts
ThomasJuster: fixed

src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -76,6 +76,14 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
     shareReplay(1),
   );
 
+  /**
+   * In some cases (ie LTI), the user visits the app only to work on an exercise, chapter or task and then returns to the original platform.
+   * For that purpose, we want to be able to restrict in-app navigation to a specific chapter.
+   * This is done by hiding the left menu and allowing only neighbor navigation while disabling others.
+   * When this variable is undefined, no restriction is applied.
```
smadbe, line 83: `null` would probably more appropriate than `undefined`.

smadbe (approved):

src/app/core/services/navigation/nav-tree.service.ts
ThomasJuster: fixed

PR #846, opened 2021-12-13
Feature/group access page add granted permissions

### LINKED ISSUE #828: Group access page: add granted permissions (a basic version)
## Overview
Add on the group access page (pretty empty for the moment) information about permission granted to this group.

You can get inspired from this design even if it does not have a lot in common: 
image

## Acceptance criteria
- [ ] On the group access page, add a new section "Specific permissions given to the group on content"
- [ ] List, in this section, the result of [https://france-ioi.github.io/algorea-devdoc/api/#operation/grantedPermissionsView](grantedPermissionsView) service (with `descendants=0`). For each distinct item:
  - [ ] Display the item name (with link)
  - [ ] for each entry, a list (one line per entry) with:
     - [ ] "Given by <source_group name with link> to <group name with link>." (if "source_group" or "group" is the current group, just use "this group" without link)
     - [ ] On the right on the line, display a summary of the permission: just in text "Can view: ..." (we'll improve that in another PR)

Iloveall:
## Description

Fixes #828

I guess access page must be separated to sub components for avoid a lot someState$ variables inside.

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/group-access-page-add-granted-permissions/en/#/groups/by-id/120505918794412400;path=2012793249973596086/details/access)
  3. Then I see list of items and permissions for it.

- [ ] Case 2:
 1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/group-access-page-add-granted-permissions/en/#/groups/by-id/120505918794412400;path=2012793249973596086/details/access)
  3. And I click on `Given by [group]` should redirect to group page

### DIFF
--- a/src/app/modules/group/components/group-permissions/group-permissions.component.html
+++ b/src/app/modules/group/components/group-permissions/group-permissions.component.html
@@ -4,7 +4,7 @@
       <div class="caption">Can View: {{ permissions.canView | groupPermissionCaption }}</div>
     </li>
     <li class="caption-item">
-      <div class="caption">Can Grand View: {{ permissions.canGrantView | groupPermissionCaption }}</div>
+      <div class="caption">Can Grant View: {{ permissions.canGrantView | groupPermissionCaption }}</div>
     </li>
     <li class="caption-item">
       <div class="caption">Can Enter From: {{ permissions.canEnterFrom | date:'d/MM/y' }}</div>
--- a/src/app/modules/group/components/group-permissions/group-permissions.component.html
+++ b/src/app/modules/group/components/group-permissions/group-permissions.component.html
@@ -1,0 +1,16 @@
+<ng-container *ngIf="permissions">
+  <ul class="captions">
+    <li class="caption-item">
+      <div class="caption">Can View: {{ permissions.canView | groupPermissionCaption }}</div>
+    </li>
+    <li class="caption-item">
+      <div class="caption">Can Grand View: {{ permissions.canGrantView | groupPermissionCaption }}</div>
+    </li>
+    <li class="caption-item">
+      <div class="caption">Can Enter From: {{ permissions.canEnterFrom | date:'d/MM/y' }}</div>
+    </li>
+    <li class="caption-item">
+      <div class="caption">Until: {{ permissions.canEnterUntil | date:'d/MM/y' }}</div>
+    </li>
+  </ul>
+</ng-container>
--- a/src/app/modules/group/http-services/granted-permissions.service.ts
+++ b/src/app/modules/group/http-services/granted-permissions.service.ts
@@ -5,7 +5,7 @@ import { decodeSnakeCase } from '../../../shared/operators/decode';
 import { dateDecoder } from '../../../shared/helpers/decoders';
 import { appConfig } from '../../../shared/helpers/config';
 import { HttpClient, HttpParams } from '@angular/common/http';
-import { permissionsDecoder } from '../../../shared/http-services/group-permissions.service';
+import { permissionsDecoder } from '../../../shared/helpers/group-permissions';
 import { pipe } from 'fp-ts/function';
 
 const groupPermissionsDecoder = pipe(
--- a/src/app/modules/group/http-services/granted-permissions.service.ts
+++ b/src/app/modules/group/http-services/granted-permissions.service.ts
@@ -5,17 +5,18 @@ import { decodeSnakeCase } from '../../../shared/operators/decode';
 import { dateDecoder } from '../../../shared/helpers/decoders';
 import { appConfig } from '../../../shared/helpers/config';
 import { HttpClient, HttpParams } from '@angular/common/http';
+import { permissionsDecoder } from '../../../shared/http-services/group-permissions.service';
+import { pipe } from 'fp-ts/function';
 
-const groupPermissionsDecoder = D.struct({
-  canEdit: D.literal('none', 'children', 'all', 'all_with_grant'),
-  canEnterFrom: dateDecoder,
-  canEnterUntil: dateDecoder,
-  canGrantView: D.literal('none', 'enter', 'content', 'content_with_descendants', 'solution', 'solution_with_grant'),
-  canMakeSessionOfficial: D.boolean,
-  canView: D.literal('none', 'info', 'content', 'content_with_descendants', 'solution'),
-  canWatch: D.literal('none', 'result', 'answer', 'answer_with_grant'),
-  isOwner: D.boolean,
-});
+const groupPermissionsDecoder = pipe(
+  permissionsDecoder,
+  D.intersect(
+    D.struct({
+      canEnterFrom: dateDecoder,
+      canEnterUntil: dateDecoder,
+    })
+  ),
+);
 
 const grantedPermissionsDecoder = D.struct({
   group: D.struct({
--- a/src/app/modules/group/http-services/granted-permissions.service.ts
+++ b/src/app/modules/group/http-services/granted-permissions.service.ts
@@ -1,0 +1,56 @@
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import * as D from 'io-ts/Decoder';
+import { decodeSnakeCase } from '../../../shared/operators/decode';
+import { dateDecoder } from '../../../shared/helpers/decoders';
+import { appConfig } from '../../../shared/helpers/config';
+import { HttpClient, HttpParams } from '@angular/common/http';
+
+const groupPermissionsDecoder = D.struct({
+  canEdit: D.literal('none', 'children', 'all', 'all_with_grant'),
+  canEnterFrom: dateDecoder,
+  canEnterUntil: dateDecoder,
+  canGrantView: D.literal('none', 'enter', 'content', 'content_with_descendants', 'solution', 'solution_with_grant'),
+  canMakeSessionOfficial: D.boolean,
+  canView: D.literal('none', 'info', 'content', 'content_with_descendants', 'solution'),
+  canWatch: D.literal('none', 'result', 'answer', 'answer_with_grant'),
+  isOwner: D.boolean,
+});
+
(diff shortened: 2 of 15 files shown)


src/app/modules/group/components/group-permissions/group-permissions.component.html
```diff
@@ -0,0 +1,16 @@
+<ng-container *ngIf="permissions">
+  <ul class="captions">
+    <li class="caption-item">
+      <div class="caption">Can View: {{ permissions.canView | groupPermissionCaption }}</div>
+    </li>
+    <li class="caption-item">
+      <div class="caption">Can Grand View: {{ permissions.canGrantView | groupPermissionCaption }}</div>
```
ThomasJuster, line 7: Small typo: "Can Gran*t* View"

src/app/modules/group/http-services/granted-permissions.service.ts
```diff
@@ -0,0 +1,56 @@
...
+const groupPermissionsDecoder = D.struct({
+  canEdit: D.literal('none', 'children', 'all', 'all_with_grant'),
+  canEnterFrom: dateDecoder,
+  canEnterUntil: dateDecoder,
+  canGrantView: D.literal('none', 'enter', 'content', 'content_with_descendants', 'solution', 'solution_with_grant'),
+  canMakeSessionOfficial: D.boolean,
+  canView: D.literal('none', 'info', 'content', 'content_with_descendants', 'solution'),
+  canWatch: D.literal('none', 'result', 'answer', 'answer_with_grant'),
+  isOwner: D.boolean,
+});
```
ThomasJuster, line 18: This decoder is very similar to the one in group permissions http service, I think the decoder could be shared between group http service like it's done in item http service (for permissions as well)

src/app/modules/group/pages/group-access/group-access.component.html
```diff
@@ -27,6 +29,53 @@
...
+          <div>
+            <i class="icon fa fa-folder"></i>
+            <strong>
+              {{ grantedPermission.item.title }}.
+            </strong>
+            <span i18n>Given by</span>
+            <span
+              class="alg-link"
+              (click)="onGroupClick(grantedPermission.sourceGroup.id)"
+              *ngIf="grantedPermission.sourceGroup.id !== group?.id; else currentGroup">
```
ThomasJuster, line 62: The spec said "if source_group id or group id is same as current group", I think only the source_group is handled here

src/app/modules/group/pages/group-access/group-access.component.html
```diff
@@ -27,6 +29,53 @@
...
+                {{ grantedPermission.sourceGroup.name }}
+            </span>
+            <ng-template #currentGroup>
+              <span i18n> this group</span>
+            </ng-template>
```
ThomasJuster, line 67: The translation will be problematic here because "Given by" and "this group"/{group name} are split into 2 different translations, it breaks the structure of the sentence.
To handle it in a i18n-friendly way, we could (for instance) apply the if higher:
```html
<ng-container *ngIf="is different group condition …; else currentGroup">
  <ng-container i18n>
    <span>Given by</span>
    <span class="alg-link" …>{{ [source group name] }}</span>
  </ng-container>
</ng-container>
<ng-template #currentGroup>
  <span i18n>Given by this group</span>
</ng-template>
```

src/app/modules/group/pages/group-access/group-access.component.ts
```diff
@@ -13,11 +16,11 @@ import { mapToFetchState } from '../../../../shared/operators/state';
 export class GroupAccessComponent implements OnChanges, OnDestroy {
   @Input() group?: Group;
 
-  private readonly rootActivityId$ = new ReplaySubject<string | null>(1);
+  private readonly group$ = new ReplaySubject<Group>(1);
 
   private refresh$ = new Subject<void>();
```
ThomasJuster, line 21: Since there are 2 refresh now, we should probably rename this one as well

src/app/shared/pipes/groupPermissionCaption.ts
```diff
@@ -0,0 +1,26 @@
...
+  answer: $localize`Answer`,
+  answer_with_grant: $localize`Answer with grant`,
+};
+
+@Pipe({ name: 'groupPermissionCaption', pure: true })
+export class GroupPermissionCaptionPipe implements PipeTransform {
+  constructor() {}
+
+  transform(value: string): string {
+    return new Map<string, string>(Object.entries(PERMISSION_CAPTIONS)).get(value) || $localize`No caption`;
```
ThomasJuster, line 24: Why not just :arrow_down: ?
```js
PERMISSION_CAPTIONS[value] ?? $localize`No caption`
```

src/app/modules/group/pages/group-access/group-access.component.html
```diff
@@ -16,8 +16,10 @@
       <ng-container *ngIf="rootActivityState.isReady">
         <ng-container *ngIf="rootActivityState.data; else noActivity">
           <div class="item">
-            <i class="icon fa fa-folder"></i>
-            <strong i18n>Root activity: {{ rootActivityState.data.string.title }}</strong>
+            <div>
+              <i class="icon fa fa-folder"></i>
+              <strong i18n>Root activity: {{ rootActivityState.data.string.title }}</strong>
+            </div>
```
ThomasJuster, line 22: What is that new div for ?

src/app/modules/group/pages/group-access/group-access.component.html
```diff
@@ -27,6 +29,53 @@
...
+          <div class="item" *ngFor="let grantedPermission of permissionState.data">
+          <div>
+            <i class="icon fa fa-folder"></i>
+            <strong>
+              {{ grantedPermission.item.title }}.
+            </strong>
+            <span i18n>Given by</span>
+            <span
+              class="alg-link"
+              (click)="onGroupClick(grantedPermission.sourceGroup.id)"
```
ThomasJuster, line 61: Can you make it a router link ? You can probably use the `groupLink` pipe

src/app/modules/group/components/group-permissions/group-permissions.component.html
Iloveall: Fixed

src/app/modules/group/pages/group-access/group-access.component.html
Iloveall: Fixed

src/app/modules/group/pages/group-access/group-access.component.html
Iloveall: Fixed

src/app/modules/group/pages/group-access/group-access.component.ts
Iloveall: refresh$ is for rootActivityState$, which is another section. I have suggestion for that in optional comment of this task

src/app/shared/pipes/groupPermissionCaption.ts
Iloveall: I got error: TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type

src/app/modules/group/pages/group-access/group-access.component.html
Iloveall: Group access page is common and has common styles. New section with permissions requires flex rules, so I had to add div here for avoid flex effect.

src/app/modules/group/pages/group-access/group-access.component.html
Iloveall: Fixed

src/app/modules/group/http-services/granted-permissions.service.ts
Iloveall: Fixed

src/app/shared/pipes/groupPermissionCaption.ts
Iloveall: Fixed

src/app/shared/http-services/group-permissions.service.ts
```diff
@@ -8,7 +8,7 @@ import * as D from 'io-ts/Decoder';
 import { pipe } from 'fp-ts/function';
 import { decodeSnakeCase } from 'src/app/shared/operators/decode';
 
-const permissionsDecoder = pipe(
+export const permissionsDecoder = pipe(
```
ThomasJuster, line 11: I think this permission decoder should be in a helper so that both services can import it without having a dependency of one service on the other

src/app/modules/group/pages/group-access/group-access.component.html
```diff
@@ -27,6 +29,55 @@
...
+      <ng-container *ngIf="permissionState.isReady">
+        <ng-container *ngIf="permissionState.data.length > 1; else noPermissions">
+          <div class="item" *ngFor="let grantedPermission of permissionState.data">
+          <div>
+            <i class="icon fa fa-folder"></i>
+            <strong>
+              {{ grantedPermission.item.title }}.
+            </strong>
+            <ng-container *ngIf="grantedPermission.sourceGroup.id !== group?.id && grantedPermission.group.id !== group?.id; else currentGroup">
+              <span i18n>Given by</span>
```
ThomasJuster, line 59: I think the `i18n` attribute should be put upper in the ng-container, that way any further translator will have the context "Given by {{ groupName }}" instead of just "Given by" which makes a lot less sense for a translator

The aim is to provide some context for anyone who (will) translate the app

src/app/shared/http-services/group-permissions.service.ts
Iloveall: Fixed

src/app/modules/group/pages/group-access/group-access.component.html
Iloveall: Fixed

ThomasJuster (approved):

PR #849, opened 2021-12-21
Refactoring watching mode (passed in url)

smadbe:
## Description

Rework the watching mode so that the watching can triggered by url.

## Notes (out of scope, known isues, hints for reviewing code, ...)  

Out of scope: 
- handling of "mode" (especially through the ModeService) is still a bit dirty. As editing mode should disappear, it is out of the scope of this PR.
- the query parameter are removed by angular, we do not re-add them automatically
- there is no url prepared, ready to be copied for developer anywhere as discussed. it can be done later if needed

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I visit the [Pixal group](https://dev.algorea.org/branch/group-watching-in-url/en/#/groups/by-id/672913018859223173;path=52767158366271444/details) 
  3. And I open the network panel 
  4. And I click on watching
  4. Then I see the UI behave correctly
  5. And no extra network request has been done to fetch group info
  5. And I see the `watchedGroupId` and `watchUser=0` in url
  6. Then I click on the "test" suggested activity
  5. And I am still watching "Pixal"
  6. If I visit the progress tab > Chapter view, I am on the progress matrix
  7. Then I leave watching mode through the button in the yellow bar
  8. Then I see `?watchedGroupId=none&watchUser=0` in the url

- [x] Case 2:
  1. Given I am the usual user
  2. When I visit the [page of user usr_5p020x2thuyu](https://dev.algorea.org/branch/group-watching-in-url/en/#/groups/users/752024252804317630) 
  3. And I open the network panel 
  4. And I click on watching
  4. Then I see the UI behave correctly
  5. And no extra network request has been done to fetch user info
  5. And I see the `watchedGroupId` and `watchUser=1` in url
  6. Then I go to the "[test](https://dev.algorea.org/branch/group-watching-in-url/en/#/activities/by-id/39530140456452546;path=;attempId=0/details)" activity 
  5. And I am still watching "usr_5p020x2thuyu"
  6. If I visit the progress tab > Chapter view, I am on the list view
  7. Then when I add `?watchedGroupId=none` to the url
  8. Then I leave watching mode

- [x] Case 3:
  1. Given I am the usual user
  2. When I visit the [test activity](https://dev.algorea.org/branch/group-watching-in-url/en/#/activities/by-id/39530140456452546;path=;attempId=0/details) without being in watching mode
  3.   If I visit the progress tab > Chapter view, I am on the list view
  4. And I open the network panel 
  5. And I add `?watchedGroupId=672913018859223173&watchUser=0`
  4. Then I am watching the "Pixal" group
  6. And I see in the network panel that the group info has been fetched  `https://dev.algorea.org/api/groups/672913018859223173`


src/app/core/services/group-watching.service.ts
```diff
@@ -0,0 +1,108 @@
...
+import { BehaviorSubject, Observable, of } from 'rxjs';
+import { distinctUntilChanged, filter, map, shareReplay, startWith, switchMap, withLatestFrom } from 'rxjs/operators';
+import { GetGroupByIdService, Group } from 'src/app/modules/group/http-services/get-group-by-id.service';
+import { GetUserService, User } from 'src/app/modules/group/http-services/get-user.service';
+import { isNotUndefined } from 'src/app/shared/helpers/null-undefined-predicates';
+import { rawGroupRoute, RawGroupRoute } from 'src/app/shared/routing/group-route';
+
+const watchedGroupQueryParam = 'watchedGroupId';
+const watchedGroupIsUserQueryParam = 'watchUser';
+const noWatchingValue = 'none'; // value of `watchedGroupQueryParam` which means that there is no watching
```
ThomasJuster, line 12: The query params are weird to me because you can have unexpected combinations like `watchGroupId=none&watchUser=[whatever]`, `watchGroupId=[actualGroupId]&watchUser=toto`
Then based on a special value we will unwatch by returning `null` instead of `undefined`, meaning we end up with a distinction between `null` and `undefined`, looks smelly to me :grimacing: 

If I understand it correctly, what we want to do is:
- Be able to watch a group or a user by user navigation (a button) or url navigation (`watchedGroupId`)
- Be able stop watching by user navigation (a button) or url navigation (for dev/admin usage only I suppose ?)
- For a technical reason, be able to make the differentiate a watched group and a watched user

To me the query parameters handling should be more something like:
1. A query param `watchedGroupId=…` (like now) and check existence of param `watchedGroupIsUser` (for instance) instead of allowing any value as falsy
2. Two query params `watchedGroupId` and `watchedUserId`, but then we'd need to throw an error if both are supplied

As for the special value `'none'`, it could be replaced by the special case `watchedGroupId=` meaning the a watched group id exists but it's empty, or another query param `stopWatching`, but it induces to throw an error as well if both query params are supplied.

Thoughts ?

smadbe: > The query params are weird to me because you can have unexpected combinations like `watchGroupId=none&watchUser=[whatever]`, `watchGroupId=[actualGroupId]&watchUser=toto` Then based on a special value we will unwatch by returning `null` instead of `undefined`, meaning we end up with a distinction between `null` and `undefined`, looks smelly to me 😬

There are different aspects here.

1. IMO, making a distinction between `null` and `undefined` is not smelly at all, there are two values in JS, not one, "null" meaning there is explicitly nothing, `undefined` meaning the value is not-defined/unknown. The semantic between them is quite clear to me (even if sometimes wrongly used in the code).

2. That's because I prefer using '1'  or '0' in the url than using an empty parameter in the url to mean "true". IMO "mypage?is_user=1&something=434" is clearer than "mypage?is_user=&something=434" to mean "is_user=true" and something=434 meaning false... especially as this whole PR make the assumption (because of the "angular limitation") that "no arg given" means "unchanged"  I saw in the meantime in LTI that you use the empty parameter in LTI. The problem is that query strings are strings and we want boolean so we have to bind an infinite set of strings to two values. So the expected values are simply "0" or "1"... the other (if somebody manually changes the url) are just interpreted as "0". In your case, "mypage?is_user=0&something=434" would mean "is_user=true" which one could find strange as well.

3. We can have inconsistent parameter set "like `watchGroupId=none&watchUser=[whatever]`", right. But I don't see anyway to prevent them, if having a single parameter `watchGroupId` or `watchUserId`, we have the same inconsistency problem if both are given. These are inputs, given only as strings, so as any input they have to be sanitize somehow to prevent impossible behavior as any set of query strings. Of course, I could document a bit more what parameter has the priority over the other one if we think it is useful (as any piece of code could navigate to such url).

> If I understand it correctly, what we want to do is:
> 
>     * Be able to watch a group or a user by user navigation (a button) or url navigation (`watchedGroupId`)
> 
>     * Be able stop watching by user navigation (a button) or url navigation (for dev/admin usage only I suppose ?)

"for dev/admin usage only I suppose ?" -> From what I remember from my code, not really, the change of "watching" is only driven by url (to have a single "entry point" for changing it)... So every change is driven by url change... there is just (because of the "angular limitation") that if params are not given, we consider the previous situation as unchanged.

> 
>     * For a technical reason, be able to make the differentiate a watched group and a watched user
> 
> 
> To me the query parameters handling should be more something like:
> 
>     1. A query param `watchedGroupId=…` (like now) and check existence of param `watchedGroupIsUser` (for instance) instead of allowing any value as falsy
> 
>     2. Two query params `watchedGroupId` and `watchedUserId`, but then we'd need to throw an error if both are supplied
> 

So these are two separate options, right?  Option 2 does not solve any inconsistency as we have to decide what happens in I get "mypage?watchedGroupId=1&watchedUserId=2". This option was considered and I preferred the other one, so coming back to your option 1.

Regarding your option 1, as I explain in (2) above, that is a choice to be done. It is just that it is designed to be "=0" or "=1. As explained above, the main problem I have with your option is that a parameter not set means "unchanged" in group watching, making it inconsistent with how `watchedGroupId` works.
 
> As for the special value `'none'`, it could be replaced by the special case `watchedGroupId=` meaning the a watched group id exists but it's empty

I could do that.

> or another query param `stopWatching`, but it induces to throw an error as well if both query params are supplied.

Yes, of course it was also consider but it involves more confusion.

So, maybe we should do quick chat to pick a choice for all these.

src/app/core/components/top-bar/top-bar.component.html
```diff
@@ -1,9 +1,9 @@
-<ng-container *ngIf="currentMode">
+<ng-container>
```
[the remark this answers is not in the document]
smadbe: fixed

src/app/core/services/group-watching.service.ts
```diff
@@ -0,0 +1,108 @@
...
+    filter(isNotUndefined), // if "unchanged from previous value", ignore (prevent issues with Angular auto-removing query params)
+    startWith(null), // the default is no watching
+    distinctUntilChanged((prev,cur) => prev?.groupId === cur?.groupId && prev?.isUser === cur?.isUser),
+  );
+
+  watchedGroup$ = this.watchedGroupParams.pipe(
+    withLatestFrom(this.cachedGroupInfo),
+    switchMap(([ id, cachedGroup ]) => {
+      if (id === null) return of(null);
+      if (cachedGroup && id.groupId === cachedGroup.route.id) return of(cachedGroup);
```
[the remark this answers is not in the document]
smadbe: I change the parameter name (I remember I saw this when writing it but I cannot remember why I didn't change it some something like "groupParams".

src/app/modules/item/pages/item-progress/item-progress.component.ts
```diff
@@ -21,29 +23,29 @@ export class ItemProgressComponent implements OnChanges {
...
-  readonly session$ = this.sessionService.session$;
+  private itemData$ = new ReplaySubject<ItemData>();
+  selectors$ = combineLatest([ this.itemData$, this.groupWatchingService.watchedGroup$ ]).pipe(
+    map(([ itemData, watchedGroup ]) => {
+      if (!watchedGroup || watchedGroup.route.isUser) {
+        return isATask(itemData.item) ? 'none' : 'withUserProgress';
+      } else {
+        return 'withUserProgress';
+      }
+    }),
```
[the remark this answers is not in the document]
smadbe: Probably a mistake. Fixed.

src/app/modules/item/pages/item-progress/item-progress.component.ts
```diff
@@ -21,29 +23,29 @@ export class ItemProgressComponent implements OnChanges {
...
+  private itemData$ = new ReplaySubject<ItemData>();
+  selectors$ = combineLatest([ this.itemData$, this.groupWatchingService.watchedGroup$ ]).pipe(
+    map(([ itemData, watchedGroup ]) => {
+      if (!watchedGroup || watchedGroup.route.isUser) {
+        return isATask(itemData.item) ? 'none' : 'withUserProgress';
+      } else {
+        return 'withGroupProgress';
+      }
+    }),
+  );
```
ThomasJuster, line 35: Do you have test cases to ensure there's no regression here ?

src/app/shared/helpers/url.ts
```diff
@@ -93,3 +94,18 @@ export function openNewTab(href: string, currentLang?: LanguageConfig): void {
...
+ * Return the value of the given field as a boolean (or null if not set)
+ * Interprets any value as true except "0" (if `boolToQueryParamValue` was used) or "false" (in case the boolean value was used directly
+ * in url)
+ */
+export function boolFromParamMap(paramMap: ParamMap, field: string): boolean|null {
+  const val = paramMap.get(field);
+  if (val === null) return null;
+  if (val === '0' || val === 'false') return false;
+  return true;
+}
```
ThomasJuster, line 107: It restricts usage to `paramMap` only, while we can also use `params` as an object and not a map with `this.activatedRoute.snapshot.queryParams`.

A `queryParamValueToBool(value: string | undefined | null): boolean | null` would be more flexible (or `boolFromQueryParamValue`).

Plus, I'm not convinced by the "false" handling, I think we should stick to `0` and `1`

ThomasJuster (approved):

src/app/modules/item/pages/item-progress/item-progress.component.ts
smadbe: scenarios updated to check this as well

src/app/shared/helpers/url.ts
smadbe: fixed

PR #853, opened 2021-12-22
Feature/on forbidden group display clean error

### LINKED ISSUE #760: On forbidden group: display clean error
## Overview
If a user visit a group he is not allowed to see, he gets
Screenshot 2021-10-15 at 09 57 22
while he could have a more specific error.

## Acceptance criteria

- [ ] On group page, when failing because of 403, display "You are not allowed to view this group page." instead of the current error.
    * Scenario:
       -  use site as user `usr_5p020x2thuyu`
       - [visit group `52767158366271444`](https://dev.algorea.org/en/#/groups/by-id/52767158366271444;path=/details)
       - You should see the new error message.
- [ ] On other errors (such as 500), display the same error as before.

Iloveall:
## Description

Fixes #760

...

## Test cases

- [x] Case 1:
  1. Given I am the demouser
  2. When I go to [this page](https://dev.algorea.org/branch/feature/on-forbidden-group-display-clean-error/en/#/groups/by-id/6884210558506774905;path=/details)
  3. Then I see: You are not allowed to view this group page.

- [x] Case 2:
  1. Given I am the usual user
  2.Then go to [this page](https://dev.algorea.org/branch/feature/on-forbidden-group-display-clean-error/en/#/groups/by-id/6884210558506774905;path=/details)
  3. Turn offline mode in browser network panel and simulate error
  3. Then I see: Error while loading the group info


src/app/modules/group/pages/group-details/group-details.component.html
```diff
@@ -84,10 +84,18 @@
...
-  ></alg-error>
+  >
+    <ng-container i18n>
+      <ng-container *ngIf="state.error.status === 403; else otherError;">
+        You are not allowed to view this group page.
+      </ng-container>
+      <ng-template #otherError>
+        Error while loading the group info
+      </ng-template>
+    </ng-container>
```
ThomasJuster, line 98: i18n attribute should be set per translation, not at top level (which probably makes the top-level `ng-container` obsolete)

src/app/modules/group/pages/group-details/group-details.component.html
```diff
@@ -84,10 +84,18 @@
   <alg-error
```
ThomasJuster, line 84: image
The reload button is a nice touch for "Error while loading group info" but I don't see the point of reloading when you are not allowed to see the content

src/app/modules/group/pages/group-details/group-details.component.html

src/app/modules/group/pages/group-details/group-details.component.html
ThomasJuster: Now it's also removed for "Error while loading group info", isn't there a way to have the refresh button for "Error while loading group info" but not "You are not allowed to view this group page" ?

ThomasJuster (commented):
Cf comment answer

src/app/modules/group/pages/group-details/group-details.component.html
Iloveall: Fixed

src/app/shared/helpers/state.ts
```diff
@@ -2,7 +2,14 @@
...
 export type Ready<T> = Readonly<{ tag: 'ready', isReady: true, data: T, isFetching: false, isError: false, error?: undefined }>;
 export type Fetching<T> = Readonly<{ tag: 'fetching', isReady: false, data?: T, isFetching: true, isError: false, error?: undefined }>;
-export type FetchError = Readonly<{ tag: 'error', isReady: false, data?: undefined, isFetching: false, isError: true, error: Error }>;
+export type FetchError = Readonly<{
+  tag: 'error',
+  isReady: false,
+  data?: undefined,
+  isFetching: false,
+  isError: true,
+  error: Error & { status?: number },
```
ThomasJuster, line 11: To me it's all good, for this part I'd like @smadbe opinion

smadbe: Hum I don't like the idea of polluting the error with possible more-specific errors. It's clear there are improvement to be done (for instance, having a fetch state with a type of data and a type of error).
I think the most pragmatic approach would be to keep the previous version of the `state.ts` file and to use in template `$any(state.error).status !== 403` so that `state.error` is converted to `any` to try to fetch more specific error properties.

smadbe (changes requested):

src/app/shared/helpers/state.ts
Iloveall: Fixed

smadbe (approved):
👍 ok for me

ThomasJuster (approved):

PR #854, opened 2022-01-03
chore: upgrade angular

### LINKED ISSUE #845: Upgrade to Angular 13
Just upgrade to Angular 13. (see https://blog.angular.io/angular-v13-is-now-available-cce66f7bc296) 

Note: I think you need to enable build cache manually (maybe the new dir should be cached in the CI as well ?)

ThomasJuster:
## Description

Fixes #845 

:one: Upgrading `@angular-eslint/schematics & co` to v13 required to upgrade `@typescript-eslint/plugin & parser` to v5 (from v4), which required to upgrade `eslint-plugin-rxjs` to v4 (from v3)

:two: The eslint rxjs rule [suffix subjects](https://github.com/cartant/eslint-plugin-rxjs/blob/main/docs/rules/suffix-subjects.md) was broken, upgrading the lib fixed the rule. But since it was not applied, I turned it off. If we want to reapply it, I suggest we make it an issue. However, I doubt the relevance of this rule since it does not apply to observables but only to subjects. And when browsing code it looks to me that we want `$` suffix on observables all the time but not necessarily on subjects (when they are private properties for instance).

:three: There is a big collision between eslint TS rules and rxjs new types for `mapTo` and `switchMapTo` operators: They were typed before as `OperatorFunction<Input, Output>`, now they are − logically − typed as `OperatorFunction<any, Output>`, inducing a lot of `@typescript-eslint/no-unsafe-argument` for almost every `(switch)mapTo` operator. I took the initiative of using `(switch)map(() => …)` instead of deactivating the eslint rule, which is debatable.

:four: I had to install `karma-coverage` package since it is required when using `ng test --code-coverage` option, I don't know why this error wasn't triggered before.

## Test cases

- [ ] Case 1:
  1. Given I am the any user
  2. When I visit [the home page](https://dev.algorea.org/branch/chore/upgrade-angular/en/#/)
  3. Then I see the app is loaded normally


src/app/modules/group/pages/user/user.component.ts
```diff
@@ -24,7 +24,7 @@ export class UserComponent implements OnInit, OnDestroy {
 
   private refresh$ = new Subject<void>();
   readonly state$ = this.route.params.pipe(
-    switchMap(({ id }) => this.getUserService.getForId(id)),
+    switchMap(({ id }) => this.getUserService.getForId(id as string)),
```
smadbe, line 27: You force-cast to string something which is not necessarily a string!

src/app/modules/item/pages/item-details/item-details.component.ts
```diff
@@ -52,7 +52,7 @@ export class ItemDetailsComponent implements OnDestroy, BeforeUnloadComponent {
   );
 
   readonly formerAnswerError$ = this.formerAnswer$.pipe(
-    catchError(error => of(errorIsHTTPForbidden(error) ? loadForbiddenAnswerError : new Error(error))),
+    catchError((error: string) => of(errorIsHTTPForbidden(error) ? loadForbiddenAnswerError : new Error(error))),
```
smadbe, line 55: How do you know for sure it is a string, I don't think you can do this assumption.

src/app/modules/item/pages/item-by-id/item-by-id.component.ts
```diff
@@ -195,7 +195,7 @@ export class ItemByIdComponent implements OnDestroy {
       })
     ).subscribe({
       next: itemRoute => this.itemRouter.navigateTo(itemRoute, { navExtras: { replaceUrl: true } }),
-      error: err => {
+      error: (err: Error) => {
```
smadbe, line 198: How do you know for sure it is a `Error`, I don't think you can do this assumption.
Isn't related to https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#defaulting-to-the-unknown-type-in-catch-variables---useunknownincatchvariables ?

src/app/modules/item/services/item-task-answer.service.ts
```diff
@@ -73,40 +73,40 @@ export class ItemTaskAnswerService implements OnDestroy {
...
-    this.saveError$.pipe(mapTo({ success: false })),
+    this.saved$.pipe(map(() => ({ success: true }))),
+    this.saveError$.pipe(map(() => ({ success: false }))),
   );
 
   private subscriptions = [
-    this.initializedTaskAnswer$.subscribe({ error: err => this.errorSubject.next(err) }),
-    this.initializedTaskState$.subscribe({ error: err => this.errorSubject.next(err) }),
+    this.initializedTaskAnswer$.subscribe({ error: (err: Error) => this.errorSubject.next(err) }),
+    this.initializedTaskState$.subscribe({ error: (err: Error) => this.errorSubject.next(err) }),
```
smadbe, line 109: same as above

src/app/modules/item/services/item-task-views.service.ts
```diff
@@ -31,7 +31,7 @@ export class ItemTaskViewsService implements OnDestroy {
   );
 
   private subscriptions = [
-    this.showViews$.subscribe({ error: err => this.errorSubject.next(err) }),
+    this.showViews$.subscribe({ error: (err: Error) => this.errorSubject.next(err) }),
```
smadbe, line 34: same as above

src/app/shared/operators/state.ts
```diff
@@ -55,6 +55,6 @@ export function mapStateData<T, U>(dataMapper: (data: T) => U): OperatorFunction
  */
 export function mapErrorToState<T>(): OperatorFunction<FetchState<T>,FetchState<T>> {
   return pipe(
-    catchError(e => of(errorState(e))),
+    catchError((e: Error) => of(errorState(e))),
```
smadbe, line 58: same as above

smadbe (changes requested):

ThomasJuster:
The merge might have to be postponed until [this issue regarding @angular/localize](https://github.com/angular/angular-cli/issues/22435) is solved

src/app/modules/group/pages/user/user.component.ts
ThomasJuster: fixed
Since it comes from route params, id is necessarily a string … or undefined. But typed as any :shrug: by angular

src/app/modules/item/pages/item-details/item-details.component.ts
ThomasJuster: fixed

src/app/modules/item/pages/item-by-id/item-by-id.component.ts
ThomasJuster: fixed

src/app/modules/item/services/item-task-answer.service.ts
ThomasJuster: fixed

src/app/modules/item/services/item-task-views.service.ts
ThomasJuster: fixed

src/app/shared/operators/state.ts
ThomasJuster: fixed

smadbe (approved):
👍 

> The merge might have to be postponed until [this issue regarding @angular/localize](https://github.com/angular/angular-cli/issues/22435) is solved

So you get this problem in practice as well? Looks annoying.

ThomasJuster:
> > The merge might have to be postponed until this issue regarding @angular/localize is solved

> So you get this problem in practice as well? Looks annoying.

Yep, it occurs at each hot reload :shrug:

PR #858, opened 2022-01-05
implement LTI login

### LINKED ISSUE #839: (LTI) LTI login
## Overview

In order to integrate with LTI as a provider, we need to be able to authenticate the user on arrival.

## Acceptance criteria

- [ ] support for "login_id" parameter (query param?) 

On landing:
- [ ] check whether the provided login_id (>< login and <> user_id) matches the current user
- [ ] if not, do a silent auth to get the cookie
- [ ] make sure it cannot loop forever

ThomasJuster:
## Description

Fixes #839 

Two main cases: when (1) login module user id matches current logged user or not (2).
Case (1) adds a step by calling backend service [check login id](https://france-ioi.github.io/algorea-devdoc/api/#operation/loginIDCheck).
For case (2) when user id does not match current user group id:
1. Store `redirect_to_sub_path: /lti/…` in session storage, this value is read and removed at app startup (ie when coming back after a login or refreshing). The LTI component will reset it to `/lti/…&is_redirection` right afterwards to stay in LTI mode even if the user refreshes the page
2. If `is_redirection` query param is set, we won't check again that the user id matches to avoid infinite loops. If the user id still mismatches the current user's group id, display an error.

## Test cases

:bulb: Be sure to clear the session storage "redirect_to_sub_path" between tests !
:warning: Tests work only for production app or localhost app because of the authorized login auth redirection urls. So you have to check out this branch to run the tests :confused: 

- [ ] Case 1: unlogged then already logged cases
  1. Given I am logged into moodle algorea as the lti demo user
  2. When I navigate to [this moodle item](http://moodletest.algorea.org/mod/lti/view.php?id=8)
  3. Then a pop-up is opened with url http://moodletest.algorea.org/mod/lti/launch.php?id=8 (useful to restart test or open a tab manually from here, clear the session storage if needed and throttle the network to see it in detail)
  4. Then there's a redirection to [algorea frontend](http://localhost:4200/en/#/lti/34689573454313988?user_id=100465069)
  5. Then there's a redirection to the login module re-redirecting to algorea frontend with proper login
  6. Then there's an internal app navigation to lti again, and finally we see the activity "Les boucles bornées ou à compteur"
  7. And clear `redirect_to_sub_path` in session storage and copy-paste [step 4 url](http://localhost:4200/en/#/lti/34689573454313988?user_id=100465069) (already logged in case)
  8. Then there's no redirection and the activity is launched

- [ ] Case 2: refresh to simulate invalid user id after login
  1. Given I am the lti demo user
  2. When I go to [home page](http://localhost:4200/#/)
  3. And I open devtools and add in session storage `redirect_to_sub_path: /lti/1625159049301502151?user_id=42&is_redirection=`
  4. And I refresh the page (still on the home page)
  5. Then I should see a login failed error

- [ ] Case 3: invalid user id in url -> force user to re-log in
  1. Given I am the any user
  2. When I go to [home page](http://localhost:4200/#/) (creates a temp user if none)
  3. And I go to [this chapter](http://localhost:4200/#/lti/1625159049301502151?user_id=43)
  4. Then I am redirected to the login module

- [ ] Case 4: no `user_id` (`loginId`) in url
  1. Given I am any user
  2. When I navigate to [this chapter in lti](http://localhost:4200/#/lti/1625159049301502151)
  3. Then I should see an error "Something wrong occurred in the login process …".

### DIFF
--- a/src/app/modules/lti/pages/lti/lti.component.ts
+++ b/src/app/modules/lti/pages/lti/lti.component.ts
@@ -1,6 +1,6 @@
 import { Component, OnDestroy } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
-import { combineLatest, EMPTY, forkJoin, Observable } from 'rxjs';
+import { combineLatest, EMPTY, forkJoin, Observable, of } from 'rxjs';
 import { catchError, filter, map, shareReplay, switchMap } from 'rxjs/operators';
 import { ActivityNavTreeService } from 'src/app/core/services/navigation/item-nav-tree.service';
 import { GetItemChildrenService, ItemChild } from 'src/app/modules/item/http-services/get-item-children.service';
@@ -8,6 +8,7 @@ import { GetItemPathService } from 'src/app/modules/item/http-services/get-item-
 import { errorIsHTTPForbidden } from 'src/app/shared/helpers/errors';
 import { isNotNull } from 'src/app/shared/helpers/null-undefined-predicates';
 import { removeSubPathRedirectionAtInit, setRedirectToSubPathAtInit } from 'src/app/shared/helpers/redirect-to-sub-path-at-init';
+import { CheckLoginService } from 'src/app/shared/http-services/check-login.service';
 import { ResultActionsService } from 'src/app/shared/http-services/result-actions.service';
 import { mapToFetchState, readyData } from 'src/app/shared/operators/state';
 import { fullItemRoute } from 'src/app/shared/routing/item-route';
@@ -26,7 +27,7 @@ const noChildError = new Error(LTIError.NoChild);
 const loginError = new Error(LTIError.LoginError);
 
 const isRedirectionParam = 'is_redirection';
-const loginIdParam = 'login_id';
+const loginIdParam = 'user_id';
 
 @Component({
   selector: 'alg-lti',
@@ -36,7 +37,11 @@ const loginIdParam = 'login_id';
 export class LTIComponent implements OnDestroy {
 
   private loginId$ = this.activatedRoute.queryParamMap.pipe(
-    map(queryParams => queryParams.get(loginIdParam)),
+    map(queryParams => {
+      const loginId = Number(queryParams.get(loginIdParam) ?? NaN);
+      if (Number.isNaN(loginId)) throw new Error('login id must be a number');
+      return loginId;
+    }),
   );
 
   private isRedirection$ = this.activatedRoute.queryParamMap.pipe(
@@ -51,8 +56,11 @@ export class LTIComponent implements OnDestroy {
     }),
   );
 
-  private isLoggedIn$ = combineLatest([ this.userSession.userProfile$, this.loginId$ ]).pipe(
-    map(([ profile, loginId ]) => !!loginId && profile.login === loginId),
+  private isLoggedIn$ = combineLatest([ this.loginId$, this.userSession.userProfile$ ]).pipe(
+    switchMap(([ loginId, profile ]) => {
+      if (profile.tempUser) return of(false);
+      return this.checkLoginService.check(loginId);
+    }),
   );
 
   readonly navigationData$ = combineLatest([ this.isLoggedIn$, this.isRedirection$ ]).pipe(
@@ -104,6 +112,7 @@ export class LTIComponent implements OnDestroy {
     private activatedRoute: ActivatedRoute,
     private itemRouter: ItemRouter,
     private userSession: UserSessionService,
+    private checkLoginService: CheckLoginService,
     private layoutService: LayoutService,
     private activityNavTreeService: ActivityNavTreeService,
     private getItemPathService: GetItemPathService,
--- a/src/app/modules/lti/pages/lti/lti.component.ts
+++ b/src/app/modules/lti/pages/lti/lti.component.ts
@@ -7,7 +7,7 @@ import { GetItemChildrenService, ItemChild } from 'src/app/modules/item/http-ser
 import { GetItemPathService } from 'src/app/modules/item/http-services/get-item-path.service';
 import { errorIsHTTPForbidden } from 'src/app/shared/helpers/errors';
 import { isNotNull } from 'src/app/shared/helpers/null-undefined-predicates';
-import { removeSubPathRedirectionAtInit, setRedirectToSubPathAtInit } from 'src/app/shared/helpers/redirect-to-sub-path-at-init';
+import { setRedirectToSubPathAtInit } from 'src/app/shared/helpers/redirect-to-sub-path-at-init';
 import { CheckLoginService } from 'src/app/shared/http-services/check-login.service';
 import { ResultActionsService } from 'src/app/shared/http-services/result-actions.service';
 import { mapToFetchState, readyData } from 'src/app/shared/operators/state';
@@ -122,16 +122,6 @@ export class LTIComponent implements OnDestroy {
 
   ngOnDestroy(): void {
     this.subscriptions.forEach(subscription => subscription.unsubscribe());
-  }
-
-  restartProcess(): void {
-    combineLatest([
-      this.contentId$,
-      this.loginId$.pipe(filter(isNotNull)),
-    ]).subscribe(([ contentId, loginId ]) => {
-      removeSubPathRedirectionAtInit();
-      window.location.href = `./#/lti/${contentId}?${loginIdParam}=${loginId}`;
-    });
   }
 
   private getNavigationData(itemId: string): Observable<{ firstChild: ItemChild, path: string[], attemptId: string }> {
--- a/src/app/modules/lti/pages/lti/lti.component.ts
+++ b/src/app/modules/lti/pages/lti/lti.component.ts
@@ -1,7 +1,7 @@
 import { Component, OnDestroy } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { combineLatest, EMPTY, forkJoin, Observable } from 'rxjs';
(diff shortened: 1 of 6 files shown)


src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -27,41 +35,76 @@ const noChildError = new Error(LTIError.NoChild);
...
       if (!contentId) throw new Error('unexpected: contentId should be defined');
       return contentId;
     }),
   );
-  readonly navigationData$ = this.contentId$.pipe(
-    switchMap(contentId => this.getNavigationData(contentId).pipe(mapToFetchState())),
+
+  private isLoggedIn$ = combineLatest([ this.userSession.userProfile$, this.userId$ ]).pipe(
+    map(([ profile, userId ]) => !!userId && profile.groupId === userId),
+    take(1),
```
smadbe, line 56: So if it is only to keep the first value, is it relevant to use Rx `userId$` and `isRedirection$` for that ? Could you just use the activated route snapshot to compute it once?
Or you assume parameters may change (they may probably)... but so the `take(1)` makes the component inconsistent in this case.

src/locale/messages.fr.xlf
```diff
@@ -672,6 +672,12 @@
           <context context-type="sourcefile">src/app/modules/lti/pages/lti/lti.component.html</context>
           <context context-type="linenumber">13</context>
         </context-group>
+      </trans-unit><trans-unit id="2205921893315386228" datatype="html">
+        <source> Something wrong occurred during the login process, please <x id="START_TAG_SPAN" ctype="x-span" equiv-text="&lt;span class=&quot;alg-link&quot; (click)=&quot;restartProcess()&quot;>"/>retry<x id="CLOSE_TAG_SPAN" ctype="x-span" equiv-text="&lt;/span>"/>. <x id="LINE_BREAK" ctype="lb" equiv-text="&lt;br>"/> If the problem persists, please contact us. </source><target state="new"> Something wrong occurred during the login process, please <x id="START_TAG_SPAN" ctype="x-span" equiv-text="&lt;span class=&quot;alg-link&quot; (click)=&quot;restartProcess()&quot;>"/>retry<x id="CLOSE_TAG_SPAN" ctype="x-span" equiv-text="&lt;/span>"/>. <x id="LINE_BREAK" ctype="lb" equiv-text="&lt;br>"/> If the problem persists, please contact us. </target>
```
smadbe, line 676: the link, click, ... is in the translated file? looks suspicious

smadbe (changes requested):

src/locale/messages.fr.xlf
ThomasJuster: weird, I'll regenerate translations

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: I removed the `take` operator
fixed

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -27,41 +35,76 @@ const noChildError = new Error(LTIError.NoChild);
...
       const contentId = params.get('contentId');
       if (!contentId) throw new Error('unexpected: contentId should be defined');
       return contentId;
     }),
   );
-  readonly navigationData$ = this.contentId$.pipe(
-    switchMap(contentId => this.getNavigationData(contentId).pipe(mapToFetchState())),
+
+  private isLoggedIn$ = combineLatest([ this.userSession.userProfile$, this.userId$ ]).pipe(
+    map(([ profile, userId ]) => !!userId && profile.groupId === userId),
```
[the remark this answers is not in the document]
ThomasJuster: The query param `login_id` should match `user.login` then ?
https://france-ioi.github.io/algorea-devdoc/api/#operation/userData
If so, fixed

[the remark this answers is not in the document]
smadbe: No, `login_id` https://franceioi-algorea.s3.eu-west-3.amazonaws.com/dbdoc/tables/users.html 🙄
... ok I check how we can get it.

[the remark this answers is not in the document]
smadbe: You should use `https://france-ioi.github.io/algorea-devdoc/api/#operation/loginIDCheck`

src/locale/messages.fr.xlf
smadbe: But these are just other developer errors 😁. It is because you have link and html entities in your `i18n` block. 
Actually, I am not 100% sure we are not supposed to do that but that looks better to me not to have them. If you fix the link in the code, translations should be have a broken version. Read a bit about i18n to see what's the best way to proceed.

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -26,7 +26,7 @@ const noChildError = new Error(LTIError.NoChild);
 const loginError = new Error(LTIError.LoginError);
 
 const isRedirectionParam = 'is_redirection';
-const userIdParam = 'user_id';
+const loginIdParam = 'login_id';
```
smadbe, line 29: No, the parameter given by the login module is still `user_id`!  It is just that "user_id" is the "loginIdParam" for the frontend.

smadbe (changes requested):
This remark + the 2 answers to previous comments.

src/locale/messages.fr.xlf
ThomasJuster: To me it's a pretty common use case to translate text with nested tags like links for instance, and if you fix the link in the code then the translations are re-generated, only the placeholder will change. It's a normal way to interpolate translations, in the example/link I provided upper, variable interpolation works exactly the same way:
```xml
<x id="INTERPOLATION" equiv-text="{{ grantedPermission.sourceGroup.name }}"/>
```
Here if we change the variable name `grantedPermission` then as you imagine it we would have a broken version as well but it's not the case.

Although it is mentioned absolutely nowhere in the Angular documentation regarding i18n (:broken_heart:), I could find issues talking something else but giving examples of template translations with nested tags:
- https://github.com/angular/angular/issues/18302
- https://github.com/mgechev/codelyzer/issues/442
- https://phrase.com/blog/posts/localize-angular-applications-with-built-in-i18n-module/ (check out the "Working with Select Input" section)

Anyway, if we don't do it with nested tags like this …
```html
<span i18n>
  Hi there, check out this awesome <a [routerLink]="…">blog post</a> …
</span>
```
… there's no way not to split the translation parts

What do you suggest to fix the link in code ?

src/app/modules/lti/pages/lti/lti.component.ts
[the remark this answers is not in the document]
ThomasJuster: fixed

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: fixed

src/locale/messages.fr.xlf
smadbe: Ok so `<x...` tab is done for that. Let's keep it like this for now at least.

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -36,7 +37,11 @@ const loginIdParam = 'login_id';
 export class LTIComponent implements OnDestroy {
 
   private loginId$ = this.activatedRoute.queryParamMap.pipe(
-    map(queryParams => queryParams.get(loginIdParam)),
+    map(queryParams => {
+      const loginId = Number(queryParams.get(loginIdParam) ?? NaN);
+      if (Number.isNaN(loginId)) throw new Error('login id must be a number');
+      return loginId;
+    }),
```
smadbe, line 44: Why converting it to a number? Because the service takes a number as an input? 
I am afraid this `integer <int64>` for input is just equivalent to `string <int64>` as query string are strings anyway... and the reason we use strings for ids is that JS does not have enough precision on numbers to support int64. As a consequence, by converting the string to number, you may loose precision while doing that, and so modify the id. :-/ 

So, tl;dr: use strings for ids in your service and here.

smadbe (changes requested):

src/app/modules/lti/pages/lti/lti.component.ts
ThomasJuster: fixed

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -27,41 +36,74 @@ const noChildError = new Error(LTIError.NoChild);
...
       const contentId = params.get('contentId');
       if (!contentId) throw new Error('unexpected: contentId should be defined');
       return contentId;
     }),
   );
-  readonly navigationData$ = this.contentId$.pipe(
-    switchMap(contentId => this.getNavigationData(contentId).pipe(mapToFetchState())),
+
+  private isLoggedIn$ = this.loginId$.pipe(
+    switchMap(loginId => (loginId ? this.checkLoginService.check(loginId) : of(false))),
```
smadbe, line 54: Do you support the case where the loginId is not given? (in another way than just sending the user to the login-module.. which should not fix the problem.

ThomasJuster: Fixed by displaying a login error. Is it ok ?
I figured that opening the app as LTI provider but no with no loginId does not make sense, but I'm not entirely sure about that.

ThomasJuster: I also added test case 4 for that

smadbe (approved):
👍

PR #868, opened 2022-01-13
fix item blocking navigation

### LINKED ISSUE #866: Nav menu blocked after visiting item in error
**Describe the bug**
The left nav menu is blocked (clicking does not navigate anymore) from time to time (when visiting specific items?)

**To Reproduce**
1. As the usual user
1. Go to https://dev.algorea.org/en/#/activities/by-id/126952315730161957;path=39530140456452546;attempId=0/details
2. Uncollapse the left menu
3. If I try to navigate using the left menu, nothing happen

**Expected behavior**

**Screenshots**

https://user-images.githubusercontent.com/1053150/149285690-ed903b70-5efa-44fb-8448-226669ed9af8.mov

ThomasJuster:
... by returning empty observable to before unload hook

## Description

Fixes #866 

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/fix/item-blocking-nav/en/#/activities/by-id/126952315730161957;path=39530140456452546;attempId=0/details)
  3. And I navigate elsewhere (using top right nav, left nav, etc.)
  4. Then I see navigation is working.

### DIFF
--- a/src/app/modules/item/services/item-task-answer.service.ts
+++ b/src/app/modules/item/services/item-task-answer.service.ts
@@ -198,6 +198,7 @@ export class ItemTaskAnswerService implements OnDestroy {
           startWith({ saving: true }),
         );
       }),
+      defaultIfEmpty({ saving: false }), // when a timeout is caught, the observable is empty
       shareReplay(1),
     );
   }


smadbe (approved):
I don' fully understand how it blocks navigation and if there would be a way to avoid this kind of issues in the future, but that looks like it is working!

ThomasJuster:
> I don' fully understand how it blocks navigation and if there would be a way to avoid this kind of issues in the future, but that looks like it is working!

The `CanDeactivate` guard must be returned an `Observable<boolean>`, but here in one specific use-case, the returned observable was `EMPTY`, meaning it would wait indefinitely. I'm not sure there are any ways to avoid this kind of issues since it's an edge case here.

PR #873, opened 2022-01-21
Adds create manager component

### LINKED ISSUE #869: Add a manager
## Overview

The current group manager page allows to remove and edit an existing manager but not to add one. Let managers add other managers.

## Acceptance criteria
- [ ] On group manager page, if the current user can manage "Memberships and group" of the group (and only in this case), add a blue box (as shown below)
Screenshot 2022-01-18 at 10 26 14
- [x] The blue has, as a title, "Add a manager" 
- [x] The placeholder is "Login of a user"
- [x] The right button is "Add" with a "+" icon in front
- [x] The right button is greyed out as long as the field is empty
- [ ] On click or 'add', the text field is greyed out, the button is "in loading" (you can suggest something more consistent with the other similar cases if you think it is better)
[https://www.primefaces.org/primeng/showcase/#/button](Screenshot 2022-01-18 at 10 31 38)
- [ ] if the enter login is already among your current manager: red toast "This user is already a manager of this group."
- [x] If not, find the entered login using https://france-ioi.github.io/algorea-devdoc/api/#operation/userViewByLogin
  - [x] on failure: red toast "The login you entered does not exist or is not visible to you."
- [ ] On success, create a manager using https://france-ioi.github.io/algorea-devdoc/api/#operation/groupManagerCreate and the minimum permissions.
- [x] on failure: red toast "Unable to add this manager."
- [x] on success: green toast "Manager added!" and refresh the manager list

Iloveall:
## Description

Fixes #869

...

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/add-manager/en/#/groups/by-id/609422240375608067;path=/details/managers)
  3. And I scroll down to add manager section
  4. Add usr_4790cnyt9hiq login
  5. Then I can see usr_4790cnyt9hiq in list of managers

- [x] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/add-manager/en/#/groups/by-id/609422240375608067;path=/details/managers)
  3. And I scroll down to add manager section
  4. Add arbonenfant login
  5. Then I can see error message "This user is already a manager of this group."

- [x] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/add-manager/en/#/groups/by-id/609422240375608067;path=/details/managers)
  3. And I scroll down to add manager section
  4. Add unknownuser login
  5. Then I can see error message "The login you entered does not exist or is not visible to you."

### DIFF
--- a/src/app/core/http-services/get-user-by-login.service.ts
+++ b/src/app/core/http-services/get-user-by-login.service.ts
@@ -1,0 +1,20 @@
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { appConfig } from '../../shared/helpers/config';
+import { decodeSnakeCase } from '../../shared/operators/decode';
+import { HttpClient } from '@angular/common/http';
+import { User, userDecoder } from '../../modules/group/http-services/get-user.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class GetUserByLoginService {
+  constructor(private http: HttpClient) {
+  }
+
+  get(login: string): Observable<User> {
+    return this.http
+      .get<unknown>(`${appConfig.apiUrl}/users/by-login/${login}`)
+      .pipe(decodeSnakeCase(userDecoder));
+  }
+}
--- a/src/app/modules/group/components/group-manager-add/group-manager-add.component.html
+++ b/src/app/modules/group/components/group-manager-add/group-manager-add.component.html
@@ -1,10 +1,8 @@
 <alg-section-paragraph icon="fa fa-sign-in-alt" i18n-label label="Add a manager">
   <div class="input-wrapper">
     <input
-      #input
       type="text"
       i18n-placeholder placeholder="Login of a user"
-      (keyup)="0"
       [(ngModel)]="login"
       [disabled]="state === 'loading'"
     >
@@ -15,7 +13,7 @@
       icon="fa fa-plus"
       i18n-label label="Add"
       class="p-button-rounded"
-      [disabled]="input.value === '' || state === 'loading'"
+      [disabled]="login === '' || state === 'loading'"
       (click)="onClick()"
     ></alg-button>
   </div>
--- a/src/app/modules/group/components/group-manager-add/group-manager-add.component.html
+++ b/src/app/modules/group/components/group-manager-add/group-manager-add.component.html
@@ -1,0 +1,22 @@
+<alg-section-paragraph icon="fa fa-sign-in-alt" i18n-label label="Add a manager">
+  <div class="input-wrapper">
+    <input
+      #input
+      type="text"
+      i18n-placeholder placeholder="Login of a user"
+      (keyup)="0"
+      [(ngModel)]="login"
+      [disabled]="state === 'loading'"
+    >
+    <span class="icon">
+      <i class="fa fa-font"></i>
+    </span>
+    <alg-button
+      icon="fa fa-plus"
+      i18n-label label="Add"
+      class="p-button-rounded"
+      [disabled]="input.value === '' || state === 'loading'"
+      (click)="onClick()"
+    ></alg-button>
+  </div>
+</alg-section-paragraph>
--- a/src/app/modules/group/components/group-manager-add/group-manager-add.component.ts
+++ b/src/app/modules/group/components/group-manager-add/group-manager-add.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input } from '@angular/core';
+import { Component, EventEmitter, Input, Output } from '@angular/core';
 import { GroupData } from '../../services/group-datasource.service';
 import { catchError, switchMap } from 'rxjs/operators';
 import { EMPTY } from 'rxjs';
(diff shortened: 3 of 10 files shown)


src/app/core/http-services/get-user-by-login.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@angular/core';
```
ThomasJuster, line 1: On click, the button is properly greyed out but there's no loading indicator, have you tried providing the input `[loading]` as specified in [primeng documentation](https://www.primefaces.org/primeng/showcase/#/button) ?

ThomasJuster, line 1: The spec says to create the manager with minimum permissions but the payload is:
```json
{
    "can_manage": "none",
    "can_grant_group_access": true,
    "can_watch_members": true
}
```
I think `can_manage` is correct and the two other props should be false, can you check ?

ThomasJuster (commented):
:+1:

src/app/core/http-services/get-user-by-login.service.ts
Iloveall: This version of primeng doesn't have loading feature for button. I've got task from Damien for update primeng but in specific branch

src/app/modules/group/components/group-manager-add/group-manager-add.component.ts
```diff
@@ -0,0 +1,68 @@
...
+        this.state = 'ready';
+        this.actionFeedbackService.success($localize`Manager added!`);
+        this.login = '';
+        this.added.emit();
+      },
+      error: () => {
+        this.state = 'error';
+        this.actionFeedbackService.error($localize`Unable to add this manager.`);
+      },
+    });
```
ThomasJuster, line 66: To me error handling is ok but I'd like the opinion of @smadbe

src/app/modules/group/components/group-manager-add/group-manager-add.component.html
```diff
@@ -0,0 +1,22 @@
+<alg-section-paragraph icon="fa fa-sign-in-alt" i18n-label label="Add a manager">
+  <div class="input-wrapper">
+    <input
+      #input
+      type="text"
+      i18n-placeholder placeholder="Login of a user"
+      (keyup)="0"
```
ThomasJuster, line 7: Is this needed ? Since there is an `ngModel` two-way binding, I suppose the angular will be run the change detection and the template will be updated.

src/app/modules/group/components/group-manager-add/group-manager-add.component.ts
smadbe: No side effects in Rx operators! So L50-51 are not ok! In case of error in `get`, the switchMap will not be called anyway.
And it is not ok to consider all kind of errors of `get` as "The login you entered does not exist or is not visible to you." ... it may be a server error (5xx). 
To adapt the error message to the service in error, either you find a way to find this info in the error metadata (?) or you write a `catchError` for each call which emit a customized error. 
Maybe @ThomasJuster has better idea?

ThomasJuster: Throw specific pre-declared errors and handle them in the `subscribe({ error: () => {…} })`
ie:
```ts
const loginNotFoundError = new Error('login_not_found')
this.getUserByLoginService.get(this.login).pipe(
  catchError((error) => {
    if (errorIsHTTPNotFound(error)) throw loginNotFoundError
    throw error
  }),
  switchMap(() => this.groupCreateManagerService.create(…)),
  catchError((error) => {
     if (errorIsHTTPForbidden(error)) throw 
  }),
)
```

Iloveall: Fixed before I've seen Thomas comment, please take a look

src/app/modules/group/components/group-manager-add/group-manager-add.component.ts
```diff
@@ -59,9 +54,16 @@ export class GroupManagerAddComponent {
...
         this.state = 'error';
-        this.actionFeedbackService.error($localize`Unable to add this manager.`);
+
+        if (errorIsHTTPNotFound(error)) {
+          this.actionFeedbackService.error($localize`The login you entered does not exist or is not visible to you.`);
+        } else if (errorIsHTTPInternalServer(error)) {
+          this.actionFeedbackService.unexpectedError();
+        } else {
+          this.actionFeedbackService.error($localize`Unable to add this manager.`);
+        }
```
ThomasJuster, line 66: I think an `unexpectedError` might happen in other cases that http internal server errors like simple JS errors.
To me here the error mapping is more like:
1. Http Not Found: The login does not exist
2. Http forbidden: Unable to add this manager
3. Any other case: Unexpected error

Iloveall: Fixed

smadbe (approved):
Reviewed only the UI: looks good 👍

PR #879, opened 2022-01-28
Fix/unhandled breacrumbs error

ThomasJuster:
## Description

Fixes #...

...

## Test cases

- [ ] Case 1:
  1. Given I am a brand new user
  2. When I go to [this activity](https://dev.algorea.org/branch/fix/unhandled-breacrumbs-error/en/#/activities/by-id/773396873861865944;path=4702,1625159049301502151;parentAttempId=0/details)
  3. Then I see the activity is loaded

### DIFF
--- a/src/app/shared/helpers/errors.ts
+++ b/src/app/shared/helpers/errors.ts
@@ -6,7 +6,7 @@ interface WithErrorTag { errorTag: string }
  * Add a tag to an error object so it can be recognized later (using the errorHasTag function)
  */
 export function tagError<T>(error: T, tag: string): T & WithErrorTag {
-  return { ...error, errorTag: tag };
+  return Object.assign(error, { errorTag: tag });
 }
 
 export function errorHasTag(error: any, tag: string): boolean {
--- a/src/app/shared/helpers/state.ts
+++ b/src/app/shared/helpers/state.ts
@@ -1,8 +1,16 @@
+import { HttpErrorResponse } from '@angular/common/http';
 
 export type FetchState<T> = Ready<T>|Fetching<T>|FetchError;
 export type Ready<T> = Readonly<{ tag: 'ready', isReady: true, data: T, isFetching: false, isError: false, error?: undefined }>;
 export type Fetching<T> = Readonly<{ tag: 'fetching', isReady: false, data?: T, isFetching: true, isError: false, error?: undefined }>;
-export type FetchError = Readonly<{ tag: 'error', isReady: false, data?: undefined, isFetching: false, isError: true, error: Error }>;
+export type FetchError = Readonly<{
+  tag: 'error',
+  isReady: false,
+  data?: undefined,
+  isFetching: false,
+  isError: true,
+  error: Error|HttpErrorResponse,
+}>;
 
 export function readyState<T>(data: T): Ready<T> {
   return { tag: 'ready', isReady: true, data: data, isFetching: false, isError: false };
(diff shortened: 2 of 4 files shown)


src/app/shared/helpers/errors.ts
```diff
@@ -6,7 +6,7 @@ interface WithErrorTag { errorTag: string }
  * Add a tag to an error object so it can be recognized later (using the errorHasTag function)
  */
 export function tagError<T>(error: T, tag: string): T & WithErrorTag {
-  return { ...error, errorTag: tag };
+  return Object.assign(error, { errorTag: tag });
```
smadbe, line 9: that's not equivalent?

ThomasJuster: Nope, `Object.assign` preserves instances while the spread operator converts any form of data to a plain JS object

src/app/shared/helpers/state.ts
```diff
@@ -1,8 +1,16 @@
...
 export type Ready<T> = Readonly<{ tag: 'ready', isReady: true, data: T, isFetching: false, isError: false, error?: undefined }>;
 export type Fetching<T> = Readonly<{ tag: 'fetching', isReady: false, data?: T, isFetching: true, isError: false, error?: undefined }>;
-export type FetchError = Readonly<{ tag: 'error', isReady: false, data?: undefined, isFetching: false, isError: true, error: Error }>;
+export type FetchError = Readonly<{
+  tag: 'error',
+  isReady: false,
+  data?: undefined,
+  isFetching: false,
+  isError: true,
+  error: Error|HttpErrorResponse,
```
smadbe, line 12: You do the same change as Alex a few weeks ago.. let me find my answer...
> Hum I don't like the idea of polluting the error with possible more-specific errors. It's clear there are improvement to be done (for instance, having a fetch state with a type of data and a type of error).
I think the most pragmatic approach would be to keep the previous version of the state.ts file and to use in template $any(state.error).status !== 403 so that state.error is converted to any to try to fetch more specific error properties.

Ok we are not exactly in the same but still, changing "my number is a positive integer" by "my number is a positive integer or a even positive integer" does not not feel right in terms of inheritance.

src/app/shared/helpers/errors.ts
smadbe: Ok, makes sense. This (my initial) trick looks a bit a dirty actually. If we are extending an object, we should use inheritance.

src/app/shared/helpers/state.ts
ThomasJuster: Problem is that HttpErrorResponse does not inherit the built-in Error class…

smadbe: 1. `Error` of es5 used here seems to be a interface. (so also the remark on assign vs spread operator looks strange)
2. `class HttpErrorResponse extends HttpResponseBase implements Error `, so that looks like there are not unrelated.

ThomasJuster: fixed

src/app/shared/operators/state.ts
```diff
@@ -55,6 +56,6 @@ export function mapStateData<T, U>(dataMapper: (data: T) => U): OperatorFunction
  */
 export function mapErrorToState<T>(): OperatorFunction<FetchState<T>,FetchState<T>> {
   return pipe(
-    catchError(e => of(errorState(e instanceof Error ? e : new Error('unknown error')))),
+    catchError(e => of(errorState(e instanceof Error || e instanceof HttpErrorResponse ? e : new Error('unknown error')))),
```
smadbe, line 59: Isn't a change you forgot to remove?

ThomasJuster: No this one was intentional, I can replace it by a predicate function that will test an implementation instead of instances, that might make actually more sense

ThomasJuster: fixed

smadbe (approved):

PR #897, opened 2022-02-07
allow domain-specific configs

### LINKED ISSUE #894: Allow domain-specific configs
## Overview
The same "instance" of the platform (and its backend/db) is meant to be used for several use at a same type, typically with different domain (one for general training, one for a specific contest, one for a specific university, ...). 
Each of these instance may have some aspects of the config should differ:
- activity on which the users land by default
- default/accepted languages on what path
- platform title
- auth type: cookies may not be usable on a different domain on http for instance
- set of enabled/disabled features
- theme

## Acceptance criteria
- [ ] create a system which allows several configs to be defined, with one default and domain-specific ones redefining (overriding) several keys
- [ ] it should be testable locally / on dev: maybe the domain could be overridden by a session variable (or something else) for

ThomasJuster:
## Description

Fixes #894 

...

## Test cases

- [ ] Case 1: No regression test
  1. Given I am the any user
  2. When I go to the [home page](https://dev.algorea.org/branch/feat/app-config/en/#/)
  3. Then I am redirected to "Parcours officiels" item
  4. Then I see the app title is "Algorea platform"
  5. And I open my devtools > session storage
  5. Then I see the platform works with an auth by cookie

- [ ] Case 2: "Example" app
  1. Given I am the any user
  2. When I go to the [home page](https://dev.algorea.org/branch/feat/app-config/en/#/?config_preset=example)
  3. Then I am redirected to "Motif art" item
  4. Then I see the app title is "Example app"
  5. And I open my devtools > session storage
  5. Then I see the platform works with an auth by token

- [ ] Case 3: "Demo" app
  1. Given I am the any user
  2. When I go to the [home page](https://dev.algorea.org/branch/feat/app-config/en/#/?config_preset=demo)
  3. Then I am redirected to "SNT" item
  4. Then I see the app title is "Demo app"
  5. And I open my devtools > session storage
  5. Then I see the platform works with an auth by cookie

- [ ] Case 4: By domain (to be tested on localhost)
  1. Given I added "demo.localhost" to my hosts and I serve the app via `ng serve --host demo.localhost`
  2. And I am the any user
  2. When I go to the [home page](http://demo.localhost/#/)
  3. Then I see the title is "Demo app"

### DIFF
--- a/src/app/shared/helpers/config.ts
+++ b/src/app/shared/helpers/config.ts
@@ -33,5 +33,48 @@ export interface Environment {
 }
 
 type Config = Environment; // config may be someday an extension of the environment
+type ConfigOverride = Omit<Config, 'production' | 'allowForcedToken'>;
+const presetNames = [ 'example', 'demo' ] as const;
+type Preset = typeof presetNames[number];
+const isPreset = (value: any): value is Preset => presetNames.includes(value as Preset);
+const presetQueryParam = 'config_preset';
 
-export const appConfig: Config = environment;
+const presets: Record<Preset, Partial<ConfigOverride>> = {
+  example: {
+    defaultActivityId: '1625159049301502151', // Motif Art
+    defaultTitle: 'Example app',
+    authType: 'tokens',
+    itemPlatformId: 'Example',
+  },
+  demo: {
+    defaultActivityId: '1352246428241737349', // SNT
+    defaultTitle: 'Demo app'
+  },
+};
+
+export const appConfig: Config = {
+  ...environment,
+  ...getPresetConfigFromUrl(), // spreading `null` is fine
+};
+
+function getPresetConfigFromUrl(): Partial<ConfigOverride> | null {
+  const preset = getPresetFromDomain() ?? getPresetFromQuery();
+  return preset && presets[preset];
+}
+
+function getPresetFromQuery(): Preset | null {
+  const url = globalThis.location.href;
+  const search = url.includes('?') ? new URLSearchParams(url.slice(url.indexOf('?'))) : null;
+  const preset = search?.get(presetQueryParam) ?? null;
+  if (!preset) return null;
+  if (!isPreset(preset)) throw new Error(`Unknown preset "${preset}", expected one of: ${presetNames.join(', ')}`);
+  return preset;
+}
+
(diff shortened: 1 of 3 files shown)


src/environments/environment.prod.ts
```diff
@@ -20,3 +20,24 @@ export const environment: Environment = {
...
+    itemPlatformId: 'Example',
+  },
+  demo: {
+    defaultActivityId: '1352246428241737349', // SNT
+    defaultTitle: 'Demo app'
+  },
+};
+
+export function getPresetNameByDomain(domain: string): string | null {
+  switch (domain) {
```
smadbe, line 38: I would have expected the record key to be the pattern matcher (possibly a regexp or a list if we want to match several ones). But it can be "fixed" later.

src/environments/environment.prod.ts
```diff
@@ -20,3 +20,24 @@ export const environment: Environment = {
...
+    case 'example.algorea.org': return 'example';
```
smadbe, line 39: I think it would make sense to include the protocol as typically we may have the "http" where cookies cannot be used and "https" using cookies.

src/app/shared/helpers/config.ts
```diff
@@ -34,4 +34,16 @@ export interface Environment {
...
 
-export const appConfig: Config = environment;
+const presetQueryParam = 'config_preset';
+function getPresetNameFromQuery(): string | null {
+  const url = globalThis.location.href;
+  const search = url.includes('?') ? new URLSearchParams(url.slice(url.indexOf('?'))) : null;
+  return search?.get(presetQueryParam) ?? null;
+}
+
+const presetName = getPresetNameByDomain(globalThis.location.hostname) ?? getPresetNameFromQuery();
```
smadbe, line 44: You have to explain in code that `getPresetNameFromQuery` in for testing purpose, otherwise each time we'll try to figure out what it is.

smadbe (approved):

src/environments/environment.prod.ts
ThomasJuster: fixed

src/app/shared/helpers/config.ts
ThomasJuster: fixed

PR #898, opened 2022-02-08
lti: keep same task on refresh

### LINKED ISSUE #893: LTI: keep the same task on refresh
## Overview

In "LTI mode", when refreshing on any  task, you come back to the first task. "Just" keep the current task.

## Acceptance criteria
- [ ] In "LTI mode", when refreshing on any task, stay on the same current task.

## Hint/idea

To prevent breaking the concept of `redirect_to_sub_path`, what you could do is to pass as query parameter, when redirecting, the landing path (let's say `?from_path=`). Then, in LTI component/service, if `from_path` present, you could redirect back to this path after having done the LTI setup (instead of navigating to the first child).

ThomasJuster:
## Description

Fixes #893 

The LTI route must filter when to take the `?from_path` into account. Here's the flow:
1. Launch AlgoreaFrontend app, land on LTI route component. LTI component sets `redirect_to_sub_path` session storage item with current url + query param `is_redirection=1` to avoid infinite loops
2. Login is performed elsewhere and the page comes back to AlgoreaFrontend app on home page
3. From App component, redirect to sub path provided by session storage, add the query param `?from_path={current url}`, which in our case redirects to LTI route component
4. LTI route component **ignores `?from_path` at this point because we don't want to redirect to the home**, checks out that the user is logged, fetches item first child info, etc. **and appends to `redirect_to_sub_path` (session storage) the query param `?allow_from_path`
5. The user can now navigate as usual in LTI mode, if he refreshed, then …
6. The app component detects the `redirect_to_sub_path` (which contains now `?allow_from_path` for step 7) and redirects in our case to the LTI route with a populated `?from_path` query param
7. The LTI route component reads the query params, finds `allow_from_path` **and** `from_path`, retrieves all the required info as usual and then resets the `redirect_to_sub_path` as in step 4 in case the user re-refreshes. And since both `allow_from_path` and `from_path` are provided, instead of navigating to item first child, now it redirects to `from_path` value.

:tada:

Since the tests imply that the LTI login part works as expected, I also updated the `is_redirection` param to set it as `is_redirection=1` instead of `is_redirection=`

## Test cases

:warning: **All tests must be run using a local environment (localhost)**

- [ ] Case 1:
  1. Given I am an authorized moodle user
  2. When I go to [moodle "Test LTI 2"](http://moodletest.algorea.org/course/view.php?id=3)
  3. And I click on "[Test dev.algorea.org (new platform localhost)](http://moodletest.algorea.org/mod/lti/view.php?id=8)"
  4. Then I am redirected to Algorea platform (as before), which opens task "Les boucles bornées ou boucles à compteur" in LTI mode
  5. And I navigate to next sibling "Les boucles imbriquées" using header right arrow
  6. (Optional) And I throttle my network to slow down the page transitions
  7. And I refresh my page
  8. Then (after loads) I am redirected to the LTI page, which redirects to same page "Les boucles imbriquées" in LTI mode
  9. Steps 7+8 can be repeated indefinitely

### DIFF
--- a/src/app/modules/lti/pages/lti/lti.component.ts
+++ b/src/app/modules/lti/pages/lti/lti.component.ts
@@ -9,9 +9,9 @@ import { errorIsHTTPForbidden } from 'src/app/shared/helpers/errors';
 import { isNotNull } from 'src/app/shared/helpers/null-undefined-predicates';
 import {
   allowFromPathKey,
+  appendUrlWithQuery,
   fromPathKey,
   getRedirectToSubPathAtInit,
-  getUrlWithAllowFromPath,
   setRedirectToSubPathAtInit,
 } from 'src/app/shared/helpers/redirect-to-sub-path-at-init';
 import { CheckLoginService } from 'src/app/shared/http-services/check-login.service';
@@ -114,7 +114,7 @@ export class LTIComponent implements OnDestroy {
 
         const redirectUrl = getRedirectToSubPathAtInit();
         if (!redirectUrl) throw new Error('redirect url should be set by now');
-        setRedirectToSubPathAtInit(getUrlWithAllowFromPath(redirectUrl));
+        setRedirectToSubPathAtInit(appendUrlWithQuery(redirectUrl, allowFromPathKey, '1'));
 
         if (fromPath) {
           void this.router.navigateByUrl(fromPath);
--- a/src/app/modules/lti/pages/lti/lti.component.ts
+++ b/src/app/modules/lti/pages/lti/lti.component.ts
@@ -33,6 +33,7 @@ const noChildError = new Error(LTIError.NoChild);
 const loginError = new Error(LTIError.LoginError);
 
 const isRedirectionParam = 'is_redirection';
+// Only handle `?from_path` query when `?lti_use_from_path=1`, ignore it otherwise
 const useFromPathKey = 'lti_use_from_path';
 const loginIdParam = 'user_id';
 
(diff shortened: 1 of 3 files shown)


smadbe (commented):
🎉 yeah, it starts to be complex.

From what I get from your explanation (and from the code with a very very check) I have the impression that:

- `redirect_to_sub_path`,  allow\_from\_path and `from_path` are LTI-agnostic and completely generic, yes nice

- for `redirect_to_sub_path` and`from_path` they have to be generic as they are handled by the app component... just a redirection and just an indication of the source. 

- thew `allow_from_path`  parameter is only written and read by LTI. 

- most complexity of the new "workflow" comes from the `allow_from_path`, which is enabling the LTI component to take into consideration another parameter 

-> from these observations, I have the impression that `allow_from_path`  should be something only linked  with LTI... keeping the global app away from the need of understanding what this  `allow_from_path` is for.
So I would even rename it to something like "lti_allow_from_path" or just "lti_use_from_path"

src/app/modules/lti/pages/lti/lti.component.ts
```diff
@@ -40,7 +46,11 @@ export class LTIComponent implements OnDestroy {
   private loginId$ = this.activatedRoute.queryParamMap.pipe(map(queryParams => queryParams.get(loginIdParam)));
 
   private isRedirection$ = this.activatedRoute.queryParamMap.pipe(
-    map(queryParams => queryParams.has(isRedirectionParam)),
+    map(queryParams => queryParams.get(isRedirectionParam) === '1'),
+  );
+
+  private fromPath$ = this.activatedRoute.queryParamMap.pipe(
+    map(queryParams => (queryParams.get(useFromPathKey) === '1' ? queryParams.get(fromPathKey) : null)),
```
smadbe, line 53: I don't feel there was real need for a observable here (versus `activatedRoute.snapshot`) (as it is an immediate process and the arg is not supposed to change I assume), but that's ok.

smadbe (approved):
few remarks but no more reviews required (except if you need so)

PR #910, opened 2022-02-10
Hides parent button when false, adds animation

### LINKED ISSUE #895: Page navigator: do not show the "up" by default
## Overview

In some case, the "up" arrow of the page navigator is misleading as users are not allowed to go up or are supposed to see the content as "flat" (just a set of tasks).

Screenshot 2022-02-04 at 18 01 23
Screenshot 2022-02-04 at 18 01 40

## Acceptance criteria
- [ ] While the UI is unsure about showing or not the "up" arrow, just don't show the "up" button (instead of having it greyed out).
- [ ] When "up" is not shown the "rounding border" should be displayed correctly on the "left" button.
- [ ] When we know we need the "up" button, make it appear with a quick "right-to-left" animation 
- [ ] Should work in collapsed (full-frame) and non-collapsed mode
- [ ] The left-right button should be shown as before (whether they are enabled or not)

Iloveall:
## Description

Fixes #895

Not sure about animation in page navigator as buttons disappears on loading time 

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/page-navigator-do-not-show-the-up-by-default/en/#/activities/by-id/4102;path=4702;parentAttempId=0/details)
  3. And I click on collapse left menu button
  4. Then I see parent/next/prev buttons
  5. Click on parent button
  6. Then I see button disappears with animation
  7. Click on browser back button
  8. Then I see button appears with animation

- [x] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/page-navigator-do-not-show-the-up-by-default/en/#/activities/by-id/4102;path=4702;parentAttempId=0/details)
  3. Then I see parent/next/prev buttons
  5. Click on parent button
  6. Then I see button disappears
  7. Click on browser back button
  8. Then I see button appears

### DIFF
--- a/src/app/core/components/content-top-bar/content-top-bar.component.ts
+++ b/src/app/core/components/content-top-bar/content-top-bar.component.ts
@@ -3,7 +3,7 @@ import { ModeAction, ModeService } from '../../../shared/services/mode.service';
 import { ContentInfo } from '../../../shared/models/content/content-info';
 import { Observable, of } from 'rxjs';
 import { CurrentContentService } from '../../../shared/services/current-content.service';
-import { delay, switchMap } from 'rxjs/operators';
+import { delay, switchMap, filter } from 'rxjs/operators';
 import { ActivityNavTreeService, SkillNavTreeService } from '../../services/navigation/item-nav-tree.service';
 import { isItemInfo } from '../../../shared/models/content/item-info';
 import { FullFrameContent } from 'src/app/shared/services/layout.service';
@@ -32,6 +32,7 @@ export class ContentTopBarComponent {
       return content.route.contentType === 'activity' ?
         this.activityNavTreeService.navigationNeighbors$ : this.skillNavTreeService.navigationNeighbors$;
     }),
+    filter(navigationNeighbors => !!navigationNeighbors?.isReady),
   );
 
   constructor(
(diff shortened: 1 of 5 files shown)


smadbe:
If I check frame-by-frame the animation, it is not really what I had in mind:
- on hide: make the "up" button slide under the right button, making the whole set of 3 button like it resize to a size of 2 keeping its rounded shape
- on show: the opposite 

https://user-images.githubusercontent.com/1053150/153831745-78068e9f-b65c-4216-b44f-6aac8a4839c8.mov

Do you think you can improve it? (without spending 4 hours on it). If not, it is ok like that.

ThomasJuster (commented):
Due to @smadbe's comment, I'll wait for changes before reviewing

smadbe:
When navigating from a chapter with a parent to chapter with a parent, I have this useless, can it be avoided? (by not animating while the status is "unknown")

https://user-images.githubusercontent.com/1053150/153889521-7db7ce1a-b50c-440f-b572-269c7465ed97.mov

smadbe:
Nice 😎 Good for me 👍

ThomasJuster (approved):

PR #920, opened 2022-02-11
do not use token for telecom paris as a backend fix should make it work with cookies

smadbe:
do not use token for telecom paris as a backend fix should make it work with cookies


smadbe:
will be done on a separate branc

PR #928, opened 2022-02-14
Adds theme supporting

### LINKED ISSUE #896: Multi-theme (color)
## Overview

Allow another theme (for now just the color set) to be defined.

## Acceptance criteria
- [ ] add a "theme" string parameter in the config
- [ ] load a different css color set (at launch) depending on this parameter 
- [ ] isolate these different themes in different files for easier modifications together with a customer

Iloveall:
## Description

Fixes #896

The client will use the app only in a mode when the left menu is always collapsed, so we don't care about any colour in the left menu

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/multi-theme-color/en/#/activities/by-id/4702;path=;parentAttempId=0/details)
  3. Then I see different colours and font family

- [x] Case 2:
  1. Given I am the usual user
   2. When I go to [this page](https://dev.algorea.org/branch/feature/multi-theme-color/en/#/activities/by-id/4702;path=;parentAttempId=0/details)
  3. Open browser inspect panel
  4. Find alg-root tag under elements tab and remove data-theme="client-demo" attribute
  5. Then I see base theme

- [x] Case 3:
  1. Given I am any user
  2. When I go to [this page with config preset](https://dev.algorea.org/branch/feature/multi-theme-color/en/#/activities/by-id/4702;path=;parentAttempId=0/details?config_preset=telecomParis)
  3. Then the theme should be applied (darker blue etc.)


src/assets/fonts/source-sans-pro/generator_config.txt
```diff
@@ -0,0 +1,5 @@
+# Font Squirrel Font-face Generator Configuration File
```
ThomasJuster, line 1: Do we actually need this file :thinking: ?

src/assets/fonts/source-sans-pro/sourcesanspro-bold-demo.html
```diff
@@ -0,0 +1,621 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
ThomasJuster, line 1: Do we actually need this file :thinking: ?

src/assets/fonts/source-sans-pro/sourcesanspro-italic-demo.html
```diff
@@ -0,0 +1,620 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
ThomasJuster, line 1: Do we actually need this file :thinking: ?

src/assets/fonts/source-sans-pro/sourcesanspro-light-demo.html
```diff
@@ -0,0 +1,621 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
ThomasJuster, line 1: Do we actually need this file :thinking: ?

src/assets/fonts/source-sans-pro/sourcesanspro-regular-demo.html
```diff
@@ -0,0 +1,621 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
ThomasJuster, line 1: Do we actually need this file :thinking: ?

src/assets/fonts/source-sans-pro/specimen_files/grid_12-825-55-15.css
```diff
@@ -0,0 +1,129 @@
+/*Notes about grid:
```
ThomasJuster, line 1: Do we actually need this file :thinking: ?

src/assets/fonts/source-sans-pro/specimen_files/specimen_stylesheet.css
```diff
@@ -0,0 +1,396 @@
+@import url('grid_12-825-55-15.css');
```
ThomasJuster, line 1: Do we actually need this file :thinking: ?

src/assets/scss/components/button.scss
```diff
@@ -1,11 +1,11 @@
...
 @import 'src/geometry.scss';
 
-$secondary-color: #4a4a4a;
-$warning-color: #f5a623;
-$danger-color: #ff001f;
-
 .alg-button {
+  --secondary-color: #4a4a4a;
+  --warning-color: #f5a623;
+  --danger-color: #ff001f;
```
ThomasJuster, line 7: Is it normal to have such variables declared at a button-level ?
I would expect colors to be declared at root or theme level :thinking:

src/app/shared/helpers/config.ts
```diff
@@ -30,6 +30,8 @@ export interface Environment {
   authType: 'tokens' | 'cookies',
 
   itemPlatformId: string,
+
+  theme: string,
```
ThomasJuster, line 34: Since the `theme` here is required to be a string, how is represented the "default" theme ?
In other words: in the PR I only saw the usage with a provided theme, how will it work without a specific theme ?

angular.json
```diff
@@ -46,7 +46,8 @@
               "node_modules/primeng/resources/primeng.min.css",
               "node_modules/@fortawesome/fontawesome-free/css/all.min.css",
               "src/styles.scss",
-              "src/assets/fonts/titillium/stylesheet.css"
+              "src/assets/fonts/titillium/stylesheet.css",
+              "src/assets/fonts/source-sans-pro/stylesheet.css"
```
ThomasJuster, line 50: About bundle size, @smadbe In the issue you wrote:

>  load a different css color set (at launch) depending on this parameter

I understood that the theme file (holding variables) had to be loaded **only** when the theme is "active". When checking at the file for "client-demo", the file itself is very lightweight therefore I don't know if the optimization is relevant ([file anchor](https://github.com/France-ioi/AlgoreaFrontend/pull/928/files#diff-23fad3928bf3e71585eb4a6e3b2ff72dae02c3ed5a44d2f0deb1d49f0cb3b1a3R33))

However, here we seem to load 2 fonts (source-sans-pro fonts are 21~34kB **per font**) to use only one in the end. In my opinion we should load one or the other (using CDN ?), how do you see it ? (I didn't find indication in the issue)

src/assets/scss/components/button.scss
Iloveall: It's not normal, but it was done before long time ago. I guess refactoring scss approach is out of task scope, otherwise need additional time for investigate why it was done like this, then fix, then tested.

src/app/shared/helpers/config.ts
Iloveall: Actually, there is no value for standard theme, because variables for standard theme declared on root lvl. For clear, I can add 'default' value and check it in condition before put to tag. But for now any value which not matched to declared data-theme=[name] would be default theme

src/assets/fonts/source-sans-pro/generator_config.txt
Iloveall: Do you want to remove those files? Actually in titillium this info was helpfully, i didn't spent time for figuring which service/approach was used before. Also I guess this file gives possibility to continue edit fonts setting but not create from scratch again

src/app/shared/helpers/config.ts
ThomasJuster: Ok, then I think the `theme` should be nullable, or required with specific themes like `theme: 'default' | 'client-demo'`

src/assets/scss/components/button.scss
ThomasJuster: Ok, can you check if there's a refactoring task for that and if not create it ?

Iloveall: Fixed. Issue #930

angular.json
ThomasJuster: ok :+1:

ThomasJuster (approved):

PR #931, opened 2022-02-15
push language to task

### LINKED ISSUE #911: Push language to tasks
## Overview
Tasks needs to know what's the current language of the platform so that the task can be presented in the given language if available.

## Acceptance criteria
- [ ] Pass the current language of the task (`language_tag` in `strings` of the item) as a `sLocale` query string parameter added to the task.

Example task: https://tasks.algorea.org/?taskId=145659473473003296

ThomasJuster:
## Description

Fixes #911 

I noticed that if the visited item/activity does not have strings for a given language (say "fr") but we are _actually visiting_ the platform in "fr", the item.string.language_tag is "en" anyway.
As a consequence, if we base the task locale on the item strings, we might have inconsistencies ; should I make a change to use the locale from the locale service instead ?

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. When I go to [this activity with app in English](https://dev.algorea.org/branch/feat/task-locale/en/#/activities/by-id/766891051147204195;path=4702,1625159049301502151,5207993233955027100;parentAttempId=0/details)
  3. Then the task is loaded in English

- [ ] Case 2:
  1. Given I am any user
  2. When I go to [this activity with app in French](https://dev.algorea.org/branch/feat/task-locale/fr/#/activities/by-id/766891051147204195;path=4702,1625159049301502151,5207993233955027100;parentAttempId=0/details)
  3. Then the task is loaded in French

### DIFF
--- a/src/app/modules/item/http-services/get-item-by-id.service.ts
+++ b/src/app/modules/item/http-services/get-item-by-id.service.ts
@@ -15,6 +15,7 @@ export const itemDecoder = pipe(
     string: pipe(
       D.struct({
         title: D.nullable(D.string),
+        languageTag: D.string,
       }),
       D.intersect(
         D.partial({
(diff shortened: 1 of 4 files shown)


smadbe:
> I noticed that if the visited item/activity does not have strings for a given language (say "fr") but we are actually visiting the platform in "fr", the item.string.language_tag is "en" anyway.
As a consequence, if we base the task locale on the item strings, we might have inconsistencies ; should I make a change to use the locale from the locale service instead ?

That's ok as it. They will have to add languages on the item (basically translate the title). The UI is just missing but that's a detail 🙄

smadbe (approved):

PR #952, opened 2022-02-23
delegate layout initialization to routes

### LINKED ISSUE #943: Defer the display of the left menu
## Overview
In order to prevent the display then hide of the left menu, wait for being sure you need the left menu before displaying it.

## Acceptance criteria
- [ ] Make sure the left menu is not shown for cases where it is hidden just after: LTI, when opening tasks
- [ ] Make sure the top-right widget is not shown while we are not sure we are not in LTI mode yet
- [ ] Show the left menu (at app launch) at the end for any non-task content (without animation)
- [ ] Make sure the top-right widget is always shown for non-LTI case, and never for LTI
- [ ] Keep the non-LTI code agnostic of LTI

ThomasJuster:
## Description

Fixes #943 

...

## Test cases
- [ ] Case: Group page
  1. Given I am usual user
  2. When I navigate to [this page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/groups/by-id/672913018859223173;path=52767158366271444/details)
  3. Then I see the left menu is displayed without animation

- [ ] Case: User page
  1. Given I am usual user
  2. When I navigate to [this page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/groups/users/752024252804317630) or [profile page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/groups/users/670968966872011405/personal-data)
  3. Then I see the left menu is displayed without animation

- [ ] Case: My groups page
  1. Given I am usual user
  2. When I navigate to [this page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/groups/mine)
  3. Then I see the left menu is displayed without animation

- [ ] Case: Item by id → Task → normal mode
  1. Given I am usual user
  2. When I navigate to [this page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/activities/by-id/6010615591035594227;path=4702,1625159049301502151;attempId=0/details)
  3. Then I see the left menu is never displayed and starts collapsed

- [ ] Case: Item by id → Task → edit mode
  1. Given I am usual user
  2. When I navigate to [this page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/activities/by-id/6010615591035594227;path=4702,1625159049301502151;attempId=0/edit)
  3. Then I see the left menu is never displayed and starts collapsed

- [ ] Case: Item by id → Chapter → normal mode
  1. Given I am usual user
  2. When I navigate to [this page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/activities/by-id/1625159049301502151;path=4702;attempId=0/details)
  3. Then I see the left menu is displayed without animation

- [ ] Case: Item by id → Chapter → edit mode
  1. Given I am usual user
  2. When I navigate to [this page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/activities/by-id/1625159049301502151;path=4702;attempId=0/edit)
  3. Then I see the left menu is displayed without animation

- [ ] Case: Item by id → Item not found
  1. Given I am any user
  2. When I navigate to [this page](https://dev.algorea.org/branch/feat/defer-left-menu-display/en/#/activities/by-id/12234104923/details)
  3. Then I see the left menu is displayed without animation

- [ ] Case: LTI page **:warning: can be tested only locally**
  1. Given I am an authenticated moodle user
  2. When I navigate to [moodle](http://moodletest.algorea.org/)
  3. And I click on LTI Test 2
  4. And I click on "Test dev.algorea.org (new platform localhost)"
  3. Then I see platform is launched in LTI mode and never displays the left menu


smadbe (changes requested):
I would have preferred avoiding this call in any routed content (if a new page is created, it is easy to miss that if that page is the first visited, there is a problem).

smadbe:
> I would have preferred avoiding this call in any routed content (if a new page is created, it is easy to miss that if that page is the first visited, there is a problem). But I don't see any much better alternative 🤷‍♂️.

Typically, it may cause issue for error cases as well... and from I quickly see, this is not handled properly if an item cannot be loaded (and it is not in the tests, so I suppose it is not).

src/app/shared/services/layout.service.ts
```diff
@@ -8,6 +8,12 @@ export interface FullFrameContent {
   animated: boolean,
 }
 
+interface ConfigureOptions {
+  expanded: boolean,
```
smadbe, line 12: add some "doc" please. I am not sure a "layout expanded" makes sense out of context.

Edit: maybe "in context" as well ;-). Isn't there a better word for that? `fullFrameContent`? ... ?

Edit2: ah I now see the doc is in the function spec... is it really useful to define this type here? cannot everything be in the function definition?

src/app/shared/services/layout.service.ts
```diff
@@ -8,6 +8,12 @@ export interface FullFrameContent {
...
+  canToggle?: boolean,
+  showTopRightControls?: boolean,
```
smadbe, line 14: Shouldn't `canToggle` be `canToggleLeftMenu` or `canToggleFullFrameContent` or whatever ?

src/app/shared/services/layout.service.ts
```diff
@@ -26,25 +32,26 @@ export class LayoutService {
...
-   * @param canToggle Defines for the lifetime of the app if the left menu can be toggled or not
-   * @param showTopRightControls Defines for the lifetime of the app if the top right controls are shown
+   * @param {object} options
+   * @param options.expanded initial expanded value
+   * @param options.canToggle Defines for the lifetime of the app if the left menu can be toggled or not
+   * @param options.showTopRightControls Defines for the lifetime of the app if the top right controls are shown
    */
-  initialize(expanded: boolean, canToggle: boolean, showTopRightControls: boolean): void {
-    if (this.initialized) return;
+  configure({ expanded, showTopRightControls = true, canToggle = true }: ConfigureOptions): void {
```
smadbe, line 45: For one-time configuration: It is as this would need to be set only once, but at the end, shouldn't the left-menu display/hide because of visiting a chapter/task (when not using the left menu) use the same mecanism?
Look strange to me that item-by-id deals with roughly the same thing (setting the "expanded") twice but by different ways.

BTW, there is maybe no reason to return when already configured. We could consider that "canToggle" and "showTopRightControls" may be changed. As there are no need for that for now, you don't need to change that, but if it may help with my remark above... why not.

smadbe (changes requested):

src/app/shared/services/layout.service.ts
ThomasJuster: fixed

src/app/shared/services/layout.service.ts
ThomasJuster: fixed

src/app/shared/services/layout.service.ts
ThomasJuster: > For one-time configuration: It is as this would need to be set only once, but at the end, shouldn't the left-menu display/hide because of visiting a chapter/task (when not using the left menu) use the same mecanism?

It is meant to be set only at app startup because it's the only case when there is no animation, that's why it was originally called `initialize()`.
After configuration, the layout can be toggled using the other method `toggleFullFrameContent` which 
has intentionally less parameters and enables animations.
In my opinion, the 2 cases at-startup+no animation/after-startup+animations obey to different use cases, thus they are separate functions.

> Look strange to me that item-by-id deals with roughly the same thing (setting the "expanded") twice but by different ways.

Fixed, only the solve missing path error is only handled.

src/app/shared/services/layout.service.ts
```diff
@@ -25,10 +26,27 @@ export class LayoutService {
...
+  private initialized = false;
+
+  /**
+   * This method allows to defer layout initialization to any consumer, expectedly routes.
+   * Only first call in taken into account, later calls are ignored.
+   * @param expanded initial expanded value
+   * @param canToggle Defines for the lifetime of the app if the left menu can be toggled or not
+   * @param showTopRightControls Defines for the lifetime of the app if the top right controls are shown
+   */
+  initialize(expanded: boolean, canToggle: boolean, showTopRightControls: boolean): void {
```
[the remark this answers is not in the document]
ThomasJuster: fixed

src/app/shared/services/layout.service.ts
smadbe: > It is meant to be set only at app startup because it's the only case when there is no animation, that's why it was originally called `initialize()`. After configuration, the layout can be toggled using the other method `toggleFullFrameContent` which has intentionally less parameters and enables animations. 

The animation could easily be solved by replacing `animated: false` by `animated: alreadyConfigured` :-)

And between `this.layoutService.configure({ expanded: false })` and `this.layoutService.toggleFullFrameContent(false)`, there is the same number of arg... and 1 more character 😁

Ok, that is because I haven't check the different in usage between the two calls but I can check it...

### First, in layout service: 
L53 is useless as `canToggle` is `false` by L52.
On L52, `this.fullFrameContent.value.canToggle` can only be false if `!this.configured` so one of the 2 is useless as well. 

So you end up with:
```ts
toggleFullFrameContent(expanded: boolean): void {
    if (!this.configured) return;
    this.fullFrameContent.next({ expanded, canToggle: false, animated: true });
  }
```

### Calls, in item-by-id
1)
```ts
this.layoutService.configure({ expanded: isTask(data.item) });
```
called on `itemDataSource.state$` change, when ready, the first time

2)
```ts
this.layoutService.toggleFullFrameContent(isTask(item) && !(history.state as Record<string, boolean | undefined>).preventFullFrame);
```
called on  `itemDataSource.state$` change, when ready, each time (including the first time I think).

I suppose the first time it is called twice. 
A) If (2) is called before (1), it will be ignored (because of L52)... so the history prevent full frame is ignored, is it ok? (possibly a bug ?)
B) If (1) is called before (2) is (1) is just useless, right?

I suppose that the order of definition is deterministic, but quite unknown to me (via command order?). Possibly risky if someone reorganize the file.
Anyway, if I haven't done any mistake in my analysis, the fact that

> item-by-id deals with roughly the same thing (setting the "expanded") twice but by different ways

(from my previous review) is not really a detail as it leads to either a useless code or a possible bug... and even if there is no problem, it is unclear enough not to treat them separately (except if I misunderstood a major aspect).

src/app/shared/services/layout.service.ts
smadbe: My 2 first comment were ignored actually. :-/

src/app/shared/services/layout.service.ts
```diff
@@ -25,10 +26,32 @@ export class LayoutService {
...
   contentFooter$ = this.contentFooter.asObservable();
 
+  private configured = false;
+
+  /**
+   * This method allows to defer layout initialization to any consumer, expectedly routes.
+   * Only first call in taken into account, later calls are ignored.
+   * @param {object} options
+   * @param options.expanded initial expanded value
+   * @param options.canToggle Defines for the lifetime of the app if the left menu can be toggled or not
```
smadbe, line 36: parameter name not adapted

src/app/shared/services/layout.service.ts
ThomasJuster: I thought they were by moving the declaration in the function. Fixed.
I did some renaming and switched from `FullFrameContent.expanded` to `FullFrameContent.active`, it help renaming the configure fumction parameter

src/app/shared/services/layout.service.ts
ThomasJuster: For the useless line in layout service: fixed
---
For calls in item by id:
A) Call (1) will always be called before call (2) because the subscription triggering call (1) is set before subscription triggering call (2). Anyway the first time, `history.state` has no chance to be populated since the history state can only be populated via an internal navigation. No possible bug here to me.
B) If (1) is called before (2), then (2) is indeed effectless. This is the case only when loading the first app route, afterwards, the call (1) will always be ignored and call (2) will be the one taken into account.

Whatever the calls order, the behavior will remain the same:
1st load: the special case
- (1) called before (2): layout is configured by (1), (2) is effectless because `fullFrame.active` is the same as current one.
- (2) called before (1): (2) is ignored because layout isn't configured yet, (1) configures the layout.
nth load: only via in-app navigation
- (1) called before (2): (1) is ignored because layout is configured, (2) is taken into account
- (2) called before (1): (2) is taken into account, (1) is ignored because layout is configured.

To me, it seems deterministic and resilient to changes, doesn't it?

src/app/shared/services/layout.service.ts
ThomasJuster: fixed

src/app/shared/services/layout.service.ts
ThomasJuster: I forgot to address this comment:
> The animation could easily be solved by replacing animated: false by animated: alreadyConfigured :-)
And between this.layoutService.configure({ expanded: false }) and this.layoutService.toggleFullFrameContent(false), there is the same number of arg... and 1 more character :grin:

I can replace `animated: false` by `animated: alreadyConfigured`, but then it still means I have to _configure_ something, causing the second comment (suggesting to merge the 2 methods) to be irrelevant, unless I'm missing something.
Anyway the 2 methods do not have the same number of arguments, `configure()` has 3 arguments with 2 having defaults values while `toggleFullFrame()` has only one.

ThomasJuster: Maybe it will be easier to discuss this over the phone

src/app/shared/services/layout.service.ts
```diff
@@ -1,42 +1,49 @@
...
+
   /** Expands the content by hiding the left menu and select headers */
-  fullFrameContent$ = this.fullFrameContent.pipe(distinctUntilChanged((a, b) => a.expanded === b.expanded && a.canToggle === b.canToggle));
+  private fullFrame = new BehaviorSubject<FullFrameContent>({ active: true, canToggle: false, animated: false });
+  fullFrame$ = this.fullFrame.pipe(distinctUntilChanged((a, b) => a.active === b.active && a.canToggle === b.canToggle));
 
-  private showTopRightControls = new BehaviorSubject(true);
-  readonly showTopRightControls$ = this.showTopRightControls.pipe(distinctUntilChanged());
+  private showTopRightControls = new BehaviorSubject(false);
+  showTopRightControls$ = this.showTopRightControls.pipe(distinctUntilChanged());
```
ThomasJuster, line 23: Usage of observables is in the end required because there's no other way to detect service prop changes in components :shrug:

smadbe: Of course, `layoutInitialized = this.layoutService.initialized` would not be updated... right using observable is the only way here.

src/app/shared/services/layout.service.ts
```diff
@@ -1,42 +1,48 @@
...
 export class LayoutService {
   // Service allowing modifications of the layout
-  private fullFrameContent = new BehaviorSubject<FullFrameContent>({ expanded: false, canToggle: true });
+
+  private initialized = false;
+
   /** Expands the content by hiding the left menu and select headers */
-  fullFrameContent$ = this.fullFrameContent.pipe(distinctUntilChanged((a, b) => a.expanded === b.expanded && a.canToggle === b.canToggle));
+  private fullFrame = new BehaviorSubject<FullFrameContent>({ active: true, canToggle: false, animated: false });
+  fullFrame$ = this.fullFrame.pipe(distinctUntilChanged((a, b) => a.active === b.active && a.canToggle === b.canToggle));
```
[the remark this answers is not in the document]
ThomasJuster: Yes, what do you animate if the `active` value does not change?

src/app/shared/services/layout.service.ts
```diff
@@ -1,42 +1,48 @@
...
-    this.fullFrameContent.next({ expanded, canToggle });
-  }
+   * Configure layout, expectedly called by routes.
+   */
+  configure({ fullFrameActive, showTopRightControls, canToggleFullFrame }: {
+    fullFrameActive: boolean,
+    canToggleFullFrame?: boolean,
+    showTopRightControls?: boolean,
+  }): void {
+    const canToggleFallback = this.initialized ? this.fullFrame.value.canToggle : true;
```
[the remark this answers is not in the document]
ThomasJuster: Fixed

src/app/shared/services/layout.service.ts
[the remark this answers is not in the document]
smadbe: Make sense.

smadbe (approved):
👍

PR #954, opened 2022-02-24
hide task loader in favor of app loader

### LINKED ISSUE #945: Hide task loading in favor of app loading
## Overview
We display a loading spinner while a task is not loaded yet, but some tasks have their own spinner as well. As a result, we may end up with double spinner such as: 
image(1)

Let's "hide" the task while it does not declare itself as loaded.

## Acceptance criteria
- [ ] Make our "loading spinner layer" opaque and displayed over the task iframe so that the task is hidden while loading
- [ ] When the task is loaded, remove our "loading spinner layer" and show the task

ThomasJuster:
## Description

Fixes #945 

...

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. When I go to [this task](https://dev.algorea.org/branch/fix/hide-task-loading/en/#/activities/by-id/8148962848812891795;path=4702,458602124916982205,4769434107088212490,8123435593727593341,5072731755477071719,8600972556668887052;parentAttempId=0/details)
  4. Then I see the task loader is hidden in favor of the app loader until the task is ready

### DIFF
--- a/src/app/modules/item/pages/item-content/item-content.component.html
+++ b/src/app/modules/item/pages/item-content/item-content.component.html
@@ -9,7 +9,7 @@
       <alg-chapter-children
         [itemData]="itemData"
         *ngIf="itemData.item.type === 'Chapter'"
-        algFullHeightContent
+        [algFullHeightContent]="true"
       ></alg-chapter-children>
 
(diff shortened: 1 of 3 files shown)


smadbe (approved):
👍

PR #956, opened 2022-03-01
On task loading error: "show task" button

### LINKED ISSUE #949: On task loading error: "show task" button
## Overview

When a task loading fails for unknown reason, it may be good for debugging reason (including debugging remotely why it fails on a specific computer) to be able to show the content of the iframe.

## Acceptance criteria
When a task loading fails for unknown reason (the task never announces itself as ready):
- [ ] add to the existing error message a "Show task anyway (for debugging)" blue link which display the task iframe (in place of the error message?... as it is easier for you)

ThomasJuster:
## Description

Fixes #949 

Since error messages are already blue, I used the usual link color (grey, not blue) because otherwise it was not obvious that a user action was available.

## Test cases

- [ ] Case 1:
  2. When I go to [this activity](https://dev.algorea.org/branch/feat/show-errored-task/en/#/activities/by-id/773396873861865944;path=4702,1625159049301502151;parentAttempId=0/details)
  3. And I wait for the task to fail
  4. Then I see the normal error message "Unable to load this task" and the message "Show task anyway"

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this activity](https://dev.algorea.org/branch/feat/show-errored-task/en/#/activities/by-id/773396873861865944;path=4702,1625159049301502151;parentAttempId=0/details)
  3. And I wait for the task to fail
  4. Then I see the two error messages: "It usually occurs when task url is invalid, if the task url is valid and the problem persists, please contact us" and "Show task anyway (for debugging)"
  5. And I click on "Show task anyway"
  6. Then I see the error message is hidden and the iframe is shown

### DIFF
--- a/src/app/modules/item/pages/item-display/item-display.component.html
+++ b/src/app/modules/item/pages/item-display/item-display.component.html
@@ -23,17 +23,18 @@
           i18n-message message="An unknown error occurred. If this problem persists, please contact us."
         ></alg-error>
       </ng-template>
-
-      <div class="text-center">
-        <a class="alg-link error-message" (click)="showTaskAnyway = true" i18n>
-          Show task anyway (for debugging)
-        </a>
-      </div>
     </ng-container>
+
+    <div class="text-center">
+      <a class="alg-link error-message" (click)="showTaskAnyway = true" i18n>
+        Show task anyway (for debugging)
+      </a>
+    </div>
   </div>
   <div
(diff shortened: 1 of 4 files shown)


smadbe:
From a usage test:
- what the difference between the two cases? Why the anonymous user do not have the "Show task anyway"?
- design-wise, that looks very strange: maybe an actual small [p-button](https://www.primefaces.org/primeng/showcase/#/button) would be more appropriate?

smadbe (approved):
👍

PR #957, opened 2022-03-02
refresh left menu

### LINKED ISSUE #951: Update left menu on related content change
## Overview

Some actions on the right pane may create/remove group/items. In such a case, the left menu may need to be reloaded.

## Acceptance criteria
- [ ] On any group change (group info (e.g. `name`) change, delete, join (by code, by invitation, by entering a code on an item), leave, manager remove myself, add subgroup, remove subgroup, ?), refresh the "group tab" of the left menu if already loaded
- [ ] On any item change (any edit, permission change, chapter children change, score change, ?),  refresh the "activity" and "skills" "tabs" of the left menu if already loaded
- [ ] The loading should be in background, the menu should not be blocked by a "loading" state,
- [ ] Keep the left menu as content agnostic as possible (for the parts which are generic)
- [ ] Allow in the future any external notification to notify these changes as well (for instance, if your group join request has been accepted by the manager)

ThomasJuster:
## Description

Fixes #951

Since it's an optimistic update, I have purposedly not handled the error because I considered it was preferable to display the nav with stale data than an error. I can make changes if this behavior is unwanted.

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this group edit page](https://dev.algorea.org/branch/feat/refresh-left-menu/en/#/groups/by-id/672913018859223173;path=52767158366271444/edit)
  3. And I change the group name and save (throttle network if you want)
  4. Then I see the group nav is refreshed after a while, the group name is updated in the nav

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this group edit page](https://dev.algorea.org/branch/feat/refresh-left-menu/en/#/groups/by-id/681265191377636279;path=52767158366271444,672913018859223173,6710944276987033666/details/members)
  3. And I create the subgroup "RefreshNavDemo" (throttle network if you want)
  4. Then I see the group nav is refreshed after a while, "RefreshNavDemo" is added in current group children in the nav
  5. And I click on "sub group" tab and remove "RefreshNavDemo" from the sub-groups list
  6. Then I see the group nav is refreshed after a while, "RefreshNavDemo" is removed from current group children in the nav

- [ ] Case 3:
  1. Given I am the usual user
  2. When I go to [this group edit page](https://dev.algorea.org/branch/feat/refresh-left-menu/en/#/groups/by-id/681265191377636279;path=52767158366271444,672913018859223173,6710944276987033666/details/members)
  3. And I create the subgroup "RefreshNavDemo" (throttle network if you want)
  4. Then I see the group nav is refreshed after a while, "RefreshNavDemo" is added in current group children in the nav
  5. And I navigate to "RefreshNavDemo"
  6. And I edit "RefreshNavDemo"
  7. And I delete "RefreshNavDemo"
  8. Then I see the nav is refreshed while I'm redirected to "My groups"

### DIFF
--- a/src/app/core/services/navigation/nav-tree.service.ts
+++ b/src/app/core/services/navigation/nav-tree.service.ts
@@ -1,8 +1,8 @@
-import { combineLatest, merge, Observable, of, OperatorFunction, Subject } from 'rxjs';
-import { delay, distinctUntilChanged, map, mergeScan, shareReplay, startWith } from 'rxjs/operators';
+import { merge, Observable, of, OperatorFunction, Subject } from 'rxjs';
+import { delay, distinctUntilChanged, map, mergeScan, shareReplay, startWith, withLatestFrom } from 'rxjs/operators';
 import { isDefined } from 'src/app/shared/helpers/null-undefined-predicates';
 import { repeatLatestWhen } from 'src/app/shared/helpers/repeatLatestWhen';
-import { fetchingState, FetchState, readyState } from 'src/app/shared/helpers/state';
+import { errorState, fetchingState, FetchState, readyState } from 'src/app/shared/helpers/state';
 import { ContentInfo, RoutedContentInfo } from 'src/app/shared/models/content/content-info';
 import { mapStateData, mapToFetchState } from 'src/app/shared/operators/state';
 import { CurrentContentService } from 'src/app/shared/services/current-content.service';
@@ -21,13 +21,16 @@ export interface NavigationNeighbors {
 export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
 
   private reloadTrigger = new Subject<void>();
+  protected reload$ = merge(
+    this.reloadTrigger,
+    this.currentContent.reload$,
+  );
 
   private content$ = this.currentContent.content$.pipe( // only keep those of interest for the current nav tree
     // map those which are not of interest to `undefined`
     map(content => (this.isOfContentType(content) ? content : undefined)),
     distinctUntilChanged(), // remove multiple `undefined`
     startWith(undefined),
-    repeatLatestWhen(this.reloadTrigger),
   );
   private children$ = merge(
     // emit `undefined` each time the content change
@@ -36,10 +39,15 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
       distinctUntilChanged(),
       map(() => undefined),
     ),
-    this.content$.pipe(this.childrenNavData())
+    this.content$.pipe(
+      repeatLatestWhen(this.reload$),
+      this.childrenNavData(),
+    ),
   );
-  state$ = combineLatest([ this.children$, this.content$ ]).pipe(
+  state$ = this.children$.pipe(
+    withLatestFrom(this.content$), // When children$ has emitted once, content$ has always emitted at least once too.
     mergeScan((prevState: FetchState<NavTreeData>, [ children, content ]) => {
+      if (children instanceof Error) return of(errorState(children));
 
       // CASE 1: the current-content does not match the type of this nav tree (so `content` has been mapped to `undefined`)
       if (!content) {
@@ -115,7 +123,7 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
   /**
    * Operator which emit the children of a content stream
    */
-  protected abstract childrenNavData(): OperatorFunction<ContentT|undefined,NavTreeElement[]>;
+  protected abstract childrenNavData(): OperatorFunction<ContentT|undefined,NavTreeElement[]|Error>;
 
   /**
    * Re-play the last change
--- a/src/app/core/services/navigation/group-nav-tree.service.ts
+++ b/src/app/core/services/navigation/group-nav-tree.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
-import { EMPTY, Observable, OperatorFunction, pipe } from 'rxjs';
-import { distinctUntilChanged, map, switchMap } from 'rxjs/operators';
+import { EMPTY, of, Observable, OperatorFunction, pipe } from 'rxjs';
+import { catchError, distinctUntilChanged, map, switchMap } from 'rxjs/operators';
+import { repeatLatestWhen } from 'src/app/shared/helpers/repeatLatestWhen';
 import { ContentInfo } from 'src/app/shared/models/content/content-info';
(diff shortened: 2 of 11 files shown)


src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -38,36 +38,45 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
...
-  state$ = combineLatest([ this.children$, this.content$ ]).pipe(
-    mergeScan((prevState: FetchState<NavTreeData>, [ children, content ]) => {
+  state$ = combineLatest([
+    this.children$,
+    this.content$.pipe(startWith(undefined), pairwise()),
+  ]).pipe(
+    mergeScan((prevState: FetchState<NavTreeData>, [ children, [ prevContent, content ] ]) => {
+      const isRefresh = content &&
+        prevContent?.route.contentType === content.route.contentType &&
+        prevContent?.route.id === content.route.id;
```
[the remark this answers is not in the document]
ThomasJuster: I think every comment has been addressed by the changes, can you check ?

src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -38,36 +38,46 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
...
-
+  state$ = combineLatest([
+    this.children$,
+    this.content$.pipe(switchMap(content => this.currentContent.contentHasChanged$.pipe(
+      map(() => true),
+      startWith(false),
+      map(hasChanged => ({ content, hasChanged }))
+    ))),
+  ]).pipe(
+    mergeScan((prevState: FetchState<NavTreeData>, [ children, { content, hasChanged }]) => {
```
smadbe, line 49: If there is a content with "hasChanged=true" and then `children` emit, the next mergeScan block will use `hasChanged=true`and so consider it has to refresh while it should not. Isn't it?

src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -38,36 +38,46 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
...
 
-        if (prevState.isReady && prevState.data.hasElement(route)) {
-          // CASE 2A : the content is among the displayed elements -> either select it if at root or shift the tree "to the left" otherwise
+        if (hasChanged) {
+          // CASE 2A: reload content after content change
+          return this.fetchNewNav(content).pipe(
+            mapStateData(data => this.mapNewNavData(data, children, content)),
+            filter(state => state.isReady),
+            startWith(fetchingState(prevState.data)),
+          );
```
smadbe, line 68: Instead of : 
1. calling `fetchNewNav` which gets the appropriate nav data and maps it to a fetch state, so add loading state when data is not ready 
2. map the data to merge with current knowledge
3. keep only the ready state
4. add a custom fetching state which emits immediately (but what's the use if there is already the same data being displayed in "ready" state)

Why not:
1. calling a `getNewNav` which gets the appropriate nav data
2. map the data to merge with current knowledge
3. wrap the data in a ready state which will just emit the new state when it will be ready
And keep a `fetchNewNav`  which just call `getNewNav` and may it to a fetch state

smadbe (changes requested):

src/app/core/services/navigation/nav-tree.service.ts
ThomasJuster: Fixed
Addressing it also fixed [this comment](https://github.com/France-ioi/AlgoreaFrontend/pull/957#discussion_r817749219)

src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -38,36 +38,46 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
...
   );
-  state$ = combineLatest([ this.children$, this.content$ ]).pipe(
-    mergeScan((prevState: FetchState<NavTreeData>, [ children, content ]) => {
-
+  state$ = combineLatest([
+    this.children$,
+    this.content$.pipe(switchMap(content => this.currentContent.contentHasChanged$.pipe(
+      map(() => true),
+      startWith(false),
+      map(hasChanged => ({ content, hasChanged }))
```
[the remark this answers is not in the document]
ThomasJuster: Fixed by https://github.com/France-ioi/AlgoreaFrontend/pull/957#discussion_r817844970

src/app/core/services/navigation/nav-tree.service.ts
ThomasJuster: Fixed

src/app/core/services/navigation/nav-tree.service.ts
```diff
@@ -25,50 +25,52 @@ export abstract class NavTreeService<ContentT extends RoutedContentInfo> {
...
-  );
-  state$ = combineLatest([ this.children$, this.content$ ]).pipe(
-    mergeScan((prevState: FetchState<NavTreeData>, [ children, content ]) => {
+
+  state$ = this.content$.pipe(
+    switchMap(({ content, hasChanged }) => of(content).pipe(
+      this.childrenNavData(),
+      map(children => ({ children, content, hasChanged })),
+      startWith({ children: undefined, content, hasChanged })
+    )),
```
smadbe, line 43: When I check with what was written before, you clearly change the logic so that cannot be valid. 
Separating the children was not for the sake of complexifying the code, the idea was to have the children fetched only once per content (even when the same content (id) is emitted several time with a build-up).

smadbe (changes requested):
If you modify the way the menu work, you need to include, in your test scenarios, checks that the menu still work as expected, including that no additional requests are done.

src/app/core/services/navigation/nav-tree.service.ts
smadbe: I do not see it fixed.

smadbe: Was missing in my "why not": error handling. 
Maybe allowing to use "mapToFetchState" without initial loading may be useful.

src/app/core/services/navigation/nav-tree.service.ts
ThomasJuster: fixed

smadbe (changes requested):
If I just open the app (as anonymous), I directly see that the menu does not behave as it should. 🤷‍♂️

ThomasJuster:
closed in favor of #963

PR #974, opened 2022-03-25
Remove enum watching

ThomasJuster:
## Description

It was a double bug. With the refactoring of watching mode, I didn't not spot at the review the `Mode` enum still had a `Watching` value that is not used anymore. I removed it to automatically find files relying on it and fixed them. What was broken:
1. Task read-only indicator
2. Item task answer service falling back to "fetch current answer" even when mode is read-only
3. Dead code regarding item reload when switching from normal to watching mode

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this user page](https://dev.algorea.org/branch/remove-enum-watching/en/#/groups/users/752024252804317630)
  4. And I observe the user
  5. And I navigate to "Blockly basic task" in the left nav
  6. Then I see the task is loaded in read-only, no answer is loaded
  7. And I exit observe mode
  8. Then I see the content is reloaded (no regression)

### DIFF
--- a/src/app/modules/item/pages/item-by-id/item-by-id.component.ts
+++ b/src/app/modules/item/pages/item-by-id/item-by-id.component.ts
@@ -147,10 +147,6 @@ export class ItemByIdComponent implements OnDestroy {
       .subscribe(() => this.layoutService.toggleFullFrameContent(false));
   }
 
-  reloadContent(): void {
-    this.fetchItemAtRoute(this.activatedRoute.snapshot.paramMap);
-  }
-
   private getItemRoute(params?: ParamMap): ReturnType<typeof itemRouteFromParams> {
     const snapshot = this.activatedRoute.snapshot;
     if (!snapshot.parent) throw new Error('Unexpected: activated route snapshot has no parent');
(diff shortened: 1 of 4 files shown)


src/app/modules/item/pages/item-by-id/item-by-id.component.ts
```diff
@@ -125,11 +125,6 @@ export class ItemByIdComponent implements OnDestroy {
         this.itemRouter.navigateTo(current.route, { page: action === ModeAction.StartEditing ? 'edit' : 'details' });
       }),
 
-      this.modeService.mode$.pipe(
-        filter(mode => [ Mode.Normal, Mode.Watching ].includes(mode)),
-        distinctUntilChanged(),
-      ).subscribe(() => this.reloadContent()),
-
```
ThomasJuster, line 132: That part (which was not working anymore) is not required since the addition of [this code in item-datasource](https://github.com/France-ioi/AlgoreaFrontend/blob/master/src/app/modules/item/services/item-datasource.service.ts#L53)

src/app/modules/item/services/item-task-answer.service.ts
```diff
@@ -51,7 +51,7 @@ export class ItemTaskAnswerService implements OnDestroy {
         return readOnly ? of(formerAnswer) : this.loadFormerAsNewCurrentAnswer(route.id, attemptId, formerAnswer);
       }
 
-      return this.getCurrentAnswer(route.id, attemptId);
+      return readOnly ? of(formerAnswer) : this.getCurrentAnswer(route.id, attemptId);
```
ThomasJuster, line 54: When `readOnly`, `formerAnswer` will be `null` but if someday it's not the service can already handle it this way.

smadbe (approved):
👍

PR #983, opened 2022-04-01
Fix bug: only emit error only when the first connection is not made

### LINKED ISSUE #981: Error displayed on "successful" task
**Describe the bug**
As shown on this screenshot, the error message is shown while the task has been loaded successfully:
Screenshot 2022-03-30 at 11 34 26

**To Reproduce**
1. As an anonymous user
2. I visit [this task](https://dev.algorea.org/en/#/activities/by-id/837046206729127555;path=4702,1352246428241737349,314613032161178344;parentAttempId=0/details)
3. It loads successfully
4. And wait for a timeout (20-30sec)
5. The error message appears in addition to the content

smadbe:
**Urgent fix.** 
Fix bug: only emit error only when the first connection is not emitted. Was expecting regular emission.

Fixes #981

### DIFF
--- a/src/app/modules/item/services/item-task-init.service.ts
+++ b/src/app/modules/item/services/item-task-init.service.ts
@@ -54,7 +54,7 @@ export class ItemTaskInitService implements OnDestroy {
 
   readonly initError$ = this.configFromIframe$.pipe(switchMap(({ iframe }) => fromEvent(iframe, 'load'))).pipe(
     switchMap(() => this.task$),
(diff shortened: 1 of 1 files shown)


smadbe:
cc @ThomasJuster

smadbe:
@ThomasJuster : `timeout(...)` without named parameter is for `each` so expect an emission every "period" while you only wanted to check the first one for your case.   I didn't know either.

Was introduced by #977 ... as error was appearing on successful task after the timeout, the tests did show the problem (as you/I probably just open the successful task, check it worked and closed it)

PR #985, opened 2022-04-04
Feature/path suggestion overlay

### LINKED ISSUE #615: Path suggestion overlay
## Overview
In some cases (such as group progress), items are listed out of their context. In such cases: 
- the link pick a random path among paths to the item (if several). It might not be the one expected by the user.
- the user may wonder what's the context of this item (for instance "Task 1" ... but from what chapter?)

To solve these 2 issues, display an overlay on hover of such items with the possible paths to it.

## Acceptance criteria
- [ ] On hover of items listed in "log view", display an overlay (similar to the one for observation: 
Screenshot 2021-08-18 at 16 21 31) after a small delay.
- [ ] This overlay displays a loader and then a list of paths (using the [itemBreadcrumbsFromRootsGet](https://france-ioi.github.io/algorea-devdoc/api/#operation/itemBreadcrumbsFromRootsGet) service) to the item in the form of a breadcrumb.
- [ ] The breadcrumb are shown as "via ItemA1/ItemA2" with appropriate ellipses when the titles are too long.
- [ ] If there is no path, display "You have no path to this content." (actually it is not supposed to happen).
- [ ] The breadcrumb should be clickable and navigate to the hovered item with the selected path (and no attempt)
- [ ] If the service returns an error, an error should be displayed in the overlay
- [ ] If leaving the hover before the end of the loading, make sure the request is cancelled!

Iloveall:
## Description

Fixes #615

While BE going to add type to the root breadcrumbs response, we can consider all items are activity.

I've found that mapToFetchState doesn't change isFetching field to true for second request, you can see this affect when change hover from row to row in table. I can add merge fetchingState but not sure if it's correct and only one way. Let me know what you think

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/path-suggestion-overlay/en/#/groups/by-id/4322568739774621128;path=52767158366271444/details)
  3. And hover on one of row table
  4. Then I see overlay with path suggestion

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/path-suggestion-overlay/en/#/groups/by-id/49268055312628123;path=/details)
  3. And hover on one row with "Locked content"
  4. Then I see overlay with path suggestion with error message: "You are not allowed to view path suggestion."

### DIFF
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.html
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.html
@@ -37,7 +37,7 @@
           let-columns="columns"
           let-index="rowIndex"
       >
-        <tr (mouseenter)="onMouseEnter($event, rowData.item.id, index)" (mouseleave)="onMouseLeave()">
+        <tr>
           <td *ngFor="let col of columns">
             <ng-container [ngSwitch]="col.field">
               <ng-container *ngSwitchCase="'activityType'">
@@ -48,6 +48,8 @@
                   class="alg-link"
                   [ngClass]="{'disabled': !rowData.item}"
                   [routerLink]="rowData.item | rawItemLink"
+                  (mouseenter)="onMouseEnter($event, rowData.item.id, index)"
+                  (mouseleave)="onMouseLeave()"
                   #contentRef
                 >
                   {{ rowData.item.string.title }}
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.html
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.html
@@ -78,8 +78,8 @@
 </ng-container>
 
 <p-overlayPanel
+    styleClass="alg-path-suggestion-overlay"
     #op
-    [style]="{ minWidth: '30rem', maxWidth: '50rem' }"
 >
   <ng-container *ngIf="op.overlayVisible">
     <alg-path-suggestion [itemId]="(showOverlay$ | async)?.itemId"></alg-path-suggestion>
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.html
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.html
@@ -37,7 +37,7 @@
           let-columns="columns"
           let-index="rowIndex"
       >
-        <tr (mouseenter)="onMouseenter($event, rowData.item.id, index)" (mouseleave)="onMouseleave()">
+        <tr (mouseenter)="onMouseEnter($event, rowData.item.id, index)" (mouseleave)="onMouseLeave()">
           <td *ngFor="let col of columns">
             <ng-container [ngSwitch]="col.field">
               <ng-container *ngSwitchCase="'activityType'">
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.html
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.html
@@ -38,7 +38,7 @@
           let-index="rowIndex"
       >
         <tr>
-          <td *ngFor="let col of columns">
+          <td *ngFor="let col of columns" (mouseleave)="onMouseLeave($event, col.field)">
             <ng-container [ngSwitch]="col.field">
               <ng-container *ngSwitchCase="'activityType'">
                 {{ rowData.activityType | logActionDisplay : rowData.score }}
@@ -49,7 +49,6 @@
                   [ngClass]="{'disabled': !rowData.item}"
                   [routerLink]="rowData.item | rawItemLink"
                   (mouseenter)="onMouseEnter($event, rowData.item.id, index)"
-                  (mouseleave)="onMouseLeave()"
                   #contentRef
                 >
                   {{ rowData.item.string.title }}
@@ -84,6 +83,6 @@
     #op
 >
   <ng-container *ngIf="op.overlayVisible">
-    <alg-path-suggestion [itemId]="(showOverlay$ | async)?.itemId"></alg-path-suggestion>
+    <alg-path-suggestion [itemId]="(showOverlay$ | async)?.itemId" (mouseleave)="closeOverlay()"></alg-path-suggestion>
   </ng-container>
 </p-overlayPanel>
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.html
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.html
@@ -35,15 +35,21 @@
           pTemplate="body"
           let-rowData
           let-columns="columns"
+          let-index="rowIndex"
       >
-        <tr>
+        <tr (mouseenter)="onMouseenter($event, rowData.item.id, index)" (mouseleave)="onMouseleave()">
           <td *ngFor="let col of columns">
             <ng-container [ngSwitch]="col.field">
               <ng-container *ngSwitchCase="'activityType'">
                 {{ rowData.activityType | logActionDisplay : rowData.score }}
               </ng-container>
               <ng-container *ngSwitchCase="'item.string.title'">
-                <a class="alg-link" [ngClass]="{'disabled': !rowData.item}" [routerLink]="rowData.item | rawItemLink">
+                <a
+                  class="alg-link"
+                  [ngClass]="{'disabled': !rowData.item}"
+                  [routerLink]="rowData.item | rawItemLink"
+                  #contentRef
+                >
                   {{ rowData.item.string.title }}
                 </a>
               </ng-container>
@@ -70,3 +76,12 @@
     </p-table>
   </ng-container>
 </ng-container>
+
+<p-overlayPanel
+    #op
+    [style]="{ minWidth: '30rem', maxWidth: '50rem' }"
+>
+  <ng-container *ngIf="op.overlayVisible">
+    <alg-path-suggestion [itemId]="(showOverlay$ | async)?.itemId"></alg-path-suggestion>
+  </ng-container>
+</p-overlayPanel>
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.ts
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.ts
@@ -47,7 +47,7 @@ export class GroupLogViewComponent implements OnChanges, OnInit, OnDestroy {
     mapToFetchState({ resetter: this.refresh$ }),
   );
   private readonly showOverlaySubject$ = new BehaviorSubject<{ event: Event, itemId: string, target: HTMLElement }|undefined>(undefined);
-  showOverlay$ = this.showOverlaySubject$.asObservable().pipe(debounceTime(750), shareReplay());
+  showOverlay$ = this.showOverlaySubject$.asObservable().pipe(debounceTime(750), shareReplay(1));
 
   constructor(
     private activityLogService: ActivityLogService,
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.ts
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.ts
@@ -9,8 +9,8 @@ import {
   ViewChild,
   ViewChildren
 } from '@angular/core';
-import { BehaviorSubject, debounceTime, Observable, ReplaySubject, Subject } from 'rxjs';
-import { filter, map, shareReplay, switchMap, takeUntil } from 'rxjs/operators';
+import { BehaviorSubject, debounceTime, Observable, ReplaySubject, Subject, Subscription } from 'rxjs';
+import { filter, map, shareReplay, switchMap } from 'rxjs/operators';
 import { mapToFetchState } from '../../../../shared/operators/state';
 import { ActivityLog, ActivityLogService } from '../../../../shared/http-services/activity-log.service';
 import { isNotUndefined } from '../../../../shared/helpers/null-undefined-predicates';
@@ -39,7 +39,8 @@ export class GroupLogViewComponent implements OnChanges, OnInit, OnDestroy {
   @ViewChild('op') op?: OverlayPanel;
   @ViewChildren('contentRef') contentRef?: QueryList<ElementRef<HTMLElement>>;
 
-  private readonly unsubscribe$ = new Subject<void>();
+  showOverlaySubscription?: Subscription;
+
   private readonly groupId$ = new ReplaySubject<string>(1);
   private readonly refresh$ = new Subject<void>();
   readonly state$ = this.groupId$.pipe(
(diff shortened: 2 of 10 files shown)


ThomasJuster (commented):
image

The path is not readable at all, but I doubt that there's actually a way to make sure it _is_ readable with such a popover. Although we could give each breadcrumb a wider width, we will always have cases of deeply nested activities with  too many parents to render correctly.
To me it doesn't makes sense to provide a UI like that to an end user. I'll think about what other UI/widget would do be appropriate, either using a popover or something else.

@smadbe what do you think about this statement? Should we stick to this popover widget?

smadbe:
> image
> 
> The path is not readable at all, but I doubt that there's actually a way to make sure it _is_ readable with such a popover. 

How do you reached this page? It looks you have a loop in your items which is normally impossible. 
It looks like the concatenation of 2 readable breadcrumbs. What should be displayed has two lines not including the last "Les tests" as this is the already visible element.

Ok, so this example will not really happen in practice...

Then we can decide to always display the tail (not including the searched item itself) and truncate the head. This is the most important part. 

Probably there will be other improvements to be done once I saw what it gives in practice.

src/app/modules/group/components/group-log-view/group-log-view.component.html
```diff
@@ -35,15 +35,21 @@
           pTemplate="body"
```
ThomasJuster, line 35: > The breadcrumb are shown as "via ItemA1/ItemA2" with appropriate ellipses when the titles are too long.

"Via [breadcrumb]" -> the "via" is missing

ThomasJuster, line 35: > The breadcrumb should be clickable and navigate to the hovered item with the selected path (and no attempt)

The breadcrumb is a link, but no path is given. For instance: "https://dev.algorea.org/branch/feature/path-suggestion-overlay/en/#/activities/by-id/369645559920184960"

ThomasJuster (commented):
I'll stick to the UI review before diving into code.

src/app/modules/group/components/group-log-view/group-log-view.component.html
Iloveall: Fixed

Iloveall: Fixed

Iloveall:
Also I've add one more test case

src/app/modules/shared-components/components/path-suggestion/path-suggestion.component.ts
```diff
@@ -0,0 +1,66 @@
...
+  state$ = this.itemId$.pipe(
+    switchMap(itemId => this.getBreadcrumbsFromRootsService.get(itemId).pipe(
+      map(group => {
+        if (group.length === 0) {
+          return undefined;
+        }
+
+        const allBreadcrumbs = group
+          .map(items => items.slice(0, items.length - 1))
+          .reduce((state, items) => [ ...state, ...items ], []);
```
ThomasJuster, line 32: FYI: This is the equivalent as the `[].flat(1)` array function I think. [See MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)

src/app/modules/shared-components/components/path-suggestion/path-suggestion.component.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        if (allBreadcrumbs.length === 0) {
+          return [];
+        }
```
ThomasJuster, line 36: I think these lines don't change anything. If I'm correct, the whole point of these lines is to return an empty array when all breadrcumbs are empty.
To me, this case is already handled on the return line: if after applying `.map(...).filter(...)` you had only empty breadcrumbs, it will also return an empty array.

src/app/modules/group/components/group-log-view/group-log-view.component.html
```diff
@@ -35,15 +35,21 @@
...
           let-rowData
           let-columns="columns"
+          let-index="rowIndex"
       >
-        <tr>
+        <tr (mouseenter)="onMouseenter($event, rowData.item.id, index)" (mouseleave)="onMouseleave()">
```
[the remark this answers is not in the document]
Iloveall: Fixed

src/app/modules/group/components/group-log-view/group-log-view.component.html
```diff
@@ -70,3 +76,12 @@
     </p-table>
   </ng-container>
 </ng-container>
+
+<p-overlayPanel
+    #op
+    [style]="{ minWidth: '30rem', maxWidth: '50rem' }"
```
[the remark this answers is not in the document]
Iloveall: Fixed

src/app/modules/shared-components/components/path-suggestion/path-suggestion.component.ts
Iloveall: App tsconfig.json lib settings is `es2018`, `flat` requires es2019

src/app/modules/shared-components/components/path-suggestion/path-suggestion.component.ts
Iloveall: Base on conversation with Damien:

(empty path != no path). In case of empty path, you can display "This content is one of your root activity/skill"

There must be 2 condition:

1. If API returns empty (no path): You have no path to this content.
2. If API returns just 1 element, which is current and which is sliced (empty path): This content is one of your root activity/skill

Iloveall: Fixed

src/app/modules/shared-components/components/path-suggestion/path-suggestion.component.ts
Iloveall: Anyway next fix will remove it

src/app/modules/group/components/group-log-view/group-log-view.component.html
Iloveall: Hmm, shall I create pipe for that?

ThomasJuster: I think constructing the url (using item-router) in the component is enough, you can map over breadcrumbs and add a new property. I was thinking [here](https://github.com/France-ioi/AlgoreaFrontend/pull/985/files#diff-6664e60575412437d7dd0a3a95ee8bb818194e7f0736f4ed3e1c6bfe6c3358c5R26)

src/app/modules/group/components/group-log-view/group-log-view.component.ts
```diff
@@ -19,22 +31,34 @@ interface Data {
...
   constructor(
     private activityLogService: ActivityLogService,
   ) {}
 
+  ngOnInit(): void {
+    this.showOverlay$.pipe(filter(isNotUndefined), takeUntil(this.unsubscribe$)).subscribe(data =>
+      this.op?.toggle(data.event, data.target)
+    );
+  }
+
```
ThomasJuster, line 61: Sorry I don't know where it went, I actually added a comment previously on these lines... :confused: 

Here `.pipe(..., takeUntil(unsubscribe$).subscribe(...)` looks like a subscription you want to unsubscribe from on delete. For that you can directly store the subscription and unsubscribe on delete:
```ts
private showOverlaySubscription = this.showOverlay$.pipe(filter(...)).subscribe(...);

ngOnDelete() {
  ...
  this.showOverlaySubscription.unsubscribe();
}
```

And you can remove the property `unsubscribe$` :grin:

ThomasJuster (commented):
Those will be my final comments, @smadbe you can take a look

src/app/modules/group/components/group-log-view/group-log-view.component.html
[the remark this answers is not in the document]
smadbe: I agree with Thomas. I would except as well "the overlay to open only when I hover the task title"

smadbe (changes requested):
2 improvements in terms of UI:
- as Thomas mentioned, the overlay should only load when hovering the item title, not the rest of row
- when the cursor is leaving the item title text (and not in the overlay itself), the overlay should immediately close

src/app/modules/group/components/group-log-view/group-log-view.component.html
Iloveall: Fixed

src/app/modules/group/components/group-log-view/group-log-view.component.ts
Iloveall: Fixed

src/app/modules/group/components/group-log-view/group-log-view.component.ts
```diff
@@ -19,22 +31,35 @@ interface Data {
...
+  private readonly showOverlaySubject$ = new BehaviorSubject<{ event: Event, itemId: string, target: HTMLElement }|undefined>(undefined);
+  showOverlay$ = this.showOverlaySubject$.asObservable().pipe(debounceTime(750), shareReplay(1));
 
   constructor(
     private activityLogService: ActivityLogService,
   ) {}
 
+  ngOnInit(): void {
+    this.showOverlaySubscription = this.showOverlay$.pipe(filter(isNotUndefined)).subscribe(data =>
+      this.op?.toggle(data.event, data.target)
```
smadbe, line 59: If you want your `mouseLeave` doing `this.showOverlaySubject$.next(undefined);` to have an effect, you do not have to filter out the undefined but just to `this.op?.show(...)` when not-undefined and `this.op?.hide()` when undefined.

src/app/modules/group/components/group-log-view/group-log-view.component.ts
```diff
@@ -88,4 +115,42 @@ export class GroupLogViewComponent implements OnChanges, OnDestroy {
...
+          closeOverlay = false;
+          break;
+        }
+        currentElement = currentElement.parentElement;
+      }
+    }
+
+    if (closeOverlay) {
+      this.closeOverlay();
+    }
```
ThomasJuster, line 148: I think this code will become hard to maintain, there should be a more expressive/declarative way to write it.
If my understanding is correct, given that the overlay is a singleton (same one is reused for each cell), we have the following cases for a specific cell:
1. No overlay is opened **or** overlay is opened for another cell
   - mouse leaves cell: noop
   - mouse enters cell: trigger a "path suggestion" call and show overlay for current cell
2. An overlay is opened for current cell
   - mouse leaves current cell and does not enter overlay: close overlay
   - mouse leaves current cell and enters overlay: noop
   - mouse leaves overlay and enters current cell: noop

This is obviously non-trivial, but rx might help us here :thinking: 
Here's a draft that might help, what do you think?
```ts
closeOverlaySubscription?: Subscription // to be cleared on destroy as well

onMouseEnter(event, itemId, index) {
    // clear previous subscription if any
    this.closeOverlaySubscription?.unsubscribe();

  // current treatment. fetch etc.
  // and at the end: (this code is designed to entirely replace the `onMouseLeave()` method)

  // add a requestAnimationFrame to give time to the overlay to get rendered
  // Might not be needed, to be checked!
  requestAnimationFrame(() => {
    this.closeOverlaySubscription = merge(
      // mouseleave event is triggered before mouseenter of another target
      fromEvent(this.op, 'mouseleave').pipe(map(() => true)), // close overlay on mouse leave overlay
      fromEvent(event.target, 'mouseleave').pipe(map(() => true)), // close overlay on mouse leave cell
      fromEvent(this.op, 'mouseenter').pipe(map(() => false)), // cancels close when mouse enters overlay
      fromEvent(event.target, 'mouseenter').pipe(map(() => false)), // cancels close when mouse (re)enters current cell
    ).pipe(
      // consider only latest value in a time window of 500ms so that:
      // if the user leaves the cell to enter the overlay within 500ms, the last `shouldClose` value will be `false`
      // if the user leaves the overlay to come back to the cell within 500ms, the last `shouldClose value will be `false`
      debounceTime(500),
      filter(shouldClose => shouldClose === true),
    ).subscribe(() => this.closeOverlay())
  })
}
```

src/app/modules/group/components/group-log-view/group-log-view.component.ts
ThomasJuster: The idea is to consider also undefined values (thus removing `.pipe(filter(isNotUndefined))`) and that an undefined value _means_ "close overlay" while a defined value _means_ "open overlay", in code terms:
- if data is defined, call `this.op.show(data.event, data.target)` (does this method exist?)
- if data is undefined, call `this.op.hide()`

And since `undefined` values are handled here, to close the overlay we can do `this.showOverlaySubject$.next(undefined)` instead of `this.op.hide()` (in the `closeOverlaly` method)

ThomasJuster (commented):
@smadbe won't have much time to review so I'll be taking over from here.

src/app/modules/group/components/group-log-view/group-log-view.component.ts
Iloveall: - On hover of items listed in "log view", display an overlay after a small delay.

`filter(isNotUndefined)` needs for that delay - `debounceTime(750)`. The improvement offered you means delay for close as well, which is not fit to the task

src/app/modules/group/components/group-log-view/group-log-view.component.ts
Iloveall: I've tried to use this way, but after spend a lot of time still can't get stabile work, it makes a lot additional problems. Actually I've spend a lot of time to get stable work for current implementation, so really not sure do we need it right now, otherwise feature will became too expensive.

src/app/modules/group/components/group-log-view/group-log-view.component.html
```diff
@@ -35,15 +35,22 @@
           pTemplate="body"
```
ThomasJuster, line 35: UI issue reported by @smadbe 
The text should not be larger here
screenshot

ThomasJuster, line 35: UI issue reported by @smadbe 
This case seems not to be handled correctly in terms of styling: (1st row of 1st test case)
screenshot

src/app/modules/group/components/group-log-view/group-log-view.component.ts
ThomasJuster: See #1030 proposal

src/app/modules/group/components/group-log-view/group-log-view.component.ts
ThomasJuster: I agree, intensive usage of rxjs will not make it easier in the future, and it will probably make it even harder.

The part I found hard to maintain or to clearly see what it does is the `while`+`if` loop, and to me it can be replaced by the `Node.closest(selector)` method, see #1030 proposal.

src/app/modules/group/components/group-log-view/group-log-view.component.html
Iloveall: Fixed

ThomasJuster (commented):
To me code is fine :+1:
For UI I wonder if horizontal padding aren't too big, UI review is more for @smadbe

PR #987, opened 2022-04-05
Version injection

smadbe:
## Description

- Inject the generated version number in code through an auto-generated file. (less worst solution)
- Also remove the version from the package.json +lock files from DRYness and as they were useless.
- Inject this version in Sentry to that sentry gets the correct version number of the app
- Make sure dev deployments have automated identifiable version numbers so that sentry bug report does not become a mess -- Also allow to verify that the currently running version of an app include a commit we've just pushed

## Test cases

- Launch the [app](https://dev.algorea.org/branch/version-injection/en/#/activities/by-id/4702;path=;parentAttempId=0/details), with debug console open
- The app version is displayed at launch 
- The app version is sent to Senty (checked ! 👍 )


src/app/core/app.component.ts
```diff
@@ -60,6 +61,8 @@ export class AppComponent implements OnInit, OnDestroy {
     if (appConfig.theme !== 'default') {
       this.renderer.setAttribute(this.el.nativeElement, 'data-theme', `${appConfig.theme}`);
     }
+    // eslint-disable-next-line no-console
+    console.log(`App version: ${version}`);
```
ThomasJuster, line 65: Angular keeps its version as an html attribute `ng-version="13.1.2"`, we could do the same with `alg-version=...`

.release-it.yml
```diff
@@ -3,7 +3,8 @@ git:
   changelog: npx auto-changelog --stdout --commit-limit 0 --unreleased --template https://raw.githubusercontent.com/release-it/release-it/master/templates/changelog-compact.hbs
 github:
   release: true
-npm:
-  publish: false
+npm: false
 hooks:
-  after:bump: npx auto-changelog --package --starting-date 2022-02-15 --commit-limit 0 --tag-pattern 'v\d+.*'
+  after:bump: |
+    npx auto-changelog --package --starting-date 2022-02-15 --commit-limit 0 --tag-pattern 'v\d+.*';
```
ThomasJuster, line 9: `--commit-limit 0` means no commit is added to changelog?

.circleci/config.yml
```diff
@@ -51,6 +51,7 @@ commands:
       path:
         type: string
     steps:
+      - run: echo "export const version = '`git describe --always --dirty`';" > ./srv/version.ts
```
ThomasJuster: I think there's a typo: `./srv/...`
I don't really get the interest of removing the version from the package.json to add it in the src folder, why do you want to do it this way?

.release-it.yml
```diff
@@ -3,7 +3,8 @@ git:
...
+    echo "// This file is auto-generated by the release script\nexport const version = 'v${version}';" > src/version.ts
```
ThomasJuster, line 10: Since src/version.ts is a versioned file, shouldn't it be commited after being generated?

.circleci/config.yml
smadbe: You were not marked as reviewer anymore while I was testing test ;-)

.release-it.yml
smadbe: Yes. But there is no change to this line.

src/app/core/app.component.ts
smadbe: What for? Instead of printing it? 
The print was a kind of easy way to ask the customer for the version while we do something better with it in a future PR (e.g. display it somewhere in small).

.circleci/config.yml
smadbe: There is no way to get the package.json version in live code in a clean, safe way (without exporting everything for the package.json) and compatible with a decent release process.

smadbe: And as I added in the PR description, so that we "Make sure dev deployments have an automated identifiable version number so that sentry bug report does not become a mess"

.release-it.yml
smadbe: it is, this is how release-it works

smadbe:
Description updated

smadbe:
We can rediscuss that in a call if you prefer.

src/app/core/app.component.ts
ThomasJuster: Because sometimes when you have to many logs it might be drowned in the logs, or one might clear logs which makes retrieving the version no longer possible, scenarios like that.
Having it as an attribute on `alg-root` or `html` makes it readable at first glance on the element panel, no matter how long you've been browsing the app.
Anyway it was only a suggestion, if there's still no interest in your opinion it's fine by me.

.circleci/config.yml
ThomasJuster: I found a way to inject env vars with webpack `DefinePlugin` ([article[(https://medium.com/@fidelisclayton/system-environment-variables-in-angular-1f4a922c7b4c)), it might be cleaner but it's too overkill in my opinion, your call.

ThomasJuster (approved):

.circleci/config.yml
smadbe: It is a quite old post, it is not working as is, so I'll give up this option for now, even it looks quite a good idea.

PR #990, opened 2022-04-05
Adds refresh button for task when failed

### LINKED ISSUE #786: Add retry button on task error
## Overview

As task has more complicated approach, implement refresh task button with the way described in slack by Thomas

## Acceptance criteria

https://france-ioidevs.slack.com/archives/C02DSRGJLP2/p1634562425006900?thread_ts=1634560440.006200&cid=C02DSRGJLP2

Iloveall:
## Description

Fixes #786

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/add-refresh-button-for-tasks/en/#/activities/by-id/6379723280369399253;path=;attempId=0/details)
  3. In network tab of browser switch to slow 3g
  4. While loading the task, switch slow 3g to offline
  5. Then I see error message with refresh button
  6. Switch offline mode to on
  7. Click on refresh button
  8. Then I see the task

### DIFF
--- a/src/app/modules/item/pages/item-display/item-display.component.html
+++ b/src/app/modules/item/pages/item-display/item-display.component.html
@@ -27,7 +27,7 @@
       </ng-template>
     </ng-container>
 
-    <div class="text-center error-buttons">
+    <div class="error-buttons">
       <alg-button
         class="error-button"
         i18n-label label="Show task anyway (for debugging)"
--- a/src/app/modules/item/pages/item-display/item-display.component.html
+++ b/src/app/modules/item/pages/item-display/item-display.component.html
@@ -3,8 +3,6 @@
     <ng-template #defaultErrorMessage>
       <alg-error
         i18n-message message="Unable to load the task"
-        [showRefreshButton]="true"
-        (refresh)="refresh.emit()"
       ></alg-error>
     </ng-template>
 
@@ -12,8 +10,6 @@
       <alg-error
         *ngIf="initError$ | async; else urlError"
         i18n-message message="It usually occurs when task url is invalid. If the task url is valid and the problem persists, please contact us."
-        [showRefreshButton]="true"
-        (refresh)="refresh.emit()"
       ></alg-error>
 
       <ng-template #urlError>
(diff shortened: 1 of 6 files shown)


src/app/modules/item/pages/item-display/item-display.component.html
```diff
@@ -8,6 +8,8 @@
       <alg-error
         *ngIf="initError$ | async; else urlError"
         i18n-message message="It usually occurs when task url is invalid. If the task url is valid and the problem persists, please contact us."
+        [showRefreshButton]="true"
+        (refresh)="refresh.emit()"
       ></alg-error>
```
ThomasJuster, line 13: The errors with refresh buttons are only displayed to authenticated users that can edit the task (and it works well by the way :+1: ).
I think we should also display that reload button in the default error message for other users (a bit upper in the code)

ThomasJuster (commented):
Works well, one small comment to add a use case and I think we're good to go :+1:

smadbe:
Screenshot 2022-04-08 at 15 16 49
I suppose that's the small refresh icon on the right of the error message?
If so, as this "show task anyway" button has been added in the meantime, I would add a button with "Retry" as text on the left of the "show task anyway" button instead.

src/app/modules/item/pages/item-display/item-display.component.html
```diff
@@ -25,11 +27,17 @@
       </ng-template>
     </ng-container>
 
-    <div class="text-center">
+    <div class="text-center error-buttons">
```
ThomasJuster, line 30: I think the class `text-center` has no effect anymore since error-buttons has justify-content:center.

ThomasJuster (approved):
Can you just take a look at my final comment and make (if relevant) before merge :pray: ?

src/app/modules/item/pages/item-display/item-display.component.html
Iloveall: Fixed

PR #995, opened 2022-04-12
handle refresh token error

### LINKED ISSUE #989: Prevent Senty "report issue" panel when token expires
## Overview
When left open for a long time (e.g. computer reopened), sessions (typically anonymous ones as there are 2h-long I think) cause a 401 when trying the extend the cookie lifetime and it is too late. The behavior is normal but the a Sentry modal is displayed in this case, which should not be the case.

## Acceptance criteria
- [ ] Don't display the sentry modal when an anonymous session expires and receive a 401.

ThomasJuster:
## Description

Fixes #989

> Make sure HTTP response errors are considered as Error

In the end this part is not handled because it needs further investigation without adding much value to the platform.

## Test cases

- [ ] Case 1: the reproduced bug
  2. When I visit the [app](https://dev.algorea.org/branch/reproduce-refresh-token-bug/en/#/activities/by-id/4702;path=;parentAttempId=0/details)
  3. And I quickly remove my authentication token http-only cookie in less than 5s (feasible in firefox, not in chrome)
  4. Then I see after 5s max the sentry modal appear

- [ ] Case 2: fix the reproduced bug
  2. When I visit the [app](https://dev.algorea.org/branch/fix-refresh-token-xhr/en/#/activities/by-id/4702;path=;parentAttempId=0/details)
  3. And I quickly remove my authentication token http-only cookie in less than 5s (feasible in firefox, not in chrome)
  4. Then I see after 5s max the sentry modal still does __not__ appear, the platform is loaded normally
  5. In the network logs, I see no 401 because the refresh call is omitted, and I see a temp-user call which corresponds to invalidating the token.

### DIFF
--- a/src/app/shared/auth/auth.service.ts
+++ b/src/app/shared/auth/auth.service.ts
@@ -15,6 +15,7 @@ import {
   forcedTokenAuthFromStorage
 } from './auth-info';
 import { LocaleService } from 'src/app/core/services/localeService';
+import { HttpErrorResponse } from '@angular/common/http';
 
 // Lifetime under which we refresh the token.
 export const minTokenLifetime = 5*MINUTES;
@@ -85,7 +86,11 @@ export class AuthService implements OnDestroy {
         const tokenIsExpired = auth.expiration.valueOf() < Date.now();
         if (tokenIsExpired) return of({ auth, tokenIsExpired }); // don't even try to refresh the token if it is expired
         return this.authHttp.refreshAuth(auth).pipe(
-          catchError(() => EMPTY), // ignore the error since the token is valid. Another try will occur the next minute.
+          catchError(err => {
+            // For any http error, ignore it since the token is valid. Another try will occur the next minute.
+            if (err instanceof HttpErrorResponse) return EMPTY;
+            else throw err; // rethrow any JS error to let our error monitor catch it.
+          }),
           map(auth => ({ auth, tokenIsExpired })),
         );
       })
--- a/src/app/shared/auth/auth.service.ts
+++ b/src/app/shared/auth/auth.service.ts
@@ -82,15 +82,15 @@ export class AuthService implements OnDestroy {
         return timer(Math.min(refreshIn, maxDelay), 1*MINUTES).pipe(map(() => auth));
       }),
       switchMap(auth =>
-        this.authHttp.refreshAuth(auth)
+        this.authHttp.refreshAuth(auth).pipe(
+          catchError(() => {
+            this.invalidToken(auth);
+            return EMPTY;
+          })
+        )
       )
-    ).subscribe({
-      next: auth => {
-        this.status$.next(auth);
-      },
-      error: () => {
-        this.failure$.next();
-      },
+    ).subscribe(auth => {
+      this.status$.next(auth);
     });
   }
 
--- a/src/app/shared/auth/auth.service.ts
+++ b/src/app/shared/auth/auth.service.ts
@@ -83,15 +83,15 @@ export class AuthService implements OnDestroy {
         return timer(Math.min(refreshIn, maxDelay), 1*MINUTES).pipe(map(() => auth));
       }),
       switchMap(auth => {
-        const tokenIsExpired = auth.expiration.valueOf() < Date.now();
-        if (tokenIsExpired) return of({ auth, tokenIsExpired }); // don't even try to refresh the token if it is expired
+        const isExpired = auth.expiration.valueOf() < Date.now();
+        if (isExpired) return of({ auth, tokenIsExpired: true }); // don't even try to refresh the token if it is expired
         return this.authHttp.refreshAuth(auth).pipe(
           catchError(err => {
             // For any http error, ignore it since the token is valid. Another try will occur the next minute.
             if (err instanceof HttpErrorResponse) return EMPTY;
             else throw err; // rethrow any JS error to let our error monitor catch it.
           }),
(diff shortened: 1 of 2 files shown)


src/app/shared/auth/auth.service.ts
```diff
@@ -82,7 +82,12 @@ export class AuthService implements OnDestroy {
         return timer(Math.min(refreshIn, maxDelay), 1*MINUTES).pipe(map(() => auth));
       }),
       switchMap(auth =>
-        this.authHttp.refreshAuth(auth)
+        this.authHttp.refreshAuth(auth).pipe(
+          catchError(() => {
+            this.invalidToken(auth);
```
smadbe, line 87: so, side effect and returning empty? 🤨
Why not just handling the errors in the subscribe?

And so you are considering all error as "the token is invalid, let's drop it disconnect the user, ..."... disregarding what the error is? That's not good. Here we try to renew a token before it expires (in the normal case)... so it is not because we receive an error that the token in invalid... if it is still valid and we are just unable to extend (refresh) it, let's just use it to the end instead of dropping it.

smadbe (changes requested):

src/app/shared/auth/auth.service.ts
ThomasJuster: If I handle the error in the subscribe, the subscription will be closed and your token won't get refreshed anymore next times.
To give an example with the case you encountered:
1. You come back after a computer suspend, this subscription falls into the error case.
2. The error fn is executed, and it does an `invalidToken` (if appropriate)
3. After the given delay, no refresh will occur because the subscription is now closed because of the previous error. To me, the expected behavior is that the platform should keep on refreshing tokens after the given delay.

About the side effect part: fixed.
About not filtering errors: fixed.

ThomasJuster:
Post-it: suppress https://github.com/France-ioi/AlgoreaFrontend/pull/998 when merged

smadbe (changes requested):
Side effect in Rx fixed 👍 

What do we do here in case of failure. Options:

1. crash to sentry: what we try to avoid
2. not doing any processing anymore: so breaking the app... so worse than (1) I would say
3. ignore completely: we need to make sure a) it makes sense to ignore, b) it will not be called in loop immediately, c) if it will be re-called and whether this is acceptable. In our case: a) probably if the token is still valid, b) no, c) it will be re-called after 1min apparently
4. invalidate the token: good if the token is invalid (already expired or just invalid), not good for other cases (temporary server problem, network connection error, ...)

From that, I would say that a sane behavior for now would be:
1. if we get a 401 -> invalidate the token
2. if we get a non-401 error -> we ignore the error and wait for the next call in 1min

For case 2, should we notify the user or log something? As a non-(201|401) are really unexpected, maybe it would make sense to display a 5sec-red-toast "Unable to extend your session. Are you connected to Internet?" 

Does it make sense to you?

ThomasJuster:
> From that, I would say that a sane behavior for now would be:

> if we get a 401 -> invalidate the token
if we get a non-401 error -> we ignore the error and wait for the next call in 1min
For case 2, should we notify the user or log something? As a non-(201|401) are really unexpected, maybe it would make sense to display a 5sec-red-toast "Unable to extend your session. Are you connected to Internet?"

> Does it make sense to you?

Most of it makes sense to me :+1:. Only 2 points bother me:

##### 1. About presenting the end user the message "Are you connected to the internet?"
I'm not fond of it because we actually have that information through `navigator.onLine` and we can detect when connection is back with `fromEvent(window, 'online')`. But that is purely wording, we could show instead "Is your connection stable?", anyway the current message we show has tacitly the same meaning.

##### 2. About ignoring (unknown) errors and waiting for the next call in 1 minute

Any API call will give 401 response as long as we don't have a valid token (correct me if I'm wrong), so either we freeze the app, either we let the user navigate in a super buggy platform.
In that context, waiting 1 minute is very long in my opinion, we could retry after a few seconds. But that induces also handling to-many-fails scenario. To me, simpler actions like reloading the window would be more appropriate.

---

To sum up, that would change the sane behavior you described as follow:
There are 4 possible source of errors:

1. 401 -> invalidate token (no change)
3. 500 -> display feedback "Our server is unavailable, we are working on it. Sorry for the inconvenience" (or something like that)
4. Any other http error (404:not-found, 0:no-network, ..?) -> reload window
5. Other errors (only JS errors) -> let it bubble so that Sentry catches it. It will display a legitimate Sentry modal.

src/app/shared/auth/auth.service.ts
```diff
@@ -81,11 +82,21 @@ export class AuthService implements OnDestroy {
...
+      switchMap(auth => {
+        const tokenIsExpired = auth.expiration.valueOf() < Date.now();
+        if (tokenIsExpired) return of({ auth, tokenIsExpired }); // don't even try to refresh the token if it is expired
+        return this.authHttp.refreshAuth(auth).pipe(
+          catchError(err => {
+            // For any http error, ignore it since the token is valid. Another try will occur the next minute.
+            if (err instanceof HttpErrorResponse) return EMPTY;
+            else throw err; // rethrow any JS error to let our error monitor catch it.
+          }),
+          map(auth => ({ auth, tokenIsExpired })),
```
smadbe, line 94: For clarity, I would return `{ auth, tokenIsExpired: false }` instead (even `false` is the only possible value here) as there is not the expiration of the previous token that we return here (that may have become expired in the meantime) but the new one (which is non-expired hopefully).

smadbe (approved):

PR #1004, opened 2022-04-25
Feature/let user modify his profile

### LINKED ISSUE #991: Let the user modify his profile (on login-module)
## Overview

Let the user modify his profile by going to the login-module.

Check how it is done is done on https://parcours.algorea.org (login (create an account if needed), then go on your profile and "edit my profile") and do something similar (open a popup and refresh on "return").

## Acceptance criteria
In the personal data page:
- [ ] As shown on this design, above "Personal Information" section, display a text "The following information is what what receive from France-IOI login platform." with "Modify" button on the right
image
- [ ] when pressing the "modify" button, a popup is open for modifying the profile (as in https://parcours.algorea.org)
- [ ] when the popup is closed, the profile is refreshed by calling the [refresh profile](https://franceioi-algorea.s3.eu-west-3.amazonaws.com/spec/v1.1.0.html#operation/userDataRefresh) service and then refetching the data.

Iloveall:
## Description

Fixes #991

...

## Test cases

- [x] Case 1:
  1. Go to [this page](https://dev.algorea.org/branch/feature/let-user-modify-his-profile)
  3. And I click on Log in / Sign in button
  4. Login with username: update-profile-test, password: q1w2e3r4t5
  5. Then I go to profile page and click on Modify button
  6. Then I see new window with profile form 
  7. Then I change profile info and click on Save button
  8. Then I see modified info in app

### DIFF
--- a/src/app/core/app.component.ts
+++ b/src/app/core/app.component.ts
@@ -12,10 +12,7 @@ import { appConfig } from '../shared/helpers/config';
 import { urlToRedirectTo } from '../shared/helpers/redirect-to-sub-path-at-init';
 import { GroupWatchingService } from './services/group-watching.service';
 import { version } from 'src/version';
-import {
-  notifyWindowOpener,
-  subWindowUserDataUpdated
-} from '../modules/group/helpers/update-user-in-sub-window';
+import { isOpenerWindow, isWindowProfileUpdate, notifyWindowOpener } from '../modules/group/helpers/update-user-in-sub-window';
 
 @Component({
   selector: 'alg-root',
@@ -52,9 +49,12 @@ export class AppComponent implements OnInit, OnDestroy {
     private renderer: Renderer2,
     private el: ElementRef,
   ) {
-    subWindowUserDataUpdated(window.name, () =>
-      notifyWindowOpener(window.opener as Window)
-    );
+
+    // Special case for detect if user profile was updated via login-module in sub window and redirected after a close/save
+    if (isWindowProfileUpdate(window.name) && isOpenerWindow(window, window.opener)) {
+      // Notify opener/parent window so main app could refresh user data and close sub window
+      notifyWindowOpener(window.opener);
+    }
 
     const title = appConfig.languageSpecificTitles && this.localeService.currentLang ?
       appConfig.languageSpecificTitles[this.localeService.currentLang.tag] : undefined;
--- a/src/app/core/app.component.ts
+++ b/src/app/core/app.component.ts
@@ -50,7 +50,6 @@ export class AppComponent implements OnInit, OnDestroy {
   ) {
     if (window.name === 'updateProfileWindow') {
       (window.opener as Window).dispatchEvent(new Event('profileUpdated'));
-      window.close();
     }
 
     const title = appConfig.languageSpecificTitles && this.localeService.currentLang ?
--- a/src/app/core/app.component.ts
+++ b/src/app/core/app.component.ts
@@ -48,6 +48,11 @@ export class AppComponent implements OnInit, OnDestroy {
     private renderer: Renderer2,
     private el: ElementRef,
   ) {
+    if (window.name === 'updateProfileWindow') {
+      (window.opener as Window).dispatchEvent(new Event('profileUpdated'));
+      window.close();
+    }
+
     const title = appConfig.languageSpecificTitles && this.localeService.currentLang ?
       appConfig.languageSpecificTitles[this.localeService.currentLang.tag] : undefined;
     this.titleService.setTitle(title ?? appConfig.defaultTitle);
--- a/src/app/core/app.component.ts
+++ b/src/app/core/app.component.ts
@@ -12,6 +12,10 @@ import { appConfig } from '../shared/helpers/config';
 import { urlToRedirectTo } from '../shared/helpers/redirect-to-sub-path-at-init';
 import { GroupWatchingService } from './services/group-watching.service';
 import { version } from 'src/version';
+import {
+  notifyWindowOpener,
+  subWindowUserDataUpdated
+} from '../modules/group/helpers/update-user-in-sub-window';
 
 @Component({
   selector: 'alg-root',
@@ -48,9 +52,9 @@ export class AppComponent implements OnInit, OnDestroy {
     private renderer: Renderer2,
     private el: ElementRef,
   ) {
-    if (window.name === 'updateProfileWindow') {
-      (window.opener as Window).dispatchEvent(new Event('profileUpdated'));
-    }
+    subWindowUserDataUpdated(window.name, () =>
+      notifyWindowOpener(window.opener as Window)
+    );
 
     const title = appConfig.languageSpecificTitles && this.localeService.currentLang ?
       appConfig.languageSpecificTitles[this.localeService.currentLang.tag] : undefined;
--- a/src/app/core/app.component.ts
+++ b/src/app/core/app.component.ts
@@ -12,7 +12,6 @@ import { appConfig } from '../shared/helpers/config';
 import { urlToRedirectTo } from '../shared/helpers/redirect-to-sub-path-at-init';
 import { GroupWatchingService } from './services/group-watching.service';
 import { version } from 'src/version';
-import { isOpenerWindow, isWindowProfileUpdate, notifyWindowOpener } from '../modules/group/helpers/update-user-in-sub-window';
 
 @Component({
   selector: 'alg-root',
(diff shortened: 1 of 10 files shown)


src/app/modules/group/pages/current-user/current-user.component.html
```diff
@@ -1,5 +1,18 @@
...
+        The following information is what what receive from France-IOI login platform.
+      </div>
+      <button
+        pButton
+        type="button"
+        class="modify-button alg-button p-button-rounded p-button-info"
+        i18n-label label="Modify"
+        icon="pi pi-pencil"
+        (click)="onModify()"
+      ></button>
```
ThomasJuster, line 14: The button style and icon is unexpected:
image
versus
image

src/app/modules/group/pages/current-user/current-user.component.html
```diff
@@ -1,5 +1,18 @@
 <ng-container *ngIf="currentUser$ | async as currentUser">
   <div class="bg-white">
+    <div class="modify-section">
+      <div class="flex-1" i18n>
+        The following information is what what receive from France-IOI login platform.
```
ThomasJuster, line 5: Although that's exactly what's written in the issue, I doubt that "The following information is **what what** .." is correct :smile: 
I suppose: "The following information is what **we** receive from France-IOI login platform."

src/app/modules/group/pages/current-user/current-user.component.ts
```diff
@@ -36,4 +37,19 @@ export class CurrentUserComponent {
...
 
+  onModify(): void {
+    window.open(
+      `${ environment.oauthServerUrl }?redirect_uri=${ location.origin }`,
+      'updateProfileWindow',
+      'popup,width=800,height=640'
+    );
+
+    const onProfileUpdated = (): void => {
+      this.userSessionService.refresh().subscribe();
```
ThomasJuster, line 48: The refresh doesn't work for me when I close the popup window.
The behavior I have after saving in the popup is: the popup window displays our app in the popup.
What I expect when reading the issue: the popup closes itself and the user is refreshed on the tab that opened the popup.

src/app/modules/group/pages/current-user/current-user.component.html
Iloveall: Fixed

src/app/modules/group/pages/current-user/current-user.component.ts
ThomasJuster: chrome and firefox

Iloveall: I've changed close window approach, please try again

ThomasJuster: OK I got it, it's because the feature has to be tested on localhost...
When using the branch, the save button redirects to the prod version of the app, not the branch version

src/app/core/app.component.ts
```diff
@@ -49,6 +49,10 @@ export class AppComponent implements OnInit, OnDestroy {
     private renderer: Renderer2,
     private el: ElementRef,
   ) {
+    if (window.name === 'updateProfileWindow') {
+      (window.opener as Window).dispatchEvent(new Event('profileUpdated'));
+    }
+
```
ThomasJuster, line 55: Over time it will be difficult to understand why this is here.
We should find a way to make understand where this comes from, like moving it to a current-user helper in the group module for instance.

Iloveall: Fixed

ThomasJuster (commented):
:+1:

src/app/core/app.component.ts
```diff
@@ -49,6 +53,10 @@ export class AppComponent implements OnInit, OnDestroy {
     private renderer: Renderer2,
     private el: ElementRef,
   ) {
+    subWindowUserDataUpdated(window.name, () =>
+      notifyWindowOpener(window.opener as Window)
```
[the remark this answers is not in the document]
Iloveall: 1) If we can put simple html file on server and setting path (on server level) like: { domain }/notify-window-opener, then no need to load whole app of course.

2) we can inject code (with explain comment) into index.html file directly or via script, it's almost the same as 1 way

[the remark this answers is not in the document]
ThomasJuster: My case was different because it was about accessing nested properties.
Here we could more simply do `if (window.opener instanceof Window)`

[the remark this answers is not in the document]
ThomasJuster: I experimented this morning, due to cross-origin policies we have limited access to the subwindow.
If the oauth server posts a message to referer window (us) we'll be able to close the popup window. Otherwise it will require to load the app (via redirect uri) to close the popup at app startup
Or like Aleksandr said have a dedicated page of our own for that. But then I suspect that it's the redirect uri that won't work anymore since currently we can only login in the main branch and it's the same server...

Since it works with redirect uri, we could also not open a popup and instead use the redirect uri + our internal mechanism to redirect the app to a sub page (feature implemented in LTI context but not LTI-specific)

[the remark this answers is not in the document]
smadbe: >  If the oauth server posts a message to referer window (us) we'll be able to close the popup window. Otherwise it will require to load the app (via redirect uri) to close the popup at app startup

"oauth server posts a message to referer window " does not work due to cors limitations. The popup has to redirect to our code so that we can close it.

>  Or like Aleksandr said have a dedicated page of our own for that. But then I suspect that it's the redirect uri that won't work anymore since currently we can only login in the main branch and it's the same server...

If you mean the callback url check... there is no check on this one.

> Since it works with redirect uri, we could also not open a popup and instead use the redirect uri + our internal mechanism to redirect the app to a sub page (feature implemented in LTI context but not LTI-specific)

The request here was to do that in a popup. May be discussed but, at use, it looks more convenient to have a popup for such a usage.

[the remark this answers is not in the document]
smadbe: >   If we can put simple html file on server and setting path (on server level) like: { domain }/notify-window-opener, then no need to load whole app of course.

Yes, I thought this is what we discussed you should do. It will be also cleaner. But I don't know if that is easy in terms of webpack etc.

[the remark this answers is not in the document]
ThomasJuster: Can a html file be added to the assets (in angular.json > architect)? That would solve the problem.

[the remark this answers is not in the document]
Iloveall: I've tried this way before but window.opener is not instanceof Window.

But i've found how we could check it, please have a look to the code 
Screenshot at May 12 21-27-02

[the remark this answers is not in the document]
Iloveall: Fixed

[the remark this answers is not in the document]
Iloveall: But doesn't make sense as it will be moved to separated html file without ts supporting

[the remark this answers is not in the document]
Iloveall: Fixed

src/app/core/app.component.ts
```diff
@@ -12,6 +12,10 @@ import { appConfig } from '../shared/helpers/config';
 import { urlToRedirectTo } from '../shared/helpers/redirect-to-sub-path-at-init';
 import { GroupWatchingService } from './services/group-watching.service';
 import { version } from 'src/version';
+import {
+  notifyWindowOpener,
+  subWindowUserDataUpdated
+} from '../modules/group/helpers/update-user-in-sub-window';
```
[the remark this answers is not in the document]
Iloveall: These files have been removed as all logic in simple html file

src/app/modules/group/pages/current-user/current-user.component.ts
```diff
@@ -36,4 +37,19 @@ export class CurrentUserComponent {
     });
   }
 
+  onModify(userId: string): void {
+    window.open(
+      `${ environment.oauthServerUrl }?all=1&client_id=${ userId }&redirect_uri=${encodeURI(`${ location.origin }/update-profile.html`)}`,
```
ThomasJuster, line 42: I think this works in localhost but won't work in production. Encoded redirect uri will be:
- on localhost: `'http://localhost:4200/update-profile.html'`
- on production algorea: `'https://dev.algorea.org/update-profile.html'` instead of `'https://dev.algorea.org/{en | fr}/update-profile.html'`

I tested, on your branch the html file is located at
`'https://dev.algorea.org/branch/feature/let-user-modify-his-profile/{en | fr}/update-profile.html'`
so in production the file will be located at
`https://dev.algorea.org/{en | fr}/update-profile.html`

If we want to make it context agnostic, we can use a relative path:
```js
const url = new URL(
  './update-profile.html' /* relative url */,
  location.href /* base of relative url */
).href;
```
That way it will work for branch build, production builds, localhost and fr/en builds.

Iloveall: Fixed

ThomasJuster (approved):

src/update-profile.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <title>Update profile</title>
```
smadbe, line 4:
```suggestion
  <title>Updating profile...</title>
```

src/update-profile.html
```diff
@@ -0,0 +1,16 @@
...
+</head>
+<body>
+<script>
+  // Special case for detect if user profile was updated via login-module in sub window and redirected after a close/save
```
smadbe, line 8:
```suggestion
  // This page is supposed to be called (only!) as a callback from login-module after a profile update (on save or close) in a subwindow (popup). 
```

src/update-profile.html
```diff
@@ -0,0 +1,16 @@
...
+  if (window.opener) {
+    // Notify opener/parent window so main app could refresh user data and close sub window
```
smadbe, line 10:
```suggestion
    // Notify the opener/parent window so that the main app can refresh the user data and close the sub-window
```

smadbe (approved):

PR #1006, opened 2022-04-26
Simpler task loading message

### LINKED ISSUE #948: Simpler task loading messages: add text when loading is long
## Overview

Should be done after #947.

The two loading messages "Starting the activity" and "Loading the task" may be long and if so, more info should be displayed to the users.

## Notes

* I have assigned this task to @Iloveall but don't hesitate to ask  @ThomasJuster if you need help on Rx.
* Don't forget to make me (@smadbe) review the UI.

## Acceptance criteria
- [ ] For both case, if the loading is still ongoing after 5sec, display another sentence under it (smaller font, in grey (?) ).
- [ ] For the first, indicate "This should not take long. If you are stuck on this message, please retry and let us know if the problem persists."
- [ ] For the second, "The task is taking an unexpected long time to load... please wait..."

ThomasJuster:
## Description

Fixes #948 

:warning: The first use case cannot be tested because "Starting the activity" is displayed while waiting for an xhr request, which has a 3s timeout. Therefore, it will fall into an error state before actually displaying a message under "Starting the activity"

Here's how the message looks like:
image

## Test cases

- [ ] Case 1:
  1. Given I am any user with a throttled network
  2. When I go to [this item](https://dev.algorea.org/branch/simpler-task-loading-message/en/#/activities/by-id/2377619260689967292;path=4702,458602124916982205,4769434107088212490,8123435593727593341,5072731755477071719,8600972556668887052;parentAttempId=0/details)
  3. And I wait 5s after I see "Task loading"
  4. Then I see an additional message "The task is taking too long..."

### DIFF
--- a/src/app/modules/item/components/task-loader/task-loader.component.ts
+++ b/src/app/modules/item/components/task-loader/task-loader.component.ts
@@ -1,14 +1,31 @@
-import { Component, Input } from '@angular/core';
+import { Component, Input, OnDestroy, OnInit } from '@angular/core';
+import { Subscription, timer } from 'rxjs';
+import { SECONDS } from 'src/app/shared/helpers/duration';
 
 @Component({
   selector: 'alg-task-loader',
   templateUrl: './task-loader.component.html',
   styleUrls: [ './task-loader.component.scss' ]
 })
-export class TaskLoaderComponent {
+export class TaskLoaderComponent implements OnInit, OnDestroy {
 
   @Input() label = '';
+  @Input() delayedLabel = '';
+  @Input() delay = 5;
+
+  showDelayedLabel = false;
+  subscription?: Subscription;
 
   constructor() { }
+
+  ngOnInit(): void {
+    this.subscription = timer(this.delay*SECONDS).subscribe(() => {
+      this.showDelayedLabel = true;
+    });
+  }
+
+  ngOnDestroy(): void {
+    this.subscription?.unsubscribe();
+  }
 
 }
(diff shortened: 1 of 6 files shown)


src/app/modules/item/components/task-loader/task-loader.component.ts
```diff
@@ -1,14 +1,31 @@
...
   selector: 'alg-task-loader',
   templateUrl: './task-loader.component.html',
   styleUrls: [ './task-loader.component.scss' ]
 })
-export class TaskLoaderComponent {
+export class TaskLoaderComponent implements OnInit, OnDestroy {
 
   @Input() label = '';
+  @Input() delayedLabel = '';
+  @Input() delay = 5;
```
smadbe, line 14: add some comment to explain what is a "delay" on a task loader.

src/app/modules/item/components/task-loader/task-loader.component.ts
```diff
@@ -1,14 +1,31 @@
...
+
+  showDelayedLabel = false;
+  subscription?: Subscription;
 
   constructor() { }
 
+  ngOnInit(): void {
+    this.subscription = timer(this.delay*SECONDS).subscribe(() => {
```
smadbe, line 22: maybe only if `this.delayedLabel` is set  (if so, this should be optional instead of defaulting to empty string)

smadbe (approved):

src/app/modules/item/components/task-loader/task-loader.component.ts
ThomasJuster: Fixed

src/app/modules/item/components/task-loader/task-loader.component.ts
ThomasJuster: Fixed

PR #1010, opened 2022-04-27
Feature/group editing move editing into the settings tab

### LINKED ISSUE #982: Group editing: move editing into the settings tab
## Overview
Get rid of the editing mode for groups. 

Instead of letting the user editing group `name`, `description` and `associated activity` in editing mode, move this to the "settings" tab.

## Acceptance criteria
- [ ] allow editing the `name`, `description` and `associated activity` in the settings tabs
- [ ] do not display the "edit" mode button in groups
- [ ] As in the current edit mode, display the "save"/"cancel changes" button in the bottom right when there are unsaved changes
- [ ] As in the current edit mode, show a modal when leaving the tab when there are unsaved changes
- [ ] Move the "Only an empty group can be deleted" indication and "delete this group" button on the left so that it does not overlap with the save/cancel buttons
- [ ] remove the "edit" route for groups
- [ ] remove the "edit" component for groups

## Non-goal

* No need to reorganize the component hierarchy for groups for now

Iloveall:
## Description

Fixes #982

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/group-editing-move-editing-into-the-settings-tab/en/#/groups/by-id/5121055722873780306;path=/details/settings)
  3. Then I see Group information and Associated Activity sections without edit button in header

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/group-editing-move-editing-into-the-settings-tab/en/#/groups/by-id/5121055722873780306;path=/details/settings)
  3. Change description under Group information section
  4. Then I see Save and Cancel Changes buttons
  5. Click on save
  6. Then I see updated information

- [ ] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/group-editing-move-editing-into-the-settings-tab/en/#/groups/by-id/5121055722873780306;path=/details/settings)
  3. Change description under Group information section
  4. Then I see Save and Cancel Changes buttons
  5. Click on cancel
  6. Then I see old information
  
- [ ] Case 4:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/group-editing-move-editing-into-the-settings-tab/en/#/groups/by-id/5121055722873780306;path=/details/settings)
  3. Change description under Group information section
  4. Then I see Save and Cancel Changes buttons
  5. Go to Access tab
  6. Then I see confirmation modal


src/app/modules/group/components/group-header/group-header.component.html
```diff
@@ -8,7 +8,6 @@
     <ng-container *ngIf="navigationNeighbors$ | async as navigationNeighbors">
```
ThomasJuster, line 8: As usual user
- when I go to [this group](https://dev.algorea.org/branch/feature/group-editing-move-editing-into-the-settings-tab/en/#/groups/by-id/5121055722873780306;path=/details/settings)
- and I edit without saving
- and I click on a sibling group in the left nav (ie 4LTDI..)

the navigation is not blocked

Iloveall: Okay, this implementation is problematic, it doesn't call can deactivate of settings when group id (which parent component) changed.

And here is one more case:

As usual user:

1.  When I go to this (group)[https://dev.algorea.org/branch/feature/group-editing-move-editing-into-the-settings-tab/en/#/groups/by-id/9124554819883263635;path=/details/settings]
2. And I click on Managers tab
3. Then I can see error message in console of browser or sentry modal

The problem is, when we are not allow to see group edit component - it is not init, so no isDirty state and error as result. 

As solution - Group Settings must be also PendingChangesComponent.

Here is PR with offer how it works with router outlet, obviously not a lot changes https://github.com/France-ioi/AlgoreaFrontend/pull/1035

I didn't spend a lot of time, just offer, so probably need to polish

What do you think? @ThomasJuster @smadbe

ThomasJuster: It looks fine to me but on such matter it will be more up to @smadbe

smadbe:
The settings with "..." section was just supposed to be there while the tab had no content. This has to be removed now.

Screenshot 2022-05-20 at 10 08 11

src/app/modules/group/components/group-header/group-header.component.html
smadbe: > As solution - Group Settings must be also PendingChangesComponent." -> GroupSettings. 
As I [commented](https://github.com/France-ioi/AlgoreaFrontend/pull/1010#issuecomment-1132614821), as the remaining section has to be removed, `GroupSettings` would be somehow empty and should be removed anyway.

It will solved the issue your mention, but I am not sure it will solve Thomas'. Does it?

Iloveall: It will, you can test it in this [PR](https://github.com/France-ioi/AlgoreaFrontend/pull/1035) .

The general idea to use standard angular way, so it must cover all usual cases in web development

smadbe: My remark was only on the "As solution - Group Settings must be also PendingChangesComponent." -> GroupSettings.", not on the next proposal of changing the tab navigation using the outlet.

If you do so, you become inconsistent with the rest of the app, and make use of something global for something which is not global from my point of view, which I find "wrong" in terms of software engineering.

So ok, I will check how we can fix cleanly the problem you mention when I'll have a few minutes available.

Iloveall: I see, yes remove GroupSettings will solve problem with PendingChangesComponent component, but not Thomas issue. 

So let me fix this issue, for another I will wait your opinion

Iloveall: After fix this I’ve found small issue:

As usual user
1. Go to https://dev.algorea.org/branch/feature/group-editing-move-editing-into-the-settings-tab/en/#/groups/by-id/5485463250412766338;path=/details/settings
2. Click on Access tab
3. Settings tab disappeared 

So GroupSettings and GroupEdit `authorized` states are not equal.

smadbe: Maybe you have dropped (in this PR) the perm check while merging component... in this case, the user is not manager, so he definitely not have the permission to view the "settings" page (even if it is harmless as he will not be able to save).

smadbe: And so, there is also Thomas' scenario which is still not working as expected. But maybe we can do this in a separate PR if it makes things cleaner ? (the bug is not very problematic)

Iloveall: > 

Fixed

Iloveall: > Maybe you have dropped (in this PR) the perm check while merging components... in this case, the user is not manager, so he definitely not have the permission to view the "settings" page (even if it is harmless as he will not be able to save).

Fixed

Iloveall: > And so, there is also Thomas' scenario which is still not working as expected. But maybe we can do this in a separate PR if it makes things cleaner ? (the bug is not very problematic)

Definitely

smadbe: Created issue #1055 for that.

smadbe (approved):
Ok. Just invert the text and button down there:
Screenshot 2022-06-02 at 11 33 17
 
The button just always be left aligned... the text is just an indication and so should displayed on the right of of the button.

Iloveall:
> Ok. Just invert the text and button down there: Screenshot 2022-06-02 at 11 33 17
> 
> The button just always be left aligned... the text is just an indication and so should displayed on the right of of the button.

Screenshot at Jun 02 13-26-44
Screenshot at Jun 02 13-20-25

PR #1017, opened 2022-05-03
use base href completely instead of deploy url

### LINKED ISSUE #905: Fix static common asset serving: tree-folder image not included as it should
**Describe the bug**

The `tree-folder.svg` image, used in the left menu seems not to be included as it should.
As a consequence, it is not served and packed as the other assets and it causes me other deployment issues.

**Screenshots**
Screenshot 2022-02-09 at 18 22 11

### LINKED ISSUE #909: Fingerprint files which are not yet with other assets (if possible)
## Overview
Some files are still not served with assets and not fingerprinted: those listed in index.html: so favicon.ico (this one probably cannot be fingerprinted/renamed/moved), favicon*png, apple*png, site.webmanifest. Those in webmanifest as well I assume. 

If possible, try to make them considered as assets by "Angular" so that they can be served as other assets (in a CDN).

ThomasJuster:
## Description

Fixes #909 
Fixes #905 

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/handle-static-assets/en/#/)
  3. And I click on ...
  4. Then I see ...
  5. And

- [ ] Case 2:
  1. Given I am the usual user
  2. When ...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ commands:
         type: string
     steps:
       - run: echo "export const version = '`git describe --always`';" > ./src/version.ts
-      - run: npx ng build --configuration production-<< parameters.lang >> --base-href //assets.algorea.org<< parameters.path >>/ --progress=false
+      - run: npx ng build --configuration production-<< parameters.lang >> --base-href //assets.algorea.org<< parameters.path >> --progress=false
 
 jobs:
   # deps has to be run before the other jobs and basically run "npm install" and cache it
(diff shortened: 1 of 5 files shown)


ThomasJuster:
Note to self (since I interrupted development):

Cannot use external base href because it breaks the navigation (history.pushState becomes an "invalid DOM operation")
Last idea I had before interruption: use different index.html per envs (prod, dev, branch)

ThomasJuster:
Close in favor #1079

PR #1021, opened 2022-05-03
fix type error on history state preventRefetch

### LINKED ISSUE #1013: Cannot read property `preventFetch` of `null` (`history.state`)
**Describe the bug**
See [Sentry report](https://sentry.io/organizations/france-ioi/issues/3227466121/?project=6257761)

ThomasJuster:
## Description

Fixes #1013 

...

## Test cases

I could not reproduce the bug, but since `history.state` can be null, adding optional chaining should suffice.


smadbe (commented):
1. I see 2 places where `history.state` is used for `preventFullFrame`. I suppose both should be fixed
2. From what I see `history.state` is `any`. So casting it to `Record<string, boolean>` looks wrong to me anyway as it is just assumption.  Could we have something more explicit / safe such as `history.state !== null && history.state.get('preventRefetch')` (? not sure it is valid) or  `decode(D.struct({
    preventRefetch: D.number }))` (require handling the error)

smadbe (approved):

PR #1026, opened 2022-05-05
Handle request timeouts

### LINKED ISSUE #1012: Some timeouts trigger Sentry modal
**Describe the bug**
Errors should be handled properly by us and not Sentry.
[Sentry report](https://sentry.io/organizations/france-ioi/issues/3228190417/?project=6257761)

Investigate where it comes from, starting by the LTI module.

**To Reproduce**
:construction: 

**Expected behavior**
No Sentry modal should be displayed

ThomasJuster:
## Description

Fixes #1012 

This initial problem should have been fixed by #1018, anyway some enhancement might benefit us:
- Because timeout error stacks are a hell to read, and since we already apply a client-side timeout, I suggest we also intercept timeout errors and replace them by http error 408 timeout. Such errors will be much more readable in the future than rxjs timeouts.
- Since the check login call looks like a source of trouble, I added a retry strategy (3 times) to avoid breaking the whole page when network is temporarily slower.

## Test cases

:warning: can be tested locally only!

- [ ] Case 1:
  1. Given I am any user
  2. When I go to [moodle test "Test LTI 2"](http://moodletest.algorea.org/course/view.php?id=3)
  3. And I click on "Test dev.algorea.org (new platform localhost)"
  4. Then I am redirected to algorea platform
  5. And I open the network panel and refresh]
  6. Then I see 2 fails on check-login service, 3rd is a success, and the platform is loaded normally.

### DIFF
--- a/src/app/shared/http-services/check-login.service.ts
+++ b/src/app/shared/http-services/check-login.service.ts
@@ -21,7 +21,7 @@ export class CheckLoginService {
     const params = new HttpParams({ fromObject: { login_id: loginId } });
 
     return this.http
-      .get<unknown>(`${appConfig.apiUrl}/current-user/check-login-id`, { params, headers: ++count <= 2 ? { timeout: '50' } : {} })
+      .get<unknown>(`${appConfig.apiUrl}/current-user/check-login-id`, { params })
       .pipe(
         decodeSnakeCase(dataDecoder),
         map(data => data.loginIdMatched),
@@ -29,5 +29,3 @@ export class CheckLoginService {
   }
 
 }
-
-let count = 0;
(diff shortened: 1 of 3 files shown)


src/app/shared/http-services/check-login.service.ts
```diff
@@ -21,11 +21,13 @@ export class CheckLoginService {
...
+      .get<unknown>(`${appConfig.apiUrl}/current-user/check-login-id`, { params, headers: ++count <= 2 ? { timeout: '50' } : {} })
       .pipe(
         decodeSnakeCase(dataDecoder),
         map(data => data.loginIdMatched),
       );
   }
 
 }
+
+let count = 0;
```
smadbe, line 33: Kind of global var? The service are supposed to be created only once if well used. If so, it could be a instance variable instead.

ThomasJuster: This is test content to revert

smadbe: ah sorry

smadbe (approved):

PR #1029, opened 2022-05-10
fix permission dialog when watching group

ThomasJuster:
## Description

Fixes #...

From my understanding, in this context (watching a group) `giverPermissions=item.permissions`, can you confirm that @smadbe and @GabrielVidal1 ?

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this group](https://dev.algorea.org/branch/fix-watchedgroup-permissions-dialog/en/#/groups/by-id/672913018859223173;path=52767158366271444/details)
  3. And I watch group
  4. And I click on first suggestion "Test 2"
  5. And I open collapsible section "Pixal can access this activity"
  6. And I click on "Edit permissions"
  7. Then I see the permissions dialog is displayed
  8. And I edit the first value of the form and save
  9. Then I see permissions are saved (network panel) and the item is reloaded seamlessly

### DIFF
--- a/src/app/modules/item/components/item-permissions/item-permissions.component.html
+++ b/src/app/modules/item/components/item-permissions/item-permissions.component.html
@@ -81,7 +81,7 @@
           Note that some subgroups or users of the groups may have higher permissions.
         </div>
 
-        <ng-container *ngIf="permissions.canGrantView !== 'none' && watchedGroup.currentUserCanGrantGroupAccess">
+        <ng-container *ngIf="permissions.canGrantView !== 'none' && watchedGroup.currentUserCanGrantGroupAccess && itemData?.item?.permissions">
           <div class="give-access-button">
             <button
                 pButton
@@ -98,6 +98,7 @@
               [title]="watchedGroup.name"
               [targetType]="'Groups'"
               [permissions]="permissions"
+              [giverPermissions]="itemData?.item?.permissions"
               (close)="closePermissionsDialog()"
               (save)="onPermissionsDialogSave($event)"
           ></alg-permissions-edit-dialog>
--- a/src/app/modules/item/components/item-permissions/item-permissions.component.html
+++ b/src/app/modules/item/components/item-permissions/item-permissions.component.html
@@ -99,7 +99,7 @@
               [title]="watchedGroup.name"
               [targetType]="'Groups'"
               [permissions]="permissions"
-              [giverPermissions]="itemData?.item?.permissions"
+              [giverPermissions]="itemData.item.permissions"
               (close)="closePermissionsDialog()"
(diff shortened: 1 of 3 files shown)


smadbe:
> From my understanding, in this context (watching a group) giverPermissions=item.permissions, can you confirm that ?

In general, yes but not always. Not when watching a user for instance. Hum I have to think how we can solve that...

smadbe:
> > From my understanding, in this context (watching a group) giverPermissions=item.permissions, can you confirm that ?
> 
> In general, yes but not always. Not when watching a user for instance. Hum I have to think how we can solve that...

In practice, for quite wrong/bad reason, there is never this "Edit permissions" on user. And for other case, it should also work "by coincidence".
So you can take this assumption for now... maybe just put into comment that it is assumption which is valid for now.

src/app/modules/item/components/item-permissions/item-permissions.component.html
```diff
@@ -93,11 +93,13 @@
             ></button>
           </div>
 
+          <!-- NOTE: For now, we can assume that giver permissions is equivalent to user permissions on item. -->
```
smadbe, line 96: I am very sorry, I've completely misunderstood your initial question. 

> From my understanding, in this context (watching a group) giverPermissions=item.permissions,

Yes and that's all. 

All my remarks were about the permission source group... but that's completely unrelated and another issue.

So this comment you added does not make sense. Can you remove it please? 😔

src/app/modules/item/components/item-permissions/item-permissions.component.html
```diff
@@ -81,7 +81,7 @@
           Note that some subgroups or users of the groups may have higher permissions.
         </div>
 
-        <ng-container *ngIf="permissions.canGrantView !== 'none' && watchedGroup.currentUserCanGrantGroupAccess">
+        <ng-container *ngIf="permissions.canGrantView !== 'none' && watchedGroup.currentUserCanGrantGroupAccess && itemData?.item?.permissions">
```
smadbe, line 84: `item` does not seem to be optional in `itemData`

smadbe (approved):

src/app/modules/item/components/item-permissions/item-permissions.component.html
ThomasJuster: Still the template optional chaining issue, TS detects it properly but angular templates don't :shrug:

src/app/modules/item/components/item-permissions/item-permissions.component.html
ThomasJuster: No pb, thanks for taking the time to check :pray: 
Fixed

Iloveall (approved):

PR #1030, opened 2022-05-11
path suggestion proposal

ThomasJuster:
## Description

Fixes #...

I created this PR because in my opinion using intensively rxjs was not _that_ readable and might also be tricky to maintain later on. When taking a step back, the goal was to fix 2 remaining code issues:
- A mix between rxjs to open the overlay with delay and imperative js to close the overlay
- The `while` + `if` loop not very maintainable in my opinion in the future

For UI issues, I did not touch the UI so I'll comment them in the original PR.

For the last row UI issue, I think it's an acceptable trade-off to leave it as is because I think the extra effort to make it work it would require is less interesting than the value it brings to the platform.

## Test cases (taken from original PR)

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/path-suggestion-proposal/en/#/groups/by-id/4322568739774621128;path=52767158366271444/details)
  3. And hover on one of row table
  4. Then I see overlay with path suggestion

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/path-suggestion-proposal/en/#/groups/by-id/49268055312628123;path=/details)
  3. And hover on one row with "Locked content"
  4. Then I see overlay with path suggestion with error message: "You are not allowed to view path suggestion."

### DIFF
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.ts
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.ts
@@ -1,19 +1,8 @@
-import {
-  Component,
-  ElementRef,
-  Input,
-  OnChanges,
-  OnDestroy,
-  OnInit,
-  QueryList,
-  ViewChild,
-  ViewChildren
-} from '@angular/core';
-import { BehaviorSubject, debounceTime, Observable, ReplaySubject, Subject, Subscription } from 'rxjs';
-import { filter, map, shareReplay, switchMap } from 'rxjs/operators';
+import { Component, ElementRef, Input, OnChanges, OnDestroy, QueryList, ViewChild, ViewChildren } from '@angular/core';
+import { BehaviorSubject, debounceTime, merge, Observable, ReplaySubject, Subject } from 'rxjs';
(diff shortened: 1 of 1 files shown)


src/app/modules/group/components/group-log-view/group-log-view.component.ts
```diff
@@ -130,27 +118,17 @@ export class GroupLogViewComponent implements OnChanges, OnInit, OnDestroy {
...
-    if (target instanceof HTMLElement && relatedTarget instanceof HTMLElement) {
-      let currentElement = relatedTarget;
-      while (currentElement.parentElement) {
-        if (currentElement.classList.contains('alg-path-suggestion-overlay')) {
-          closeOverlay = false;
-          break;
-        }
-        currentElement = currentElement.parentElement;
-      }
-    }
```
ThomasJuster, line 144: I tested it and it looked to me to be exactly equivalent to [`.closest(selector)` function](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest).

Iloveall (approved):

PR #1032, opened 2022-05-11
Fix jschannel error

### LINKED ISSUE #1031: jschannel complete call to already completed transaction
**Describe the bug**
See [sentry report](https://sentry.io/organizations/france-ioi/issues/3235068884/?project=6257761&query=is%3Aunresolved)

**To Reproduce**
I (@ThomasJuster) could not reproduce in production. After checking the [source code](https://github.com/mozilla/jschannel/blob/master/src/jschannel.js), it might happen when the jschannel is destroyed before an bound method observable emits a value. Sentry reporting a navigation event before the error advocates for this hypothesis.

As done before, I'll create a sandbox reproducing the error in a commit and a fix in another commit.

**Expected behavior**
No error.

ThomasJuster:
## Description

Fixes #1031 

...

## Test cases

- [ ] Case 1 - broken sample:
  1. Given I am the any user
  2. When I go to [this activity](https://dev.algorea.org/branch/reproduce-jschannel-error/en/#/activities/by-id/6379723280369399253;path=4702,1625159049301502151,5207993233955027100;parentAttempId=0/details)
  3. And I throttle my network to a bad enough connection, slow 3G for instance.
  4. And I complete puzzle version 1
  5. Then I see sentry modal.

- [ ] Case 1 - fixed sample:
  1. Given I am the any user
  2. When I go to [this activity](https://dev.algorea.org/branch/fix-jschannel-error/en/#/activities/by-id/6379723280369399253;path=4702,1625159049301502151,5207993233955027100;parentAttempId=0/details)
  3. And I throttle my network to a bad enough connection, slow 3G for instance.
  4. And I complete puzzle version 1
  5. Then I see no sentry modal.

### DIFF
--- a/src/app/modules/item/task-communication/rxjschannel.ts
+++ b/src/app/modules/item/task-communication/rxjschannel.ts
@@ -23,6 +23,8 @@ export function rxBuild(config: Omit<ChannelConfiguration, 'onReady'>): Observab
 }
 
 export class RxMessagingChannel {
+  private destroyed = false;
+
   constructor(private channel: MessagingChannel) {
     Object.assign(window, {
       destroyChan: () => {
@@ -38,8 +40,8 @@ export class RxMessagingChannel {
   /** Bind a local method, allowing the remote task to call it */
   bind<T extends Observable<any> | any>(method: string, mapper?: (params: unknown) => T, doNotPublish?: boolean): MessagingChannel {
     // Create a callback wrapping the observable bound
-    function callback(transaction: MessageTransaction, params: unknown): void {
-      if (!mapper) return;
+    const callback = (transaction: MessageTransaction, params: unknown): void => {
+      if (!mapper || this.destroyed) return;
 
       const forwardError = (error: unknown): void => transaction.error(error, error instanceof Error ? error.toString() : '');
       try {
@@ -50,13 +52,15 @@ export class RxMessagingChannel {
         result
           .pipe(take(1))
           .subscribe({
-            next: data => transaction.complete(data),
+            next: data => {
+              if (!this.destroyed) transaction.complete(data);
+            },
             error: forwardError,
           });
       } catch (error) {
         forwardError(error);
       }
-    }
+    };
     return this.channel.bind(method, callback, doNotPublish);
   }
 
@@ -81,7 +85,8 @@ export class RxMessagingChannel {
     this.channel.notify(message);
   }
 
-  destroy(): void{
+  destroy(): void {
+    this.destroyed = true;
     this.channel.destroy();
   }
 }
--- a/src/app/modules/item/task-communication/rxjschannel.ts
+++ b/src/app/modules/item/task-communication/rxjschannel.ts
@@ -25,13 +25,7 @@ export function rxBuild(config: Omit<ChannelConfiguration, 'onReady'>): Observab
 export class RxMessagingChannel {
   private destroyed = false;
 
-  constructor(private channel: MessagingChannel) {
-    Object.assign(window, {
-      destroyChan: () => {
-        this.destroy();
-      }
-    });
-  }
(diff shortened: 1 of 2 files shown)


src/app/modules/item/task-communication/rxjschannel.ts
```diff
@@ -50,13 +52,15 @@ export class RxMessagingChannel {
         result
           .pipe(take(1))
           .subscribe({
-            next: data => transaction.complete(data),
+            next: data => {
+              if (!this.destroyed) transaction.complete(data);
```
smadbe, line 56: We could even save the subscription and unsubscribe on "destroy" so that there is not a piece of code still waiting on something that does not make sense anymore. 
At first, it looks a bit like  `if (!this.destroyed)` is enough for safety.... but just reading the class (so at least in the scope of this class), I don't see the guaranty that `result` always emits... with the risk of retaining the submission (and so `this` as well).  -> I am not 100% of what I am writing here ;-)
(and so, if we do the unsubscription, probably `if (!this.destroyed)` is not useful anymore)

src/app/modules/item/task-communication/rxjschannel.ts
```diff
@@ -81,7 +85,8 @@ export class RxMessagingChannel {
     this.channel.notify(message);
```
smadbe, line 85: Shouldn't all call to `channel` (here, on line 80, 64, 37) make sure it is not destroyed first? (maybe with exceptions here)
From what I see it should not really happen if the class is well used, but so that's relying on the caller, so an extra checks looks useful.

smadbe (approved):
nice 👍 👍

src/app/modules/item/task-communication/rxjschannel.ts
ThomasJuster: This method is unused, so I'd rather do nothing on it :sweat_smile:

src/app/modules/item/task-communication/rxjschannel.ts
ThomasJuster: For instance here we are saving a grade (xhr) and then notifying the task that the operation went OK (the example works with other xhr requests).
Unsubscribing would mean cancelling the saving grade xhr request when the task is destroyed and I don't think that's what we want, is it?
To me what we want is: 1) if task is active, notify it that the operation went OK, and 2) if the task is destroyed, don't notify it that it went ok.
And personally I don't see which observable might never end in this context since here we use observables that emit a single value (`take(1)`). The only risk is an observable that would never completes or fall in error state, and even that I don't know where it would be declared :thinking: 

By writing that, I realise that I have handle only success use case and not failure, I'll add a fix.

src/app/modules/item/task-communication/rxjschannel.ts
ThomasJuster: (But ok for other calls)

src/app/modules/item/task-communication/rxjschannel.ts
smadbe: > To me what we want is: 1) if task is active, notify it that the operation went OK, and 2) if the task is destroyed, don't notify it that it went ok.

Makes sense! Agreed.

>  And personally I don't see which observable might never end in this context since here we use observables that emit a single value (take(1)). The only risk is an observable that would never completes or fall in error state, and even that I don't know where it would be declared 🤔

So because you do assumptions on the way caller use this function... of course if it xhr only.
Maybe you should add this in the `bind` spec.  We should also add a (large) timeout, but that may be overkill.

src/app/modules/item/task-communication/rxjschannel.ts
ThomasJuster: fixed

PR #1035, opened 2022-05-17
Feature/group edit can deactivate offer

### LINKED ISSUE #982: Group editing: move editing into the settings tab

Iloveall:
## Description

Fixes #982

Here is offer for solution to fix can deactivate issue with little refactoring. 

The Group Settings component doesn't use groupRefreshRequired output, but it defined. Also Group Edit component already has all needed data without Inputs, so it uses the same group source. Obviously only one thing we need to change in Group Settings is authorized state and it's the same group source.

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/group-edit-can-deactivate-offer/en/#/groups/by-id/5121055722873780306;path=/details/settings)
  3. Change description under Group information section
  4. Then I see Save and Cancel Changes buttons
  5. Go to Access tab
  6. Then I see confirmation modal

- [x] Case 2:
  1. Given I am the usual user
  2. When I go to this [group](https://dev.algorea.org/branch/feature/group-edit-can-deactivate-offer/en/#/groups/by-id/9124554819883263635;path=/details/settings)
  3. And I click on Managers tab
  4. Then I see managers content


smadbe:
We can keep this as a reference but cannot use it as it is as it introduces other problems.

PR #1046, opened 2022-05-27
Feature/split personal data from platform settings

### LINKED ISSUE #997: Profile: split personal data from platform settings
Require #991 to be merged.

## Overview

The current "Personal data" tab include both platform settings (such as the language) and personal data fetched from the login-module. Now the "Modify" button has been added, it makes more sense to extract the non personal data out from the "personal data" tab.

## Acceptance criteria
- [x] Create a new tab (and route) "Settings" (last tab) 
- [x] Use the same restrictions to view the "Settings" tab as the "Personal data" tab
- [x] Move the language dropdown on this tab
- [x] Use "Platform settings" as  section name instead of "Parameters"
- [x] Make sure the language switch still work as expected
- [x] Make sure the personal data "modify" is not broken
- [x] Make sure the arrival on tabs works as expected, including when the current user does not have the permission to view the personal data or settings

Iloveall:
## Description

Fixes #997

I've improved "You cannot access this page for this user" message to bring it to standard style

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/split-personal-data-from-platform-settings/en/#/groups/users/670968966872011405/settings)
  3. The I see Platform settings section
  4. Change language to FR
  5. Then I see UI text in FR

- [x] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/split-personal-data-from-platform-settings/en/#/groups/users/670968966872011405/personal-data)
  3. The I see only personal data sections
  4. Click on Modify button and update info of user
  5. Then I see updated personal data

- [x] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/split-personal-data-from-platform-settings/en/#/groups/users/752024252804317630/personal-data)
  3. The I see "You cannot access this page for this user"

- [x] Case 4:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/split-personal-data-from-platform-settings/en/#/groups/users/752024252804317630/settings)
  3. The I see "You cannot access this page for this user"


src/app/modules/group/group-routing.module.ts
```diff
@@ -18,6 +18,9 @@ const routes: Routes = [
     children: [
```
ThomasJuster, line 18: I found an unrelated bug:
When using the update-profile-test user and updating its data via the modify button, the section above the tabs is not refreshed. (I edited the website and the first name.)
I'll create an issue for that.
image

ThomasJuster, line 18: I found a weird behavior:
When visiting [this page](https://dev.algorea.org/branch/feature/split-personal-data-from-platform-settings/en/#/groups/users/752024252804317630/settings)
Then clicking on "Personal data" works fine
Then clicking on "Progress" I have a weird behavior, all tabs disappear.

As I understand, what we did until today was displaying the allowed tabs and only the focused unallowed tab if there's one.
Here when landing on "Settings" we also display the "Personal data" tab but the access to it is forbidden. I think we should not display the "Personal data" tab
Same, when landing on "Personal data" we also display the "Settings" tab but the access to it is forbidden. I think we should not display the "Settings" tab.

Iloveall: Fixed

ThomasJuster (commented):
:+1:

smadbe (approved):
looks good

PR #1054, opened 2022-05-31
Bugfix/error report modal on profile refresh error

### LINKED ISSUE #1050: Do not show error report modal on profile refresh error
**Describe the bug**
When the profile service refresh fail (500) after a profile modification, the error report modal is shown.

**To Reproduce**
1. go on https://dev.algorea.org/ as the usual user,
2. go on profile, 
3. click "modify" on profile... the popup opens the login as we are not logged in on login-module as it should as we forced token (so ok so far) 
4. if close the popup using "fermer" (close) on the top right, I get an error reporting modal!  

It looks it is because the refresh service returns an error... ok so there is a backend issue, but also the frontend should not show such a modal on a 5xx server error...

**Expected behavior**
Show the usual red toast on the top right instead of this error reporting modal

Iloveall:
## Description

Fixes #1050

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/error-report-modal-on-profile-refresh-error/en/#/groups/users/670968966872011405/personal-data)
  3. And I click on "Modify" button
  4. Then I see "LOGIN MODULE" modal
  5. And I click on "Fermer" in top right corner
  6. Then I see error notifier without sentry modal

### DIFF
--- a/src/app/shared/services/user-session.service.ts
+++ b/src/app/shared/services/user-session.service.ts
@@ -63,6 +63,7 @@ export class UserSessionService implements OnDestroy {
     const refresh$ = this.currentUserService.refresh().pipe(shareReplay(1));
     refresh$.subscribe({
       next: () => this.userProfileUpdated$.next(),
+      // Required for avoid to show Sentry modal
       error: () => {},
     });
     return refresh$;
--- a/src/app/shared/services/user-session.service.ts
+++ b/src/app/shared/services/user-session.service.ts
@@ -63,7 +63,7 @@ export class UserSessionService implements OnDestroy {
     const refresh$ = this.currentUserService.refresh().pipe(shareReplay(1));
     refresh$.subscribe({
       next: () => this.userProfileUpdated$.next(),
-      // Required for avoid to show Sentry modal
+      // error has to be handled in the caller of the `refresh()` function 
       error: () => {},
     });
     return refresh$;
--- a/src/app/shared/services/user-session.service.ts
+++ b/src/app/shared/services/user-session.service.ts
@@ -61,7 +61,10 @@ export class UserSessionService implements OnDestroy {
 
   refresh(): Observable<void> {
     const refresh$ = this.currentUserService.refresh().pipe(shareReplay(1));
-    refresh$.subscribe(() => this.userProfileUpdated$.next());
+    refresh$.subscribe({
+      next: () => this.userProfileUpdated$.next(),
+      error: () => {},
+    });
     return refresh$;
   }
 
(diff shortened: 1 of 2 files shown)


src/app/shared/services/user-session.service.ts
```diff
@@ -61,7 +61,10 @@ export class UserSessionService implements OnDestroy {
 
   refresh(): Observable<void> {
     const refresh$ = this.currentUserService.refresh().pipe(shareReplay(1));
-    refresh$.subscribe(() => this.userProfileUpdated$.next());
+    refresh$.subscribe({
+      next: () => this.userProfileUpdated$.next(),
+      error: () => {},
```
ThomasJuster, line 66: Can you add a comment indicating why it's necessary to have an empty function here?

ThomasJuster (commented):
:+1:

ThomasJuster (commented):
:heavy_check_mark:

src/app/shared/services/user-session.service.ts
```diff
@@ -61,7 +61,11 @@ export class UserSessionService implements OnDestroy {
 
   refresh(): Observable<void> {
     const refresh$ = this.currentUserService.refresh().pipe(shareReplay(1));
-    refresh$.subscribe(() => this.userProfileUpdated$.next());
+    refresh$.subscribe({
+      next: () => this.userProfileUpdated$.next(),
+      // Required for avoid to show Sentry modal
+      error: () => {},
```
smadbe, line 67: @ThomasJuster why not doing the "usual"
`if (!(err instanceof HttpErrorResponse)) throw err;` here?
Don't we have the same case as usual where if it is not a request error but a JS one, we want to report it ?

src/app/shared/services/user-session.service.ts
```diff
@@ -61,7 +61,11 @@ export class UserSessionService implements OnDestroy {
 
   refresh(): Observable<void> {
     const refresh$ = this.currentUserService.refresh().pipe(shareReplay(1));
-    refresh$.subscribe(() => this.userProfileUpdated$.next());
+    refresh$.subscribe({
+      next: () => this.userProfileUpdated$.next(),
+      // Required for avoid to show Sentry modal
```
smadbe, line 66: instead:
```suggestion
      // error is handled in the function (`refresh()`) caller 
```

src/app/shared/services/user-session.service.ts
smadbe: Ok. fixed the comment to make it clear.

smadbe (approved):
Just fix this comment and that's ok for me. Have you tested it? (with the forced_token?)

Iloveall:
> Just fix this comment and that's ok for me. Have you tested it? (with the forced_token?)

Yes, of course

PR #1062, opened 2022-06-07
Fixes log view for current user

### LINKED ISSUE #1059: User's progress: error while seeing one's own progress
**Describe the bug**
Most user cannot see their own progress. They get an error instead.

**To Reproduce**
1. As user usr_lydk36y8j9 (token ending with `456`) 
2. Go to your own progress page https://dev.algorea.org/fr/#/groups/users/320223660724875443
3. See the fetching error

The reason for that error is that `watched_group_id` parameter is given for every request, but in the case of this user (unlike the usual user) the user is not in a group he can watch. (while of course, a user can always see his own progress)

So the solution is just not to provide `watched_group_id` parameter to the service when a user is checking his own profile.

Iloveall:
## Description

Fixes #1059

...

## Test cases

- [x] Case 1:
  1. As user usr_lydk36y8j9 (token ending with `456`)
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/user-progress-error-while-seeing-own-progress/en/#/groups/users/320223660724875443)
  3. Then I see progress list

- [x] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/user-progress-error-while-seeing-own-progress/en/#/groups/users/752024252804317630;path=5121055722873780306)
  3. Then I see progress list

### DIFF
--- a/src/app/modules/group/components/group-log-view/group-log-view.component.ts
+++ b/src/app/modules/group/components/group-log-view/group-log-view.component.ts
@@ -28,10 +28,10 @@ export class GroupLogViewComponent implements OnChanges, OnDestroy {
   @ViewChild('op') op?: OverlayPanel;
   @ViewChildren('contentRef') contentRef?: QueryList<ElementRef<HTMLElement>>;
 
-  private readonly groupId$ = new ReplaySubject<string>(1);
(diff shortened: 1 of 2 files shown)


ThomasJuster (approved):
:+1:

PR #1075, opened 2022-06-23
Fixes Composition tab

### LINKED ISSUE #1073: No tab when navigating to a group I am just member of
**Describe the bug**
When I navigate from the composition tab of a group to another group for which I can only see one tab (so, for which the tab bar is not shown by default), I have no tab to go the main group page.

**To Reproduce**
1. As the usual user
2. (could come from the [composition page of "Pixal" group](https://dev.algorea.org/en/#/groups/by-id/672913018859223173;path=52767158366271444/details/members) but let's go directly to the buggy page)
3. Go to the [non-allowed composition page of DemoSNT](https://dev.algorea.org/en/#/groups/by-id/6884210558506774905;path=/details/members)
4. I have no tab to go to the main "overview" tab of this group.
 
**Expected behavior**
When we are on a not-allowed tab, this current page tab should be shown and so the tab bar, so that I know on which tab I am and so that I can navigate back to the "overview" tab.

Iloveall:
## Description

Fixes #1073

...

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/no-tab-when-navigating-to-group/en/#/groups/by-id/6884210558506774905;path=/details/members)
  3. Then I see the message "You are not allowed to see this page." and "Composition" tab
  4. Then I click "Overview" tab
  5. Then I see "Composition" is disappeared

### DIFF
--- a/src/app/modules/group/pages/group-details/group-details.component.html
+++ b/src/app/modules/group/pages/group-details/group-details.component.html
@@ -14,7 +14,7 @@
     >
     </alg-group-indicator>
     <!-- tabs -- if only the first tab is visible, do not show the tab bar -->
-    <div [hidden]="!group.isCurrentUserManager && group.currentUserMembership === 'none' && !adminTab.isActive && !accessTab?.isActive && !compositionTab.isActive">
+    <div [hidden]="!group.isCurrentUserManager && group.currentUserMembership === 'none' && !adminTab.isActive && !accessTab?.isActive && !compositionTab.isActive && !settingsTab.isActive">
       <div class="nav-tab">
         <a
           class="nav-tab-item"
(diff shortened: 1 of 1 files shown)


src/app/modules/group/pages/group-details/group-details.component.html
```diff
@@ -14,7 +14,7 @@
     >
     </alg-group-indicator>
     <!-- tabs -- if only the first tab is visible, do not show the tab bar -->
-    <div [hidden]="!group.isCurrentUserManager && group.currentUserMembership === 'none' && !adminTab.isActive && !accessTab?.isActive">
+    <div [hidden]="!group.isCurrentUserManager && group.currentUserMembership === 'none' && !adminTab.isActive && !accessTab?.isActive && !compositionTab.isActive">
```
ThomasJuster, line 17: I think the settings tab can be included as well:
https://dev.algorea.org/branch/bugfix/no-tab-when-navigating-to-group/en/#/groups/by-id/6884210558506774905;path=/details/settings

ThomasJuster (approved):

src/app/modules/group/pages/group-details/group-details.component.html

PR #1096, opened 2022-07-08
Item permission models

smadbe:
## Description

Create item permissions models and use it across the app so that:
- strings for item perms are not used anymore (error prone)
- the higher-level logic are all implemented at a same "centralized" place 

## Test cases

A test showing that one of the major part isn't broken.

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [SNT chapter page while observing the "UE Info 2019-2020" group](https://dev.algorea.org/branch/perm-models/en/#/activities/by-id/1352246428241737349;path=4702;parentAttempId=0/details/progress/chapter?watchedGroupId=472375073987524072&watchUser=0)
  3. And I click on the top-left cell > access
  4. And I change the view permission, and click "proceed"
  6. Then I see it works


src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
```diff
@@ -32,10 +36,10 @@ export class PermissionsEditDialogComponent implements OnChanges, OnDestroy {
...
 
   form = this.fb.group({
-    canView: [ 'none' ],
-    canGrantView: [ 'none' ],
-    canWatch: [ 'none' ],
-    canEdit: [ 'none' ],
+    canView: [ ITEMVIEWPERM_NONE ],
+    canGrantView: [ ITEMGRANTVIEWPERM_NONE ],
+    canWatch: [ ITEMWATCHPERM_NONE ],
+    canEdit: [ ITEMEDITPERM_NONE ],
```
ThomasJuster, line 42: If we have a `_MAX` approach upper, shouldn't we have a `_MIN` approach here?
Unless the lowest level is always `'none'`

src/app/shared/http-services/group-permissions.service.ts
```diff
@@ -5,25 +5,23 @@ import { map } from 'rxjs/operators';
...
+  itemCorePermDecoder,
+  D.intersect(itemSessionPermDecoder),
+  D.intersect(itemEntryFromPermDecoder),
 );
 
 const groupPermissionsInfoDecoder = D.struct({
-  granted: pipe(groupPermissionsDecoder, D.intersect(D.struct({
-    canEnterUntil: dateDecoder
-  }))),
+  granted: D.intersect(groupPermissionsDecoder)(itemEntryUntilPermDecoder),
```
ThomasJuster, line 24: In the past you made a remark about using `D.intersect` with `pipe`

src/app/shared/models/domain/item-edit-permission.ts
```diff
@@ -0,0 +1,56 @@
...
+  CHILDREN = 'children',
+  ALL = 'all',
+  ALL_WITH_GRANT = 'all_with_grant';
+
+export const
+  ITEMEDITPERM_NONE = NONE,
+  ITEMEDITPERM_CHILDREN = CHILDREN,
+  ITEMEDITPERM_ALL = ALL,
+  ITEMEDITPERM_ALL_WITH_GRANT = ALL_WITH_GRANT,
+  ITEMEDITPERM_MAX = ALL_WITH_GRANT;
```
ThomasJuster, line 16: I suppose you have considered using enums and dropped the idea, for my information, why?

In my opinion uppercase variable names like these are not very readable.
As I understand it, the idea is to follow the constant naming convention, that would give `ITEM_EDIT_PERM_*` instead which makes very long names now and tons of imports in some other files
I would rather picture exporting an object `itemEditPermissions` - or `itemEditPerm`, `editPermissions`, `editPerm`, whatever - containing the values

src/app/shared/models/domain/item-edit-permission.ts
```diff
@@ -0,0 +1,56 @@
...
+
+@Pipe({ name: 'allowsEditingChildren', pure: true })
+export class AllowsEditingChildrenItemPipe implements PipeTransform {
+  transform = allowsEditingChildren;
+}
+
+@Pipe({ name: 'allowsEditingAll', pure: true })
+export class AllowsEditingAllItemPipe implements PipeTransform {
+  transform = allowsEditingAll;
+}
```
ThomasJuster, line 56: About pipes, since angular template system type-checking differs from the JS one and understands badly optional chaining, I think it would ease using those pipe if they tolerate nil value (and when nil they can return `false`)
That way we could use them as `*ngIf="itemData?.permissions | allowsEditingAll"`

src/app/shared/http-services/group-permissions.service.ts
ThomasJuster: Same here: https://github.com/France-ioi/AlgoreaFrontend/pull/1096/files#diff-627a1ba3e4c45c92a1eea14c11165a322a2f9204452bb22f2992e4ea9498e38eR25

src/app/shared/models/domain/item-edit-permission.ts
ThomasJuster: FYI, you can also use the `export as` notation:
```ts
export {
   NONE as ITEMEDITPERM_NONE,
 }
```
I'm not sure it makes much a difference though :thinking:

ThomasJuster (commented):
Nice splitting of concerns
I find particularly readable the idea of having many dead-simple expressive functions :+1:

src/app/modules/item/components/permissions-edit-dialog/permissions-edit-dialog.component.ts
smadbe: yeah I assume the min was quite straightforward (cannot have a new lower perm).. while max is not and may still change in the future

src/app/shared/http-services/group-permissions.service.ts
smadbe: With 2, I found them more readable like that than with a pipe. You prefer I use a pipe?

src/app/shared/models/domain/item-edit-permission.ts
smadbe: It was done on purpose, I did not want to accept falsy values as it ends up keeping bugs silenced.

Actually, I forgot to implement an extra commit that I wanted to implement last: add an assertion in these pipes, in dev throw an exception if falsy, and prod console.err if so.

What do you think about that?

src/app/shared/http-services/group-permissions.service.ts
ThomasJuster: I personally don't care

src/app/shared/models/domain/item-edit-permission.ts
ThomasJuster: Not much, to me both approaches are valid, as long as we use one that works it's ok.

ThomasJuster (approved):

PR #1098, opened 2022-07-08
Fixes sort for all descendants

### LINKED ISSUE #1083: Error on listing group descendants
**Describe the bug**
I get an error on listing group descendants. This seems to be due to a sort parameter which is empty.

**To Reproduce**
1. As the usual user
2. Go to [any group I can manage](https://dev.algorea.org/en/#/groups/by-id/120505918794412400;path=6884210558506774905/details/members)
3. Click on 'All descendants'
4. The list in in error

Iloveall:
## Description

Fixes #1083

...

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/error-on-listing-group-descendants/en/#/groups/by-id/120505918794412400;path=6884210558506774905/details/members)
  3. And I click on "All Descendants"
  4. Then I see no any errors

### DIFF
--- a/src/app/shared/http-services/get-group-descendants.service.ts
+++ b/src/app/shared/http-services/get-group-descendants.service.ts
@@ -57,7 +57,7 @@ export class GetGroupDescendantsService {
 
   getUserDescendants(groupId: string, options: { sort?: string[], limit?: number, fromId?: string }): Observable<UserDescendant[]> {
     let params = new HttpParams();
-    if (options.sort) params = params.set('sort', options.sort.join(','));
+    if (options.sort && options.sort.length > 0) params = params.set('sort', options.sort.join(','));
     if (options.limit !== undefined) params = params.set('limit', options.limit);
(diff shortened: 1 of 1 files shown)


src/app/shared/http-services/get-group-descendants.service.ts
```diff
@@ -57,7 +57,7 @@ export class GetGroupDescendantsService {
```
ThomasJuster, line 57: When switching between "direct children" and "all descendants", requests are triggered twice, can you have a look?
image

If it's straightforward to fix, can you fix it, and if not, create an issue?

ThomasJuster (approved):
:+1:

src/app/shared/http-services/get-group-descendants.service.ts
Iloveall: Already have [issue](https://github.com/France-ioi/AlgoreaFrontend/pull/1098) for this

PR #1116, opened 2022-07-29
Feature/item dependency allow adding

### LINKED ISSUE #1066: Item dependency: allow adding
## Overview
In the item dependency tab, allow adding new dependencies.

Require #1064 to be completed.

## Acceptance criteria

In the "Dependencies" tab:
- [ ] Under the list, add the component to add content (as the one which is used under the chapter children list) with only the possibility to search for content (no content creation)
- [ ] In search result, the items which are already in the list below are greyed out with "already added" button (the same behavior as for the chapter children)
- [ ] When "add" is pressed, the ["add dependency" service](https://franceioi-algorea.s3.eu-west-3.amazonaws.com/spec/v1.2.0.html#operation/itemDependencyCreate) is called with `grant_content_view=true`, the current item as the `dependent_item_id` and the searched item as `prerequisite_item_id`. 
- [ ] no action on dependencies should be possible while processing
- [ ] on success/failure: display the green/red toast, re-enable frozen buttons and refresh the list

Iloveall:
## Description

Fixes #1066

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/item-dependency-allow-adding/en/#/activities/by-id/4728890134807328169;path=5846722634000078615;attempId=0/details/dependencies)
  3. And I search for content
  4. Then I click on Add button
  5. Then I see success notification and updated prerequisites
  6. And I do search for the same item
  7. Then I see the item with button caption is "Already Added"

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/item-dependency-allow-adding/en/#/activities/by-id/4728890134807328169;path=5846722634000078615;attempId=0/details/dependencies)
  3. And I search for content
  4. Then I switch to offline mode in network browser tab
  5. Then I click on Add button
  8. Then I see an error notification

### DIFF
--- a/src/app/modules/item/components/add-dependency/add-dependency.component.html
+++ b/src/app/modules/item/components/add-dependency/add-dependency.component.html
@@ -1,0 +1,11 @@
+<alg-section icon="pi pi-plus" i18n-label label="Create a new dependency">
+  <alg-add-content
+    [allowedTypesForNewContent]="allowedNewItemTypes"
+    [searchFunction]="searchFunction"
+    [addedIds]="addedIds"
+    [loading]="false"
+    [showCreateUI]="false"
+    (contentAdded)="addDependency($event)"
+    inputCreatePlaceholder="Enter a title to create a new dependency"
+  ></alg-add-content>
+</alg-section>
--- a/src/app/modules/item/components/add-dependency/add-dependency.component.html
+++ b/src/app/modules/item/components/add-dependency/add-dependency.component.html
@@ -6,6 +6,6 @@
     [loading]="false"
     [showCreateUI]="false"
     (contentAdded)="addDependency($event)"
-    inputCreatePlaceholder="Enter a title to create a new dependency"
+    i18n-inputCreatePlaceholder inputCreatePlaceholder="Enter a title to create a new dependency"
   ></alg-add-content>
 </alg-section>
--- a/src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
+++ b/src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
@@ -23,7 +23,7 @@ export class ItemDependenciesComponent implements OnChanges, OnDestroy {
   @ViewChild('op') op?: OverlayPanel;
   @ViewChildren('contentRef') contentRef?: QueryList<ElementRef<HTMLElement>>;
 
-  readonly itemId$ = new ReplaySubject<string>(1);
+  private readonly itemId$ = new ReplaySubject<string>(1);
   private readonly refresh$ = new Subject<void>();
 
   state$ = this.itemId$.pipe(
--- a/src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
+++ b/src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
@@ -89,11 +89,11 @@ export class ItemDependenciesComponent implements OnChanges, OnDestroy {
 
   addDependency(item: AddedContent<ItemType>): void {
     if (!item.id) {
-      throw new Error('Unexpected: Missed item ID');
+      throw new Error('Unexpected: item id is missing');
     }
     const dependentItemId = this.itemData?.item.id;
     if (!dependentItemId) {
-      throw new Error('Unexpected: Missed dependent item id');
+      throw new Error('Unexpected: dependent item id is missing');
     }
     this.createInProgress$.next(true);
     this.addItemPrerequisiteService.create(dependentItemId, item.id).subscribe({
--- a/src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
+++ b/src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
@@ -1,11 +1,16 @@
 import { Component, ElementRef, Input, OnChanges, OnDestroy, QueryList, ViewChild, ViewChildren } from '@angular/core';
 import { GetItemPrerequisitesService } from '../../http-services/get-item-prerequisites.service';
 import { BehaviorSubject, debounceTime, merge, ReplaySubject, Subject, switchMap } from 'rxjs';
-import { mapToFetchState } from '../../../../shared/operators/state';
-import { distinctUntilChanged, filter, map, shareReplay } from 'rxjs/operators';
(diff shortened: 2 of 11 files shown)


src/app/modules/item/components/add-dependency/add-dependency.component.html
```diff
@@ -0,0 +1,11 @@
+<alg-section icon="pi pi-plus" i18n-label label="Create a new dependency">
```
ThomasJuster, line 1: When I disable my network connection and then search for content, the UI appears to be loading infinitely.
And the component does not seem to recover from the error because when I re-enable my network and search for something else, no xhr request is sent.
[link](https://dev.algorea.org/branch/feature/item-dependency-allow-adding/en/#/activities/by-id/4728890134807328169;path=5846722634000078615;attempId=0/details/dependencies)
image

ThomasJuster, line 1: In terms of UX, I think the search input should be cleared only when the dependency is properly added, not before. That way if anything goes wrong, the user can retry his search very easily.

src/app/modules/item/components/add-dependency/add-dependency.component.html
```diff
@@ -0,0 +1,11 @@
...
+  <alg-add-content
+    [allowedTypesForNewContent]="allowedNewItemTypes"
+    [searchFunction]="searchFunction"
+    [addedIds]="addedIds"
+    [loading]="false"
+    [showCreateUI]="false"
+    (contentAdded)="addDependency($event)"
+    inputCreatePlaceholder="Enter a title to create a new dependency"
```
ThomasJuster, line 9: I think this label `inputCreatePlaceholder` should be translated, right?

src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
```diff
@@ -74,4 +86,27 @@ export class ItemDependenciesComponent implements OnChanges, OnDestroy {
...
     this.refresh$.next();
   }
+
+  addDependency(item: AddedContent<ItemType>): void {
+    if (!item.id) {
+      throw new Error('Unexpected: Missed item ID');
+    }
+    const dependentItemId = this.itemData?.item.id;
+    if (!dependentItemId) {
+      throw new Error('Unexpected: Missed dependent item id');
```
ThomasJuster, line 96: For the error messages, I'd say: `throw new Error('Unexpected: item id is missing')` and `throw new Error('Unexpected: dependent item id is missing')`

src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
```diff
@@ -37,7 +43,13 @@ export class ItemDependenciesComponent implements OnChanges, OnDestroy {
     data ? this.op?.toggle(data.event, data.target) : this.op?.hide();
   });
 
-  constructor(private getItemPrerequisitesService: GetItemPrerequisitesService) {
+  createInProgress$ = new BehaviorSubject<boolean>(false);
```
ThomasJuster, line 46: Here when looking at the usage of this subject, I think a boolean instead of a subject is enough

src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
```diff
@@ -18,15 +23,16 @@ export class ItemDependenciesComponent implements OnChanges, OnDestroy {
   @ViewChild('op') op?: OverlayPanel;
   @ViewChildren('contentRef') contentRef?: QueryList<ElementRef<HTMLElement>>;
 
-  private readonly itemId$ = new ReplaySubject<string>(1);
+  readonly itemId$ = new ReplaySubject<string>(1);
```
ThomasJuster, line 26: I think this subject can stay private, can you check?

ThomasJuster (commented):
:+1:

src/app/modules/item/components/add-dependency/add-dependency.component.html
Iloveall: Fixed. Also I've fixed it for `add-group.component`

src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
Iloveall: Fixed

src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
Iloveall: Fixed

src/app/modules/item/components/item-dependencies/item-dependencies.component.ts
Iloveall: Fixed

src/app/modules/item/components/add-dependency/add-dependency.component.html
Iloveall: I'm totally agree. Probably we should have another issue as this component `AddContentComponent` is old and used in many places.

https://github.com/France-ioi/AlgoreaFrontend/blob/853750469f8906b5793f2ee30cbc5d154d0d29b9/src/app/modules/shared-components/components/add-content/add-content.component.ts#L95

Iloveall: I guess it should be separated issue, as all logic in `AddContentComponent`, used in many places, and seems like it requires to add one more state 'error'

https://github.com/France-ioi/AlgoreaFrontend/blob/853750469f8906b5793f2ee30cbc5d154d0d29b9/src/app/modules/shared-components/components/add-content/add-content.component.ts#L30

smadbe: I've created a https://github.com/France-ioi/AlgoreaFrontend/issues/1129 for it.

smadbe: I've created the issue: https://github.com/France-ioi/AlgoreaFrontend/issues/1130

smadbe (approved):
The UI/UX looks ok. I've checked the changes that Thomas asked as well.

PR #1136, opened 2022-09-02
Adds placeholder for score while its unknown

### LINKED ISSUE #1053: Use a placeholder for score while it is unknown
## Overview

On activity pages, we have the title jumping on the right because the score is not known initially and then arrive with the result. Let's try to improve this behavior.

https://user-images.githubusercontent.com/1053150/171203029-30b34ea8-33bb-4620-b173-407bee22754d.mov

I don't think it requires a placeholder the score usually arrives very quickly.

## Acceptance criteria
- [ ] on task and chapter, save the space for the score (empty at first then) so that the title do not jump when score is known
- [ ] if the user cannot view item content (can_view<=content), display a "lock" icon instead of the score (same size)
- [ ] there is still a case when no result is received at the end (on explicit-entry item, if the user has not entered the item yet)... in this case, this area will remain empty for now, that's not a problem
- [ ] make sure skills are not affected by this change

Iloveall:
## Description

Fixes #1053

...

## Test cases

- [ ] Case 1:
  1. Given I am non auth user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/use-placeholder-for-score-while-it-is-unknown/en/#/activities/by-id/1980584647557587953;path=4702,4102;parentAttempId=0/details)
  3. Then I see lock icon for score ring

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/use-placeholder-for-score-while-it-is-unknown/en/#/activities/by-id/4702;path=;attempId=0/details)
  3. Then I see score ring without jumping while loading

- [ ] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/use-placeholder-for-score-while-it-is-unknown/en/#/skills/by-id/3000;path=;attempId=0/details)
  3. Then I see no score ring

### DIFF
--- a/src/app/modules/item/components/item-header/item-header.component.html
+++ b/src/app/modules/item/components/item-header/item-header.component.html
@@ -1,12 +1,21 @@
 <div class="task-header-container" *ngIf="itemData">
   <div class="task-header">
-    <alg-score-ring
-      *ngIf="itemData.item.type !== 'Skill' && itemData.currentResult"
-      [bestScore]="itemData.item.bestScore"
-      [currentScore]="itemData.currentResult.score"
-      [isValidated]="itemData.currentResult.validated"
-      [diameter]="70"
-    ></alg-score-ring>
+    <div class="task-score-ring" *ngIf="itemData.item.type !== 'Skill'">
+      <ng-container *ngIf="itemData.item.permissions | allowsViewingContent; else lock">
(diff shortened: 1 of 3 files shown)


smadbe (approved):
looks good 👍

PR #1141, opened 2022-09-14
fix former answer 403 response

ThomasJuster:
## Description

Fixes #...

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  3. And I click on ...
  4. Then I see ...
  5. And

- [ ] Case 2:
  1. Given I am the usual user
  2. When ...


smadbe:
As template error should not throw error (`ngIf` should handle correctly error)... I don't think this "fix" is useful... at least as long as I don't understand why.

smadbe:
This solution breaks the `formerAnswerError$` which use it and display a proper msg in case of error.

PR #1158, opened 2022-10-25
Fixes pending changes guard

### LINKED ISSUE #1151: Bug: on item children edit: do not navigation to group page after "lose change" confirmation
**Describe the bug**

If I edit item children and try to navigate to a group page, I get the "are sure to lose changes" dialog, but if I click yes, it does not navigate.

**To Reproduce**
1. As the usual user
1. I go to an [item edit page](https://dev.algorea.org/en/activities/by-id/6841500513635377116;path=;attempId=0/details/edit-children)
2. I change the order of children (without saving)
3. I click on the "groups" tab in the left menu and click on a group
4. I get the "This page has unsaved changes. Do you want to leave this page and lose its changes?" dialog
5. If I click "Yes, leave this page",  nothing happen (no navigation)

Iloveall:
## Description

Fixes #1151

In some cases guard declares 2 times, that's why for leave the page need to emit true 2 times (so in 2 guards), but we have only one confirmation modal (from one of guard). So it emits for first guard but doesn't for second.

I've created common state for fix this problem.

## Test cases

- [x] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/does-not-navigate-from-item-to-group/en/#/groups/by-id/5121055722873780306;path=/details/settings)
  3. Update the description
  4. Click on "4LTDI1819" in left menu
  5. Then I see confirmation modal
  6. And I click on "Yes, leave page"
  7. Then I see settings of "4LTDI1819" group page

- [x] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/does-not-navigate-from-item-to-group/en/#/groups/by-id/5121055722873780306;path=/details/settings)
  3. Update the description
  4. Click on "4LTDI1819" in left menu
  5. Then I see confirmation modal
  6. And I click on "No"
  7. Then I see settings of "A super new group" group page

- [x] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/does-not-navigate-from-item-to-group/en/#/groups/by-id/5121055722873780306;path=/details/settings)
  3. Update the description
  4. Click on "Activities" -> "Parcours officiels" in left menu
  5. Then I see confirmation modal
  6. And I click on "Yes, leave page"
  7. Then I see "Parcours officiels" item page

- [x] Case 4:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/does-not-navigate-from-item-to-group/en/#/activities/by-id/899800937596940136;path=;attempId=0/details)
  3. Then I click on switcher "Edit the list"
  4. Then I see edit list
  5. Then I change order of list and click on switcher "Edit the list"
  6. Then I see confirmation modal and click No
  7. Then I see edit list and switcher "Edit the list" is on

- [x] Case 5:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/does-not-navigate-from-item-to-group/en/#/activities/by-id/899800937596940136;path=;attempId=0/details)
  3. Then I click on switcher "Edit the list"
  4. Then I see edit list
  5. Then I change order of list and click on switcher "Edit the list"
  6. Then I see confirmation modal and click "Yes, leave page"
  7. Then I see Content list, switcher "Edit the list" is off

- [x] Case 6:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/does-not-navigate-from-item-to-group/en/#/activities/by-id/899800937596940136;path=;attempId=0/details)
  3. Then I click on switcher "Edit the list"
  4. Then I see edit list
  5. Then I change order of list
  6. And I click to "Groups" -> "!634"
  7. Then I see confirmation modal and click "Yes, leave page"
  8. Then I see "!634" group page

### DIFF
--- a/src/app/shared/guards/pending-changes-guard.ts
+++ b/src/app/shared/guards/pending-changes-guard.ts
@@ -1,8 +1,7 @@
 import { Injectable } from '@angular/core';
 import { ActivatedRouteSnapshot, CanDeactivate, RouterStateSnapshot } from '@angular/router';
 import { ConfirmationService } from 'primeng/api';
-import { BehaviorSubject, delay, Observable, of, take } from 'rxjs';
-import { filter, map } from 'rxjs/operators';
+import { Observable, of, Subject } from 'rxjs';
 import { PendingChangesService } from '../services/pending-changes-service';
 
 export interface PendingChangesComponent {
@@ -12,11 +11,12 @@ export interface PendingChangesComponent {
   isDirty(): boolean,
 }
 
-const dialogResponse = new BehaviorSubject<'pending' | 'accepted' | 'declined'>('pending');
-const dialogResponse$ = dialogResponse.asObservable();
-
-@Injectable()
+@Injectable({
+  providedIn: 'root'
+})
 export class PendingChangesGuard implements CanDeactivate<PendingChangesComponent> {
+
+  private dialogResponse = new Subject<boolean>();
 
   constructor(
     private confirmationService: ConfirmationService,
@@ -46,26 +46,15 @@ export class PendingChangesGuard implements CanDeactivate<PendingChangesComponen
       icon: 'pi pi-exclamation-triangle',
       acceptLabel: $localize`Yes, leave page`,
       accept: () => {
-        dialogResponse.next('accepted');
+        this.dialogResponse.next(true);
       },
       rejectLabel: $localize`No`,
       reject: () => {
-        dialogResponse.next('declined');
+        this.dialogResponse.next(false);
       },
     });
 
-    dialogResponse$.pipe(
-      filter(status => status !== 'pending'),
-      take(1),
-      delay(0),
-    ).subscribe(() =>
-      dialogResponse.next('pending')
-    );
-
-    return dialogResponse$.pipe(
-      filter(status => status !== 'pending'),
-      map(status => status === 'accepted'),
-    );
+    return this.dialogResponse;
   }
 
 }
--- a/src/app/shared/guards/pending-changes-guard.ts
+++ b/src/app/shared/guards/pending-changes-guard.ts
@@ -1,7 +1,8 @@
 import { Injectable } from '@angular/core';
 import { ActivatedRouteSnapshot, CanDeactivate, RouterStateSnapshot } from '@angular/router';
 import { ConfirmationService } from 'primeng/api';
-import { Observable, of, Subject } from 'rxjs';
+import { BehaviorSubject, delay, Observable, of, take } from 'rxjs';
+import { filter, map } from 'rxjs/operators';
 import { PendingChangesService } from '../services/pending-changes-service';
 
 export interface PendingChangesComponent {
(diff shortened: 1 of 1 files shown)


src/app/shared/guards/pending-changes-guard.ts
```diff
@@ -44,16 +46,26 @@ export class PendingChangesGuard implements CanDeactivate<PendingChangesComponen
...
       rejectLabel: $localize`No`,
       reject: () => {
-        dialogResponse.next(false);
-        dialogResponse.complete();
+        dialogResponse.next('declined');
       },
     });
-    return dialogResponse;
+
+    dialogResponse$.pipe(
```
smadbe, line 57: if the goal of the `dialogResponse$` variable is only to be used here, just use `dialogResponse` here, that's the same.

src/app/shared/guards/pending-changes-guard.ts
```diff
@@ -11,6 +12,9 @@ export interface PendingChangesComponent {
   isDirty(): boolean,
 }
 
+const dialogResponse = new BehaviorSubject<'pending' | 'accepted' | 'declined'>('pending');
```
smadbe, line 15: That's a kind of global variable? 
There should not be global variable in the app, the shared resources are supposed to use the angular DI mechanis for a similar results, in cleaner and testable way.

smadbe, line 15: For 3 cases like that, I prefer you define a enum that using the strings. Strings are just good as they can be used in templates as well, but here you do not need this case.

src/app/shared/guards/pending-changes-guard.ts
```diff
@@ -44,16 +46,26 @@ export class PendingChangesGuard implements CanDeactivate<PendingChangesComponen
...
       },
     });
-    return dialogResponse;
+
+    dialogResponse$.pipe(
+      filter(status => status !== 'pending'),
+      take(1),
+      delay(0),
+    ).subscribe(() =>
+      dialogResponse.next('pending')
```
smadbe, line 62: A subject subscription which publishes on itself... that's very smelly. That shows a "invalid" usage of RxJs.

src/app/shared/guards/pending-changes-guard.ts
```diff
@@ -44,16 +46,26 @@ export class PendingChangesGuard implements CanDeactivate<PendingChangesComponen
...
+    );
+
+    return dialogResponse$.pipe(
+      filter(status => status !== 'pending'),
+      map(status => status === 'accepted'),
+    );
```
smadbe, line 68: I don't even see what you need the pending state actually as you never output it... ah that's because you have done a BehaviorSubject which keeps and re-emits its last value... but a normal subject (correctly used) should  have been enough.

smadbe (commented):
Hum I think your usage of RxJS here is not very correct. I think I understood what you wanted to do, I will do the implementation myself and you for review.

src/app/shared/guards/pending-changes-guard.ts
Iloveall: Must be service. It was done quickly more for you to look this "proof of concept" solution

smadbe (approved):

PR #1188, opened 2022-11-21
Feature/forum thread design

### LINKED ISSUE #1186: Forum thread design
## Overview

Create a proper design for the forum discussion thread. 

## Acceptance criteria

- For all the examples I give below, these are just example, they do not have to be reproduced exactly.
- I let you choose the colors for now

### events with "message" label

- display as a "classical" chat bubble ([like here](https://cdn.dribbble.com/users/4708/screenshots/14355627/media/acaca4cbfc108f4c1107c5f013f76be4.png)), on the right when it corresponds to oneself, on the left for other users.
- for all others' message, display the name (login/name, same pattern as usual) of the user on top of the button  f
- no need to group several bubbles for now (display one once the name)
- for each bubble, display the time below the bubble ([like here](https://cdn.dribbble.com/users/2569995/screenshots/14938384/media/3a0c8e168481953c15418778618d5945.jpg))

### events with "result_started" label

- display a centered transparent bubble (such as for "Emily joined the chat" [here](https://cdn.dribbble.com/users/4708/screenshots/14355627/media/db6ce9f2cc516b8f486d16ecec0177a6.png)) 
- write "You/<watched user login/name> started the activity"
- and the time like in the example

### events with "submission" label

- same as result_started
- write "You/<watched user login/name> submitted a solution." or "You/<watched user login/name> submitted a solution which scored <score>." if there is a score.

### Send Message input/button

* Just a a simple input (not resizable) field with a consistent design with other forms.
* On its right, a simple button "pi-send" which is greyed out with the (trimmed) message is empty.
* pressing the enter key when entering a message should post it

### Window

* If the content does not fill the window vertically, everything is bottom-aligned.
* Otherwise, we are by default positioned at the bottom of the page.
* All events described above are chronologically ordered (so older messages are on top, possibly not visible without scrolling)

Iloveall:
## Description

Fixes #1186

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-thread-design/en/activities/by-id/6379723280369399253;path=4702,1625159049301502151,5207993233955027100;attempId=0/details)
  3. And I click on open thread button
  4. Then I see thread container
  5. Then I can see simple messages, scored messages, start activity message

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-thread-design/en/activities/by-id/6379723280369399253;path=4702,1625159049301502151,5207993233955027100;attempId=0/details)
  3. And I click on open thread button
  4. Then I see thread container
  5. Then I scroll down
  6. Then I can see textarea and disabled send button
  7. And I type new message and click on send button
  8. Then I see new message in list

- [ ] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-thread-design/en/activities/by-id/6379723280369399253;path=4702,1625159049301502151,5207993233955027100;attempId=0/details)
  3. And I click on open thread button
  4. Then I see thread container
  5. Then I scroll down
  6. Then I can see textarea and disabled send button
  7. And I type new message and press on Enter key
  8. Then I see new message in list

- [ ] Case 4:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-thread-design/en/activities/by-id/6379723280369399253;path=4702,1625159049301502151,5207993233955027100;attempId=0/details)
  3. And I click on open thread button
  4. Then I see thread container
  5. Then I scroll down
  6. Then I can see textarea and disabled send button
  7. And I type long message: "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum." 
  8. Then I see textarea is autoscaled 
  9. And press on Enter key
  10. Then I see new message in list
  
- [ ] Case 5:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-thread-design/en/activities/by-id/6379723280369399253;path=4702,1625159049301502151,5207993233955027100;attempId=0/details)
  3. And I click on open thread button
  4. Then I see thread container
  5. Then I scroll down
  6. Then I can see textarea and disabled send button
  7. And I press on space key few times
  8. Then I see the send button is disabled
    
- [ ] Case 6:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-thread-design/en/activities/by-id/6379723280369399253;path=4702,1625159049301502151,5207993233955027100;attempId=0/details?watchedGroupId=752024252804317630&watchUser=1)
  3. And I click on open thread button
  4. Then I see thread container
  5. Then I can see simple messages, scored messages from other users
  
- [ ] Case 7:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-thread-design/en/activities/by-id/5251664001655454910;path=6707691810849260111,7258397269994729518;attempId=0/details?watchedGroupId=752024252804317630&watchUser=1)
  3. And I click on open thread button
  4. Then I see thread container
  5. Then I can see start activity message from other users

### DIFF
--- a/src/app/modules/item/components/thread-message/thread-message.component.ts
+++ b/src/app/modules/item/components/thread-message/thread-message.component.ts
@@ -1,0 +1,34 @@
+import { Component, Input, OnChanges, OnDestroy } from '@angular/core';
+import { IncomingThreadEvent } from '../../services/threads-inbound-events';
+import { UserSessionService } from '../../../../shared/services/user-session.service';
+import { ReplaySubject, combineLatest } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+@Component({
+  selector: 'alg-thread-message[event]',
+  templateUrl: './thread-message.component.html',
+  styleUrls: [ './thread-message.component.scss' ],
+})
+export class ThreadMessageComponent implements OnChanges, OnDestroy {
+  @Input() event!: IncomingThreadEvent;
+
+  currentUser$ = this.userSessionService.userProfile$;
+  event$ = new ReplaySubject<IncomingThreadEvent>(1);
+  isMessageCreatedByCurrentUser$ = combineLatest([
+    this.currentUser$,
+    this.event$,
+  ]).pipe(
+    map(([ currentUser, event ]) => currentUser.groupId === event.createdBy),
+  );
+
+  constructor(private userSessionService: UserSessionService) {
+  }
+
+  ngOnChanges(): void {
+    this.event$.next(this.event);
+  }
+
+  ngOnDestroy(): void {
+    this.event$.complete();
+  }
+}
--- a/src/app/modules/item/components/thread-message/thread-message.component.ts
+++ b/src/app/modules/item/components/thread-message/thread-message.component.ts
@@ -12,10 +12,9 @@ import { map } from 'rxjs/operators';
 export class ThreadMessageComponent implements OnChanges, OnDestroy {
   @Input() event!: IncomingThreadEvent;
 
-  currentUser$ = this.userSessionService.userProfile$;
   event$ = new ReplaySubject<IncomingThreadEvent>(1);
   isMessageCreatedByCurrentUser$ = combineLatest([
-    this.currentUser$,
+    this.userSessionService.userProfile$,
     this.event$,
   ]).pipe(
     map(([ currentUser, event ]) => currentUser.groupId === event.createdBy),
--- a/src/app/modules/item/components/thread/thread.component.ts
+++ b/src/app/modules/item/components/thread/thread.component.ts
@@ -1,4 +1,4 @@
-import { Component } from '@angular/core';
+import { Component, ElementRef, ViewChild } from '@angular/core';
 import { ThreadService } from '../../services/threads.service';
 import { FormBuilder } from '@angular/forms';
 
@@ -8,6 +8,8 @@ import { FormBuilder } from '@angular/forms';
   styleUrls: [ './thread.component.scss' ],
 })
 export class ThreadComponent {
+  @ViewChild('messagesScroll') messagesScroll?: ElementRef<HTMLDivElement>;
+
   form = this.fb.nonNullable.group({
     messageToSend: [ '' ],
   });
@@ -26,6 +28,17 @@ export class ThreadComponent {
     this.form.reset({
       messageToSend: '',
     });
+  }
+
+  scrollDown(): void {
+    if (!this.messagesScroll) {
+      throw new Error('Unexpected: Missed scroll el');
+    }
+
+    this.messagesScroll.nativeElement.scrollTo(
+      0,
+      this.messagesScroll.nativeElement.scrollHeight - this.messagesScroll.nativeElement.offsetHeight
+    );
   }
 
 }
--- a/src/app/modules/item/components/thread/thread.component.ts
+++ b/src/app/modules/item/components/thread/thread.component.ts
@@ -1,5 +1,6 @@
 import { Component } from '@angular/core';
 import { ThreadService } from '../../services/threads.service';
+import { FormBuilder } from '@angular/forms';
 
 @Component({
   selector: 'alg-thread',
@@ -7,19 +8,24 @@ import { ThreadService } from '../../services/threads.service';
   styleUrls: [ './thread.component.scss' ],
 })
 export class ThreadComponent {
-  messageToSend = '';
+  form = this.fb.nonNullable.group({
+    messageToSend: [ '' ],
+  });
 
   state$ = this.threadService.state$;
 
   constructor(
     private threadService: ThreadService,
+    private fb: FormBuilder,
   ) {}
 
   sendMessage(): void {
-    const messageToSend = this.messageToSend.trim();
-    if (!messageToSend) return;
+    const messageToSend = this.form.value.messageToSend;
+    if (!messageToSend || !messageToSend.trim()) return;
     this.threadService.sendMessage(messageToSend);
-    this.messageToSend = '';
+    this.form.reset({
+      messageToSend: '',
+    });
   }
 
 }
--- a/src/app/modules/item/components/thread/thread.component.ts
+++ b/src/app/modules/item/components/thread/thread.component.ts
@@ -1,13 +1,16 @@
-import { Component, ElementRef, ViewChild } from '@angular/core';
+import { AfterViewInit, Component, ElementRef, OnDestroy, ViewChild } from '@angular/core';
 import { ThreadService } from '../../services/threads.service';
 import { FormBuilder } from '@angular/forms';
+import { readyData } from '../../../../shared/operators/state';
+import { Subscription, combineLatest } from 'rxjs';
+import { UserSessionService } from '../../../../shared/services/user-session.service';
 
 @Component({
   selector: 'alg-thread',
   templateUrl: './thread.component.html',
   styleUrls: [ './thread.component.scss' ],
 })
-export class ThreadComponent {
+export class ThreadComponent implements AfterViewInit, OnDestroy {
   @ViewChild('messagesScroll') messagesScroll?: ElementRef<HTMLDivElement>;
 
   form = this.fb.nonNullable.group({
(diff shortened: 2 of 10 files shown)


smadbe:
First reviewing the design. Nice 👍 

First set of changes: could you try to compact everything ? 

For classical events:
* font of "You submitted a solution which scored ..." could be a bit smaller
* all dates should be much smaller, lighter (weight), lighter (color) and with less spacing with the surroundings

Here is a small example of how it could be:
Screenshot 2022-11-22 at 11 00 30

For bubbles:
* when the message is shorted than the date, the bubble width matches the date size and the text is left aligned... it should not be
Screenshot 2022-11-22 at 11 03 28
* for size/spaces, I would expect something more like "new messages" here:
Screenshot 2022-11-22 at 11 07 45

Message sending part:
- The space with the bottom could be a few pixel only.
- probably the padding inside the text field can be much smaller (border radius can be a bit smaller as well)

The specs states: "we are by default positioned at the bottom of the page." if there is more content than one page. So we opening the panel, you should always see the "send message" part.

smadbe:
Good 👍 . Another sets of small improvements:

- There seems to be more space on top of the text in bubble than above it -> decrease a bit the space above the text (0.4rem->0.2rem) 
Screenshot 2022-11-22 at 17 32 10
- could you decrease a bit the spacing between events (from 1rem -> 0.6rem)
- dates can be in 0.8rem
- when you post a message, you should still see your "input+send button", so it should scroll down auto. It should looks like the message is just "inserted" between the input field and the previous most recent message
- isn't it possible to make the "send" icon (in the button) thinner?
- other user's name should be smaller as well. Something like this maybe:
Screenshot 2022-11-22 at 17 56 00
- line height between line of a same text can be a bit smaller. For example:
Screenshot 2022-11-22 at 18 02 29

smadbe:
I have the impression that the left blue message (left ones) have the "vertical centering problem" that I mentioned in the previous comment about white messages. Have you only fixed the white messages? You need to apply the same spacing on blue ones as well. 
Screenshot 2022-11-23 at 16 58 26

src/app/modules/item/components/thread-message/thread-message.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  @Input() event!: IncomingThreadEvent;
+
+  event$ = new ReplaySubject<IncomingThreadEvent>(1);
+  isMessageCreatedByCurrentUser$ = combineLatest([
+    this.userSessionService.userProfile$,
+    this.event$,
+  ]).pipe(
+    map(([ currentUser, event ]) => currentUser.groupId === event.createdBy),
+  );
+
```
smadbe, line 22: I think it would have be better to pass the current user (from parent component) and then to work without Rx at the level of this component... (so that we do not have 100 messages each subscribing to the profile)
Anyway, we can change that later.

src/app/modules/item/components/thread/thread.component.ts
```diff
@@ -1,25 +1,68 @@
...
+      this.state$.pipe(readyData()),
+      this.userSessionService.userProfile$,
+    ]).subscribe(([ events, currentUser ]) => {
+      const lastEvent = events[events.length - 1];
+      if (lastEvent && lastEvent.createdBy === currentUser.groupId) {
+        setTimeout(() => {
+          this.scrollDown();
+        });
+      }
+    });
```
smadbe, line 41: It means, you only scroll down if the new message is yours.
But we may consider also that anyone on the bottom of the scroll (the default behavior) just want to keep staying at the bottom and see new messages. So I think the rule should be instead: "if the user is already at the bottom of the scroll, keep him at the bottom of the scroll by scrolling when there are new events".

smadbe (changes requested):
Design: 👍 
Code: just one thing to be changed

src/app/modules/item/components/thread/thread.component.ts
Iloveall: Fixed

smadbe:
Just got this exception: Screenshot 2022-11-24 at 15 01 03

smadbe:
And:

https://user-images.githubusercontent.com/1053150/203803515-225da4e7-d1c4-46a6-ba57-43d53e5d3285.mov

Iloveall:
> Just got this exception: Screenshot 2022-11-24 at 15 01 03

I couldn't reproduce this issue and second with video, but I've improved the logic where it could be happened

smadbe:
> > Just got this exception: Screenshot 2022-11-24 at 15 01 03
> 
> I couldn't reproduce this issue and second with video, but I've improved the logic where it could be happened

You get this error when you visit a task on which there were no prior discussion I think, you get it very often. Ok, I'll check your code.

src/app/modules/item/components/thread/thread.component.ts
```diff
@@ -24,16 +24,13 @@ export class ThreadComponent implements AfterViewInit, OnDestroy {
...
-    ]).subscribe(([ events, currentUser ]) => {
-      const lastEvent = events[events.length - 1];
-      if (lastEvent && lastEvent.createdBy === currentUser.groupId) {
+    this.subscription = this.state$.pipe(readyData()).subscribe(() => {
+      if (this.messagesScroll && this.sendMessageForm && (this.messagesScroll.nativeElement.scrollHeight
+        <= (this.messagesScroll.nativeElement.scrollTop + this.messagesScroll.nativeElement.offsetHeight
+        + this.sendMessageForm.nativeElement.offsetHeight + parseInt(getComputedStyle(this.messagesScroll.nativeElement).paddingBottom)))) {
         setTimeout(() => {
           this.scrollDown();
         });
```
smadbe, line 44: This `setTimeout(() => {}` is like putting a subscribe in a subscribe.. or at least mixing Rx with another async technique.
I think this is somehow equivalent to:
```
this.subscription = this.state$.pipe(
  filter(s => s.isReady),
  filter(() => this.messagesScroll && this.se ... ment).paddingBottom)),
  delay(0),
).subscribe(() => {
  this.scrollDown();
});

**Review by smadbe — CHANGES_REQUESTED** [review]
Test using a "You submitted a solution ..." event:
https://user-images.githubusercontent.com/1053150/204251849-30f0c423-a211-40b1-8d6f-34666726a1f5.mov
... does not scroll

File: src/app/modules/item/components/thread/thread.component.ts (hunk 3 above)
Comment (Iloveall) [reply src/app/modules/item/components/thread/thread.component.ts]: Fixed

**Review by smadbe — CHANGES_REQUESTED** [review]

File: src/app/modules/item/components/thread/thread.component.ts [hunk 4]
```diff
@@ -1,25 +1,69 @@
...
+  }
+
+  ngOnDestroy(): void {
+    this.subscription?.unsubscribe();
+  }
+
   sendMessage(): void {
-    const messageToSend = this.messageToSend.trim();
-    if (!messageToSend) return;
+    const messageToSend = this.form.value.messageToSend;
```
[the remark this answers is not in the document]
Iloveall: Fixed

src/app/modules/item/components/thread/thread.component.ts
```diff
@@ -34,16 +34,13 @@ export class ThreadComponent implements AfterViewInit, OnDestroy {
...
-    ).subscribe(() => {
-      if (this.messagesScroll && this.sendMessageForm && (this.messagesScroll.nativeElement.scrollHeight
-        <= (this.messagesScroll.nativeElement.scrollTop + this.messagesScroll.nativeElement.offsetHeight
-        + this.sendMessageForm.nativeElement.offsetHeight + parseInt(getComputedStyle(this.messagesScroll.nativeElement).paddingBottom)))) {
-        setTimeout(() => {
-          this.scrollDown();
-        });
-      }
-    });
+      filter(([ events , { visible }]) => visible && events.length > 0 && (events[events.length - 1]?.label === 'submission'
```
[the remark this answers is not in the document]
Iloveall: Fixed

smadbe (approved):
👍 👍

PR #1194, opened 2022-11-25
Merge item by id details

smadbe:
Remove "details" page as there are no "edit" page anymore, so removing one unless depth level in all item-related paths.

So:
- merge item "by-id" and "details" pages
- bit of cleanup on the resulting component

Tests to be done: all kind of navigation to item pages. Looks ok but could you do it more into details @Iloveall ?


src/app/modules/item/item-routing.module.ts
```diff
@@ -3,64 +3,55 @@ import { RouterModule } from '@angular/router';
...
       component: ItemByIdComponent,
+      canDeactivate: [ BeforeUnloadGuard, PendingChangesGuard ],
       children: [
-        { path: '', redirectTo: 'details', pathMatch: 'full' },
+        { path: '', pathMatch: 'full', children: [] },
         {
-          path: 'details',
-          component: ItemDetailsComponent,
-          canDeactivate: [ BeforeUnloadGuard, PendingChangesGuard ],
-          // Children below do not use routing but there are defined here so that the router can validate the route exists
```
Iloveall, line 19: I guess this comment could be still useful

Iloveall (approved):
Went through different pages, looks good 👍

PR #1212, opened 2022-12-02
Feature/do not block UI on font load postbuild

### LINKED ISSUE #1119: Don't block UI on font load
## Overview
If analyzing the app loading (with "lighthouse" with instance), we can see that the display of the website is slowed down by the loading of some fonts which happens once the css has been loaded/parsed, which happens once the initial file has been loaded/parsed. The problem is quite common: https://web.dev/font-display/
Screenshot 2022-08-01 at 17 05 03

One way to solve that would be to [preload the blocking fonts](https://web.dev/font-display/#preload-web-fonts). That looks like the best solution to me. This requires to disable hashing for the fonts. 

Side question: why do we have so many fonts if only 3 are loaded at the end?

Iloveall:
## Description

Fixes #1119

...

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/do-not-block-ui-on-font-load-postbuild/en/activities/by-id/4702;path=;parentAttempId=0)
  3. And I click on Network tab of browser
  4. Then I see loaded fonts after first html loaded

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,7 @@ commands:
     steps:
       - run: echo "export const version = '`git describe --always`';" > ./src/version.ts
       - run: npm run injectDeployUrlForAssets --url='//assets.algorea.org<< parameters.path >><< parameters.lang >>/'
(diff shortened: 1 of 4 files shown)


smadbe (approved):
looks good 👍

PR #1216, opened 2022-12-05
forum thread: add user info and separate user-me-outsider bubbles

smadbe:
## Description

Add missing user information in the forum threads.

### Changes: 

* Use "The user  submitted a solution which ..." instead of repeating the user name each time

* Implement the fetching of missing user info  + adding metadata whether the user is the current user and is the watched user

* Bubble:
  - The trainee's messages are always on the right. They are white if we are this user, orange if we are watched.
  - The trainer's messages are always on the left. They are white if you are the user, light blue if not (another trainer).

* Do not use Rx in the thread message component anymore

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this task](https://dev.algorea.org/branch/forum-bind-user/en/activities/by-id/6379723280369399253;path=;parentAttempId=0?watchedGroupId=752024252804317630&watchUser=1)
  3. And I open the thread 
  5. Then I see the trainee's messages on the right in orange
  6. And I see my messages on the left in white
  7. And I see the other's messages on the left in blue/transparent

- [ ] Case 2:
  1. Given I am the demo user 
  2. When I go to [this same task](https://dev.algorea.org/branch/forum-bind-user/en/activities/by-id/6379723280369399253;path=;parentAttempId=0)
  3. And I open the thread 
  6. And I see my messages on the right in white
  7. And I see the other's messages on the left in blue/transparent

### DIFF
--- a/src/app/modules/item/components/thread/thread.component.html
+++ b/src/app/modules/item/components/thread/thread.component.html
@@ -7,12 +7,14 @@
   <ng-template #eventsBlock>
     <div class="widget-body">
       <div class="widget-scroll" #messagesScroll>
-        <alg-thread-message
-          class="thread-message"
-          [event]="event"
-          [userCache]="(userCache | async) || []"
-          *ngFor="let event of state.data"
-        ></alg-thread-message>
+        <ng-container *ngIf="((userCache$ | async) || []) as userCache">
+          <alg-thread-message
+            class="thread-message"
+            [event]="event"
+            [userCache]="userCache"
+            *ngFor="let event of state.data"
+          ></alg-thread-message>
+        </ng-container>
         <form class="footer-panel" [formGroup]="form" (ngSubmit)="sendMessage()">
           <textarea
             class="textarea"
--- a/src/app/modules/item/components/thread/thread.component.html
+++ b/src/app/modules/item/components/thread/thread.component.html
@@ -10,6 +10,7 @@
         <alg-thread-message
           class="thread-message"
           [event]="event"
+          [userCache]="(userCache | async) || []"
           *ngFor="let event of state.data"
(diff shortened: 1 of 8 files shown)


src/app/modules/item/components/thread/thread.component.html
```diff
@@ -10,6 +10,7 @@
         <alg-thread-message
           class="thread-message"
           [event]="event"
+          [userCache]="(userCache | async) || []"
```
Iloveall, line 13: I've found that it does subscribe as much as events count, as result I guess the problem is the same as before (subscribe in each message).

What if we move `(userCache | async)` out of *ngFor?

Iloveall (commented):
image
When you go between pages the events steam increases

Iloveall (changes requested):
Probably need to add `visible` checking in events, otherwise it does requests for users event when chat is not available:

As usual user
1. I go to this [page](https://dev.algorea.org/branch/forum-bind-user/en/activities/by-id/6379723280369399253;path=;parentAttempId=0?watchedGroupId=752024252804317630&watchUser=1)
2. And I open network tab in browser
3. Then I click "manuelIsn (Python)" in left menu
4. Then I see no chat button
5. And I see requests for user in network panel

src/app/modules/item/components/thread/thread.component.html
smadbe: You are completely right!

smadbe: fixed

smadbe:
I think everything you mentioned is fixed. Thanks for noticing them!

Iloveall (approved):
👍

PR #1229, opened 2022-12-12
fix: show score in collapsed-mode top bar when available

smadbe:
The score was not displayed anymore in the top bar when left-menu is collapsed.

Still TODO:  Fix the `ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked.` error

Scenario:

- as the usual user
- I go on the [home page](https://dev.algorea.org/branch/fix-score-in-top-bar/)
- and I collapse the left menu
- I should see the score in the top bar
- And if I navigate to other item, there is no jump in the top bar when the score is known, thanks to the placeholder.

### DIFF
--- a/src/app/core/components/content-top-bar/content-top-bar.component.html
+++ b/src/app/core/components/content-top-bar/content-top-bar.component.html
@@ -5,14 +5,15 @@
     <alg-breadcrumb [contentBreadcrumb]="currentContent?.breadcrumbs"></alg-breadcrumb>
   </div>
   <div class="full-frame-content-top-bar" *ngIf="fullFrameContent?.active || scrolled">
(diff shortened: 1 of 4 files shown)


smadbe:
@Iloveall could you please review + fix the `ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked.` which happens. It seems to complain about the `<alg-thread-container...>` while this wasn't changed in this PR.

Iloveall (approved):
I've add fix for ExpressionChangedAfterItHasBeenCheckedError error, please review.

smadbe:
Looks good. I don't get how my change impacts this but ... ok ;-)

PR #1238, opened 2022-12-20
refresh the left menu content when permissions have been updated

smadbe:
## Description

When permissions are changed, the left menu is not updated... however if some content were not available (lock icon), these permission change may impact these item (lock icon removed/added).

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/refresh-left-menu-on-perm-update/en/activities/by-id/6707691810849260111;path=;parentAttempId=0?watchedGroupId=672913018859223173&watchUser=0)
  3. And I edit permission
  4. And change the view permissions
  5. I see the left menu is refetched
  6. ... and I restore the view permission as it was before


Iloveall (approved):
👍

PR #1246, opened 2022-12-22
improve a bit the top bar layout to reduce the left margin for groups

smadbe:
## Description
improve a bit the top bar layout to reduce the left margin for groups

BEFORE
Screenshot 2022-12-22 at 16 27 54
Screenshot 2022-12-22 at 16 28 24

AFTER
Screenshot 2022-12-22 at 16 28 43
Screenshot 2022-12-22 at 16 29 02


Iloveall (approved):
👍

PR #1259, opened 2023-01-11
Fix asset distribution

smadbe:
## Description

Goals:
- fix default grid image which was not deployed
- fix that all CSS images should reference where they are actually deployed 
- don't deploy assets which are not used (stylesheet, ..) because they are put at root by angular

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. I open the network panel
  3. When I [open the app](https://dev.algorea.org/branch/fix-asset-distribution/en/activities/by-id/4702;path=;parentAttempId=0)
  4. And I can see in the network panel that font are loaded correctly, from the deployment dir
  4. And I can see in the network panel that images in html (for instance `favicon-16x16.png`) are loaded correctly, from the deployment dir
  4. And I can see in the network panel that images from css (for instance `corner-tab.svg`) are loaded correctly, from the deployment dir


Iloveall (approved):
👍

PR #1270, opened 2023-01-19
Fix answer loading in obs

smadbe:
## Description

Fix the loading of answers in observation which was incorrectly using the task token.

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [a task which has been started by another user](https://dev.algorea.org/branch/fix-answer-loading-in-obs/en/activities/by-id/1357671178979168705;path=4702,1625159049301502151,5207993233955027100;parentAttempId=0;answerId=5986937724744682977?watchedGroupId=4035378957038759250&watchUser=0)
  4. Then I see the ongoing code

- [ ] Case 2: the regular case is still working
  1. Given I am the usual user
  2. When I go to [the same task "as myself"](https://dev.algorea.org/branch/fix-answer-loading-in-obs/en/activities/by-id/1357671178979168705;path=4702,1625159049301502151,5207993233955027100;parentAttempId=0)
  3. I see my own edits.

### DIFF
--- a/src/app/modules/item/services/item-task-init.service.ts
+++ b/src/app/modules/item/services/item-task-init.service.ts
@@ -28,7 +28,12 @@ export class ItemTaskInitService implements OnDestroy {
   readonly config$ = this.configFromItem$.asObservable();
   readonly iframe$ = this.configFromIframe$.pipe(map(config => config.iframe));
   readonly taskToken$: Observable<TaskToken> = this.config$.pipe(
-    switchMap(({ attemptId, route }) => this.taskTokenService.generate(route.id, attemptId)),
+    switchMap(({ readOnly, formerAnswer, attemptId, route }) => {
+      if (readOnly && formerAnswer) return this.taskTokenService.generateForAnswer(formerAnswer.id);
+      // if we are no observing (= not in readOnly) -> we need a token for our user and the task may be edited by ourself
+      // if there are no answer loaded -> we currently want an empty task, so using our own task token
+      else return this.taskTokenService.generate(route.id, attemptId);
+    }),
     shareReplay(1),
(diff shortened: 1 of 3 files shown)


src/app/modules/item/services/item-task-init.service.ts
```diff
@@ -28,7 +28,12 @@ export class ItemTaskInitService implements OnDestroy {
   readonly config$ = this.configFromItem$.asObservable();
   readonly iframe$ = this.configFromIframe$.pipe(map(config => config.iframe));
   readonly taskToken$: Observable<TaskToken> = this.config$.pipe(
-    switchMap(({ attemptId, route }) => this.taskTokenService.generate(route.id, attemptId)),
+    switchMap(({ readOnly, formerAnswer, attemptId, route }) => {
+      if (readOnly && formerAnswer) return this.taskTokenService.generateForAnswer(formerAnswer.id);
+      // if we are no observing (= not in readOnly) -> we need a token for our user and the task may be edited by ourself
+      // if there are no answer loaded -> we currently want an empty task, so using our own task token
+      else return this.taskTokenService.generate(route.id, attemptId);
```
Iloveall, line 35: Do we need else here?

Iloveall (changes requested):

src/app/modules/item/services/item-task-init.service.ts
smadbe: Both are ok in my opinion, here there are 2 options, it is not really like the "assertion pattern" where you have a few specific cases which are handled on the top of a function and then the regular case without any "else".

Iloveall (approved):

PR #1278, opened 2023-01-23
Allows to start watching group in collapsed mode

### LINKED ISSUE #1247: Add observation button in the top bar (in collapsed mode)
## Overview
Add observation button in the top bar (in collapsed mode) so that mobile user can use the observation mode.

## Acceptance criteria
- [ ] same behavior as the current observation button, but displayed in the top bar (on the left of navigation buttons) when the user is in collapsed mode

Iloveall:
## Description

Fixes #1247

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/collapse-mode-observation-button/en/groups/by-id/5121055722873780306;path=)
  3. And I click on collapse left menu
  4. Then I see "Start Watch" button in top bar
  5. And I click on "Start Watch" button
  6. Then I see suggested list
  7. And I click on first item of suggested list
  8. Then I see tooltip is disappeared 

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/collapse-mode-observation-button/en/groups/by-id/5121055722873780306;path=)
  3. And I click on collapse left menu
  4. Then I see "Start Watch" button in top bar
  5. And I click on left arrow of neighbours widget
  6. Then I "Start Watch" button is disappeared
  7. And I click on right arrow of neighbours widget
  8. Then I "Start Watch" button is appeared

- [ ] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/collapse-mode-observation-button/en/groups/by-id/5121055722873780306;path=)
  3. And I click on collapse left menu
  4. Then I see "Start Watch" button in top bar
  5. And I click on "Start Watch" button
  6. Then I see observing mode
  8. Then I click on "Started Watch" button
  9. Then I see observing mode is stopped

- [ ] Case 4:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/collapse-mode-observation-button/en/groups/by-id/5121055722873780306;path=)
  3. And I click on collapse left menu
  4. Then I see "Start Watch" button in top bar
  5. And I click on "Start Watch" button
  6. And I click on right arrow of neighbours widget
  7. Then I "Start Watch" button is not highlighted 
  8. And I click on left arrow of neighbours widget
  10. Then I "Start Watch" button is highlighted


src/app/core/services/group-watching.service.ts
```diff
@@ -20,6 +20,12 @@ export interface WatchedGroup {
   currentUserCanGrantAccess: boolean,
 }
 
+export interface StartWatchGroupInfo {
+  id: string,
+  name: string,
+  currentUserCanGrantGroupAccess?: boolean,
```
smadbe, line 26: why optional?

src/app/core/components/content-top-bar/content-top-bar.component.ts
```diff
@@ -44,6 +47,27 @@ export class ContentTopBarComponent {
     filter(navigationNeighbors => !!navigationNeighbors?.isReady),
   );
 
+  allowWatchingGroup$ = this.currentContent$.pipe(
```
smadbe, line 50: I suppose most of what you have added here is repeated from the other place where this button appears... so there is a lot of code repeated. Probably it would have been wise to place that in a new component used at 2 places.

smadbe (changes requested):

src/app/core/app.module.ts
```diff
@@ -81,6 +83,8 @@ const DEFAULT_PERFECT_SCROLLBAR_CONFIG: PerfectScrollbarConfigInterface = {
     MenuModule,
     FormsModule,
     ConfirmPopupModule,
+    OverlayPanelModule,
+    GroupModule,
```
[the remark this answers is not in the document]
Iloveall: Fixed

src/app/core/services/group-watching.service.ts
Iloveall: Fixed

src/app/core/components/content-top-bar/content-top-bar.component.ts
Iloveall: Please, look the last commit:

I haven't found the way how current content could be used in page navigator, I suppose when we will move current content into for user page into `UserInfo` the page navigator will use the same approach as `GroupInfo` and then params of page navigator could be clarified

src/app/core/components/content-top-bar/content-top-bar.component.html
```diff
@@ -26,6 +26,18 @@
...
       (click)="toggleThreadVisibility(!discussion.visible)"
     ></button>
+    <ng-container *ngIf="allowWatchingGroup$ | async">
+      <alg-watch-button
+        [ngClass]="{ 'margin-right': !(navigationNeighbors$ | async) }"
+        [route]="groupInfo.route"
+        [startWatchGroupInfo]="{
+          id: groupInfo.route.id,
+          name: groupInfo.title || '',
+          currentUserCanGrantGroupAccess: !!groupInfo.currentUserCanGrantGroupAccess
```
smadbe, line 36: These two cannot be undefined in practice if `allowWatchingGroup` is known... This should be the case "by design".
Actually, group-info should have a optional "details" property like items, which would have `name`, `currentUserCanGrantGroupAccess` and `currentUserCanWatchGroup` as non optional. This way, you would pass group info detail full or undefined.
`groupInfo$` should not be separated from `allowWatchingGroup$` at the end

smadbe (changes requested):

src/app/core/components/content-top-bar/content-top-bar.component.html
Iloveall: Fixed

src/app/modules/group/components/group-header/group-header.component.html
```diff
@@ -7,9 +7,14 @@
     </div>
     <ng-container *ngIf="navigationNeighbors$ | async as navigationNeighbors">
       <alg-page-navigator
-        *ngIf="!!groupWithManagement"
+        *ngIf="!!groupWithManagement && groupData"
```
smadbe, line 10: `groupWithManagement` is supposed to be `groupData` augmented with some management info... so `groupData` is not supposed to be used in the template and you are supposed to be able to use `groupWithManagement` everywhere you would use `groupData`

src/app/modules/shared-components/components/page-navigator/page-navigator.component.html
```diff
@@ -1,6 +1,11 @@
 <div class="page-nav">
+  <alg-watch-button
+    [route]="route"
+    [startWatchGroupInfo]="startWatchGroupInfo"
+    *ngIf="allowWatching && startWatchGroupInfo"
+  ></alg-watch-button>
   <span
-    *ngIf="allowWatching"
```
smadbe, line 3: What is it? Is it the button to stop watching? The button to start watching is a specific component while the one to stop is different is managed by the parent? I don't get it.

src/app/modules/shared-components/components/page-navigator/page-navigator.component.ts
```diff
@@ -9,6 +11,8 @@ export class PageNavigatorComponent {
   @Input() allowWatching = false;
   @Input() isWatched = false;
   @Input() allowFullScreen = false;
+  @Input() startWatchGroupInfo?: StartWatchGroupInfo;
+  @Input() route?: GroupRoute;
```
smadbe, line 15: A "route" argument for a generic `PageNavigatorComponent` which is a `GroupRoute`... 
Ok, answer the previous questions first but probably this one will have to be fixed as well.

src/app/core/components/content-top-bar/content-top-bar.component.ts
```diff
@@ -20,6 +21,8 @@ export class ContentTopBarComponent {
   @Input() fullFrameContent?: FullFrameContent;
   @Input() scrolled = false;
 
+  @ViewChild('op') op?: OverlayPanel;
```
smadbe, line 24: isn't it managed by the subcomponent?

src/app/core/components/content-top-bar/content-top-bar.component.ts
```diff
@@ -28,6 +31,10 @@ export class ContentTopBarComponent {
     delay(0),
   );
 
+  groupInfo$ = this.currentContentService.content$.pipe(
```
smadbe, line 34: This file is generic and should have anything specific to groups. To be checked  after the previous remarks.

src/app/modules/shared-components/components/watch-button/watch-button.component.ts
```diff
@@ -0,0 +1,67 @@
+import { Component, Input, OnChanges, OnDestroy, ViewChild } from '@angular/core';
```
smadbe, line 1: Note to myself: not reviewed yet

smadbe (changes requested):

src/app/modules/shared-components/components/page-navigator/page-navigator.component.html
Iloveall: The problem is - we still need to support start watching user, as our new `watch button` component doesn't support user type yet.

I remember you've wrote that probably we will move content type of user into user info as group info. So then we can remove parent component params of watching feature and leave just `startWatchInfo` for start watch button component.

src/app/modules/group/components/group-header/group-header.component.html
Iloveall: Fixed

src/app/modules/shared-components/components/page-navigator/page-navigator.component.html
Iloveall: Fixed

src/app/core/components/content-top-bar/content-top-bar.component.ts
Iloveall: Fixed

src/app/core/components/content-top-bar/content-top-bar.component.ts
smadbe: You cannot just have a variable called `groupInfo$` is a generic component.. what does it mean when the content is an item?
Actually we do the same just above for items. Just use `currentContent.type === 'group'` to check that it is a group and access the detail info with `$any(...)`.

src/app/modules/shared-components/components/page-navigator/page-navigator.component.html
```diff
@@ -1,6 +1,12 @@
 <div class="page-nav">
+  <alg-watch-button
+    [route]="route"
+    [startWatchGroupInfo]="startWatchGroupInfo"
+    *ngIf="allowWatching && startWatchGroupInfo"
```
smadbe, line 5: Actually, make the watch button follow the current content, it will cleaner for 2 generic places where they are used.

src/app/modules/shared-components/components/page-navigator/page-navigator.component.ts
smadbe: So they will disappear with my previous remark.

src/app/core/components/content-top-bar/content-top-bar.component.ts
smadbe: So they will disappear with my previous remark.

smadbe (changes requested):

src/app/modules/shared-components/components/page-navigator/page-navigator.component.html
Iloveall: Fixed

src/app/modules/shared-components/components/watch-button/watch-button.component.ts
```diff
@@ -1,55 +1,47 @@
...
 
-  @Input() route?: GroupRoute;
-  @Input() startWatchGroupInfo?: StartWatchGroupInfo;
-
-  startWatchGroupInfo$ = new ReplaySubject<StartWatchGroupInfo>(1);
+  groupInfo$ = this.currentContentService.content$.pipe(
+    filter(isGroupInfo),
+  );
   isCurrentGroupWatched$ = combineLatest([
     this.groupWatchingService.watchedGroup$,
```
smadbe, line 21: If used on a page which could change content then when changing from a group to an item, as nothing would be emitted, the watch button related to the previous content would stay :-/
Also the two observables are not independent and so it forces you to do 2 async in cascade in the template. 

It would be cleaner to have something like.
```ts
currentGroupInfo$ = combineLatest([
  this.currentContentService.content$,
  this.groupWatchingService.watchedGroup$
]).pipe(
  map(([ currentContent, watchedGroup ) => (isGroupInfo(currentContent) ? { ...currentContent, isBeingWatched: currentContent.route.id === watchedGroup?.route.id } : undefined )
  })
);
```

smadbe (changes requested):

src/app/modules/shared-components/components/watch-button/watch-button.component.ts
Iloveall: Fixed

smadbe (approved):
Should be good 👍 . I assume you have fully retested after the recent changes.

PR #1281, opened 2023-01-25
Fixes text_id in item parameters page

### LINKED ISSUE #1279: `text_id` not displayed in parameters
**Describe the bug**
The `text_id` of items is currently always empty in the "parameters" page (even when the service returns a value).

**To Reproduce**
Just go on any task parameters page, and try to change the text id.

Changes to be done:
- [ ] make sure the `text_id` returned by the service is displayed on the "parameters" page
- [ ] if the text_id entered is an empty string, push "null" as value

Iloveall:
## Description

Fixes #1279

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/text-id-not-displayed-in-parameters/en/activities/by-id/8340190805761209501;path=;attempId=0/parameters)
  3. And I change "TextID" field
  4. Then I click on "Save" button
  5. Then I see saved "TextID" value
  6. Then I clear "TextID" filed
  7. Then I click on "Save" button
  8. Then I see empty "TextID" field

### DIFF
--- a/src/app/modules/item/components/item-edit-wrapper/item-edit-wrapper.component.ts
+++ b/src/app/modules/item/components/item-edit-wrapper/item-edit-wrapper.component.ts
@@ -139,8 +139,8 @@ export class ItemEditWrapperComponent implements OnInit, OnChanges, OnDestroy, P
     const usesApi = formControls.usesApi?.value as boolean;
     if (isNotUndefined(this.initialFormData.usesApi) && usesApi !== this.initialFormData.usesApi) itemFormValues.uses_api = usesApi;
 
-    const textId = formControls.textId?.value as string;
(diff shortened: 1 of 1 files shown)


smadbe (changes requested):
Could you trim it as well so that extra space is not added ?

smadbe (changes requested):
The trim must be done before the `''->null`, otherwise one space `' '` is still pushed as one space instead of null.

smadbe (changes requested):
Screenshot 2023-01-27 at 15 15 28

smadbe (approved):

PR #1297, opened 2023-02-21
Redirect (hardcoded) paths to ids

smadbe:
## Description

Introduce a mechanism for redirecting (hardcoded in config) path to ids. This can be used for migrating website to our platform... their urls "domain.com/something" can be converted to "domain.com/r/something" by the entry point (load balancer) which can then be redirected to any id via the config.

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. When I go to [an existing label](https://dev.algorea.org/branch/redirect-labels-to-id/en/r/home)
  3. And I see I am redirected to the item "4702" as defined in the config

- [ ] Case 2:
  1. Given I am any user
  2. When I go to [an non-existing label](https://dev.algorea.org/branch/redirect-labels-to-id/en/r/notexisting/url)
  3. And I see I get an error page.


Iloveall (approved):
👍

PR #1316, opened 2023-03-03
Fix do not drop path when valid

smadbe:
## Description

Fixes 2 (related) issues:
1. when a temp user navigate to path he hasn't explored yet (if he receives a link not at the root)... (depending on the order of the requests) most of the time the navigation cannot be load in the left menu as parents have not be started (to reproduce: for instance go to [this page](https://dev.algorea.org/en/activities/by-id/1980584647557587953;path=4702,4102;parentAttempId=0) as a temp user (incognito/private mode))
2. when there was a failure because a path has not been started, the path was completely dropped to restart a new one... but this may cause to change path while it is not what we want

## Test cases

- [ ] Case 1:
  1. Given I am a **temp** user on a new incognito/private tab
  2. Then I open network panel 
  3. When I go to [a page not at root](https://dev.algorea.org/branch/fix-do-not-drop-path-when-valid/en/activities/by-id/1980584647557587953;path=4702,4102;parentAttempId=0)
  4. Then I see in the url bar that the path is not dropped and rebuilt
  5. Then I see the left menu is loaded 

- [ ] Case 2: The previous case is still working (invalid numeric path (forbidden))
  1. Given I am a **temp** user on a new incognito/private tab
  3. When I go to [a page with an invalid path](https://dev.algorea.org/branch/fix-do-not-drop-path-when-valid/en/activities/by-id/1980584647557587953;path=123,456;parentAttempId=0)
  4. Then I see in the url bar that the path is dropped and rebuilt to a correct one "4702, 4102"

- [ ] Case 3: The previous case is still working (char path)
  1. Given I am a **temp** user on a new incognito/private tab
  3. When I go to [a page with an invalid path](https://dev.algorea.org/branch/fix-do-not-drop-path-when-valid/en/activities/by-id/1980584647557587953;path=AB,cd;parentAttempId=0)
  4. Then I see in the url bar that the path is dropped and rebuilt to a correct one "4702, 4102"

### DIFF
--- a/cypress/e2e/item.e2e-spec.cy.ts
+++ b/cypress/e2e/item.e2e-spec.cy.ts
@@ -9,8 +9,8 @@ describe('item page', () => {
     const pageLoadUseCases = [
       { label: 'without path', path: undefined },
       { label: 'with correct path', path: motifArt.path },
-      { label: 'with incorrect path', path: [ '111', '999' ] },
-      { label: 'with incorrect path with non-numbers', path: [ 'invalid', 'path' ] },
+      { label: 'with incorrect path', path: [ '111', '999' ] }, // trigger forbidden
+      { label: 'with incorrect path with non-numbers', path: [ 'invalid', 'path' ] }, // trigger 404
     ];
     for (const useCase of pageLoadUseCases) {
       it(`should load ${useCase.label}`, () => {
--- a/cypress/e2e/item.e2e-spec.cy.ts
+++ b/cypress/e2e/item.e2e-spec.cy.ts
@@ -9,7 +9,8 @@ describe('item page', () => {
     const pageLoadUseCases = [
       { label: 'without path', path: undefined },
       { label: 'with correct path', path: motifArt.path },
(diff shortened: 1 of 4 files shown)


smadbe:
@Iloveall could you check what's wrong with the e2e test ? I don't understand the problem.

Iloveall:
> @Iloveall could you check what's wrong with the e2e test ? I don't understand the problem.

Redirect doesn't work as before - https://dev.algorea.org/en/activities/by-id/1625159049301502151;path=invalid,path;parentAttempId=0

smadbe:
> 
> Redirect doesn't work as before - https://dev.algorea.org/en/activities/by-id/1625159049301502151;path=invalid,path;parentAttempId=0
> 
> After: https://dev.algorea.org/branch/fix-do-not-drop-path-when-valid/en/activities/by-id/1625159049301502151;path=invalid,path;parentAttempId=0

I do not see any difference between your two link. 🤷‍♂️

smadbe:
Ok, good. Could you review it @Iloveall ?

Iloveall (approved):

PR #1318, opened 2023-03-06
fix many memory leaks due to many unreleased observables

smadbe:
## Description

Fixes many memory leaks (some are just "securities without being sure", but many are actual leaks) due to unreleased observables.

I have written the recommandations here: https://france-ioi.github.io/algorea-devdoc/ops/frontend/#avoiding-memory-leaks

## Test cases

Testing the app globally, including tasks.


Iloveall (approved):
👍

PR #1334, opened 2023-03-30
Do not wait for answer to gen token

smadbe:
## Description

Do not wait for the answer fetching to start generating the token, so that we save 1-request time when loading a task.

In practice, it means that when it is possible (i.e., when the token generation does not need the answer), set the answer to `undefined` at first to let the task initialize (starting by generating token).

## Test cases

- [ ] Case 1: autoloading the last answr
  1. Given I am the usual user
  3. And I click on ...
  4. Then I see ...
  5. And

- [ ] Case 2: loading an existing answer
  1. Given I am the usual user
  2. When ...

- [ ] Case 3: loading an existing answer from another user
  1. Given I am the usual user
  2. When ...

- [ ] Case 4: loading a best answer
  1. Given I am the usual user
  2. When ...

- [ ] Case 4: loading a best answer from another user
  1. Given I am the usual user
  2. When ...

### DIFF
--- a/src/app/modules/item/pages/item-by-id/item-by-id.component.ts
+++ b/src/app/modules/item/pages/item-by-id/item-by-id.component.ts
@@ -160,7 +160,6 @@ export class ItemByIdComponent implements OnDestroy, BeforeUnloadComponent, Pend
     switchMap(data => {
       if (!isTask(data.item)) return of(null); // config for non-task's is null
       return this.initialAnswerDataSource.answer$.pipe(
-        filter(isNotUndefined), // undefined is for not-known-yet and so have to be skipped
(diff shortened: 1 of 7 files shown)


src/app/modules/item/pages/item-by-id/item-by-id.component.ts
```diff
@@ -226,7 +225,6 @@ export class ItemByIdComponent implements OnDestroy, BeforeUnloadComponent, Pend
 
     this.itemRouteState$.subscribe(state => {
       if (state.isReady) {
-        // configure initialAnswerDataSource
```
smadbe, line 229: this line is just a cleaning, not directly related with this PR

Iloveall (approved):

PR #1348, opened 2023-04-27
New design/content header

Iloveall:
## Description

Fixes #...

Changes design for item header, fixes e2e

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/new-design/content-header/en/a/1135065237661691731;p=4702,1471479157476024035;a=0)
  3. Then I see the new title without score ring or skill progress, new neighbour buttons, new chat button

### DIFF
--- a/src/assets/scss/_text-sizes.scss
+++ b/src/assets/scss/_text-sizes.scss
@@ -1,0 +1,19 @@
+.alg-h1 {
+  font-size: 3rem;
+}
+
+.alg-h2 {
+  font-size: 2.5rem;
+}
+
+.alg-h3 {
+  font-size: 2rem;
+}
+
+.alg-h4 {
+  font-size: 1.5rem;
+}
+
+.alg-h5 {
+  font-size: 1.25rem;
+}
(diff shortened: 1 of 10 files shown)


smadbe (approved):

src/assets/scss/_text-sizes.scss
```diff
@@ -0,0 +1,19 @@
...
+  font-size: 2rem;
+}
+
+.alg-h4 {
+  font-size: 1.5rem;
+}
+
+.alg-h5 {
+  font-size: 1.25rem;
+}
```
smadbe, line 19: > >  `<p class="item-title alg-h1" data-cy="item-title">{{ itemData.item.string.title }}</p>` -> why not using a h1 directly instead of p with a h1 tag? (would be more "semantic")
>
>  Could be, I’ll change

Iloveall: Fixed #1362

PR #1351, opened 2023-04-28
New design/item description

Iloveall:
## Description

Fixes #...

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/new-design/item-description/en/a/39530140456452546;p=;a=0)
  3. Then I see a new description design

### DIFF
--- a/src/assets/scss/_text-sizes.scss
+++ b/src/assets/scss/_text-sizes.scss
@@ -17,3 +17,26 @@
 .alg-h5 {
   font-size: 1.25rem;
 }
+
+.alg-text-body-1 {
+  color: var(--alg-secondary-color);
+  font-size: 1.25rem;
+  line-height: 1.2;
+  font-weight: 400;
+}
+
+.alg-text-body-2 {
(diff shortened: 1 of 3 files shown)


src/assets/scss/_text-sizes.scss
```diff
@@ -17,3 +17,26 @@
...
 }
+
+.alg-text-body-1 {
+  color: var(--alg-secondary-color);
+  font-size: 1.25rem;
+  line-height: 1.2;
+  font-weight: 400;
+}
+
+.alg-text-body-2 {
```
smadbe, line 28: could we try to find a more appropriate naming than `alg-text-body-1` `alg-text-body-2` ? Does not seem practical when using it.

smadbe (approved):

src/assets/scss/_text-sizes.scss
Iloveall: Fixed

PR #1366, opened 2023-05-09
New design/mobile

Iloveall:
## Description

Fixes #...

Improves mobile supporting 

## Test cases

Branch for [test](https://dev.algorea.org/branch/new-design/mobile/en/a/home;pa=0)

### DIFF
--- a/src/app/modules/item/pages/item-by-id/item-by-id.component.ts
+++ b/src/app/modules/item/pages/item-by-id/item-by-id.component.ts
@@ -13,6 +13,7 @@ import {
   catchError,
   mergeWith,
   takeUntil,
+  withLatestFrom,
 } from 'rxjs/operators';
 import { defaultAttemptId } from 'src/app/shared/helpers/attempts';
 import { errorState, fetchingState, FetchState, isFetchingOrError, readyState } from 'src/app/shared/helpers/state';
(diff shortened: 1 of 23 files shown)


smadbe:
In this PR or another: On mobile, when "opening" the menu, the (right) content first switch to the "full frame mode" so displaying the title in big etc... but which is never really shown on mobile... so the change does not look great.

https://github.com/France-ioi/AlgoreaFrontend/assets/1053150/35901146-a41c-4834-9160-506fdd3bffe6

src/app/modules/item/pages/item-by-id/item-by-id.component.ts
```diff
@@ -312,9 +313,10 @@ export class ItemByIdComponent implements OnDestroy, BeforeUnloadComponent, Pend
       filter(([ previous, current ]) => (!!current && (!previous || isTask(previous.item) || isTask(current.item)))),
       map(([ , current ]) => ensureDefined(current).item),
     ).subscribe(item => {
+      const isInitFromMobile = !this.layoutService.fullFrameValue.animated && this.breakpointObserver.isMatched('(max-width: 479px)');
```
smadbe, line 316: probably that logic should be moved the the layout service. This value of `479` is very layout specific and shouldn't be repeated at many places in the code. (you cannot have a single constant that you use both the css and ts?)

smadbe, line 316: I see that fullFrame was initially a observable... so instead of adding a function `fullFrameValue`, you should `withLatestFrom(this.layoutService.fullFrame$)`.

smadbe (changes requested):

src/app/modules/item/pages/item-by-id/item-by-id.component.ts
Iloveall: The value `479` - fixed for ts. I've took standard (Handset: (max-width: 599.98px)) from https://material.angular.io/cdk/layout/overview#predefined-breakpoints and also changed it for css. But I can't fix quickly variable for css as `var()` won't work here.

Iloveall: Fixed

Iloveall: Layout better to do separated as after move it need to test a lot of cases

smadbe (changes requested):
This remark is still to be fixed: https://github.com/France-ioi/AlgoreaFrontend/pull/1366#issuecomment-1540380333

smadbe:
Actually, there are more problem with this title with or without mobile. I will take over this PR for the moment.

smadbe:
Will not merge. Superseeded by #1375 which includes these commits.

PR #1375, opened 2023-05-26
New design: new behavior for narrow and header

smadbe:
## Description

Change the behavior related to the visibility of the left menu and to the header.

### Left menu

1) when landing on the app: 
  - when the screen is narrow (<600px), the menu should not be shown
  - when in LTI, the menu should not be shown
  - in the other cases, it should be shown

2) when in LTI, it should not be possible to show the menu

3) otherwise, the button to manually show/hide the menu should work (both on small and not-small screen)

4) when we navigate to a task (not using the left menu) that has to be shown full screen (i.e. a blockly task), the menu is hidden after its loading 

5) when using the left menu to navigate to a content (even blockly tasks), the menu is not hidden

### Header

1) when on a task that has to be shown full screen (i.e. a blockly task), the header is not shown

2) in all the other cases, the header is shown... the top bar still switch to the title only (instead of breadcrumb) after scroll

## To be checked / to be done

- [ ] ~when reducing the window size while the menu is not shown, when we reach 600px, there is an animation to "rehide" the already-hidden menu -> should not happen~
- [ ] in the content-top-bar : move the "open menu" button so that it may be displayed in front of the breadcrumb as well (so it has to be always shown if `showLeftMenuOpener`  even in `*ngIf="fullFrameContentDisplayed || scrolled"` condition does not match)
- [x]  I've removed the condition in `src/app/modules/item/components/chapter-children/chapter-children.component.html` line 21 because I didn't understand why it was there. check that.
- [x] in `src/app/core/app.component.scss`, I've removed the `mobile` as I had no condition anymore to check that... and switch to the usual way to do it with ` @media screen and (max-width: 599.98px)` , but maybe that's wrong?

## Test cases

The cases described above.

### DIFF
--- a/src/app/shared/services/layout.service.ts
+++ b/src/app/shared/services/layout.service.ts
@@ -45,7 +45,7 @@ export class LayoutService implements OnDestroy {
     scan((prev, [ isNarrowScreen, canShowLeftMenu, isFullFrameContent, manualMenuToggle ], idx) => {
       if (!canShowLeftMenu) return { shown: false, animated: false, isNarrowScreen };
       if (idx === 0) return { shown: !isNarrowScreen, animated: false, isNarrowScreen };
-      if (!prev.isNarrowScreen && isNarrowScreen) return { shown: false, animated: true, isNarrowScreen };
+      if (prev.isNarrowScreen !== isNarrowScreen) return { shown: !isNarrowScreen && !isFullFrameContent, animated: true, isNarrowScreen };
       if (manualMenuToggle !== undefined) return { shown: manualMenuToggle, animated: true, isNarrowScreen };
       if (isFullFrameContent) return { shown: false, animated: true, isNarrowScreen };
       return { ...prev, isNarrowScreen };


src/app/modules/item/components/chapter-children/chapter-children.component.html
```diff
@@ -18,8 +18,7 @@
 
     <ng-container *ngIf="itemData && itemData.item.childrenLayout === 'Grid'; else listSection">
       <div
-        class="grid-container"
-        [ngClass]="{ 'full-frame': (fullFrame$ | async)?.active }"
+        class="grid-container full-frame"
```
Iloveall, line 21: I was done for more flexible setting the size of cards for full and non full mode -> https://github.com/France-ioi/AlgoreaFrontend/commit/534ec5f5b036992f6dc8bda31519a9ac0a38fdb4#diff-15882a5899129eb357146a6893c239a703799d352627ae4a10b54390dcc2e0f8R74

src/app/core/app.component.scss
```diff
@@ -73,7 +73,16 @@
     padding-top: 7.5rem;
   }
 
-  &.collapsed, &.mobile {
```
Iloveall, line 76: So don't wee need to hide title on mobile? https://github.com/France-ioi/AlgoreaFrontend/pull/1366#issuecomment-1540380333

Iloveall:
How to get open the menu button when full frame mode? https://dev.algorea.org/branch/new-design/narrow-and-title/en/groups/by-id/2713577096475953687;p=

smadbe:
> How to get open the menu button when full frame mode? https://dev.algorea.org/branch/new-design/narrow-and-title/en/groups/by-id/2713577096475953687;p=

That's related with the 2nd todo: for now, you need to scroll. But that needs to be fixed.

src/app/core/app.component.scss
smadbe: Hum I need to test a bit.

src/app/modules/item/components/chapter-children/chapter-children.component.html
smadbe: Would you have screenshots? When I tested, I didn't see something in particular.

Iloveall: As usual

1. Go to this [page](https://dev.algorea.org/branch/bugfix/prevent-scroll-bar-jump-forced-scroll/en/a/899800937596940136;p=;a=0)
2. And I change width of browser
3. Then I see specific settings for size of cards which depends of size of non full frame mode
4. Then I activate full frame mode
5. Then I see another specific settings for size of cards which depends of another size of full frame mode

src/app/core/app.component.scss
smadbe: From current "Sketch"es , no.

Iloveall (approved):

PR #1439, opened 2023-07-25
Adds style improvements

### LINKED ISSUE #1435: More minor style improvements
Set of small style improvement suggestions coming from the designer:
- add `{ line-height: 1em; margin-bottom : .5rem }` to titles (item title for instance)
- add ` { font-size: 1.25rem; line-height: 1em }` to item subtitles
- add a `max-width: 920px` to the item descriptions
- reduce the spacing between title-subtitle and the content: (should work with/without subtitle, with/without description, with/without tabs) 
  - there is a margin at the bottom of the subtitle (only when present?), probably it should removed?
  - probably no margin is needed on top of the content
  - probably no margin needed on top of the description
  - may require (not sure) a small margin below the tab bar

(don't hesitate to create some content in the [Algorea Testing Content for devs](https://dev.algorea.org/en/a/7528142386663912287;p=;a=0) (a new "Basic platform layouts" chapter for intance where you test  with/without subtitle, with/without description, grid view), and you can use locally `featureFlags: { hideTaskTabs: [ 'progress' ] } `  in your config to hide tabs.

- use opacity 50% on the "Search ?" placeholder
- search field: have its background white when on focus
- search warning message:
  - the height is fixed size so the padding is a bit small when on several lines (for instance " Type at least 3 letters to search content ")
  - padding of the container may be `.719rem .75rem`
- use a text color of `RGBA(0, 0, 0, .5)` (new color to be defined) to be used on all colored backgrounds (such as yellow warning msgs)

Iloveall:
## Description

Fixes #1435

...

## Test cases

Chapter with [test cases](https://dev.algorea.org/branch/feature/style-improvements/en/a/6621899821435600308;p=7528142386663912287;a=0)

### DIFF
--- a/src/app/core/components/left-menu-search/left-menu-search.component.html
+++ b/src/app/core/components/left-menu-search/left-menu-search.component.html
@@ -1,5 +1,5 @@
 <alg-input
-  class="transparent"
+  class="transparent search-placeholder"
   inputIcon="ph-duotone ph-magnifying-glass"
   i18n-placeholder placeholder="Search ?"
   name="search"
--- a/src/app/core/components/left-search-result/left-search-result.component.html
+++ b/src/app/core/components/left-search-result/left-search-result.component.html
@@ -14,13 +14,13 @@
 </div>
 
 <div class="container" *ngIf="data === undefined">
-  <alg-message-info icon="ph-duotone ph-warning-circle" i18n>
+  <alg-message-info class="search" icon="ph-duotone ph-warning-circle" i18n>
     Type at least 3 letters to search content
   </alg-message-info>
 </div>
 
 <div class="container" *ngIf="data !== undefined && data.length === 0">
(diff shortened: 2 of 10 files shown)


smadbe (changes requested):
I think there are still too much spacing between the title and the description. In the example, it is a bit biased by the fact there is the tab tab, but you have to imagine without tab bar (as it is done on https://beta.opentezos.com/en/a/home;pa=0 on most items when you are not logged-in). If I am correct, the result would be the following, which has too much useless spacing:
Screenshot 2023-07-27 at 11 59 35

The issue had "probably no margin needed on top of the description", I seems to still be there. (they might require a (small) spacing under the tab bar?)

The search warning has the same height when the message is on one and two lines... I would expect it to be smaller when there is one line. Is there a reason ? 
Screenshot 2023-07-27 at 12 04 38
Screenshot 2023-07-27 at 12 04 50

smadbe (approved):

PR #1469, opened 2023-09-12
Quick-fix: make sure the breadcrumb is not shown in the top when using LTI

smadbe:
## Description

We got the complain that, when using LTI, users were seeing the breadcrumb while they were not supposed to. Fix that.

## Test cases

- [x] Case 1: BEFORE 
  2. When I go to [the moodle test platform](http://moodletest.algorea.org/course/view.php?id=3#section-1)
  3. And I click on "Test dev.algorea.org (new platform)"
  4. Then I see that the breadcrumbs are shown during the loading

- [x] Case 2: FIX 
  2. When I go to [the moodle test platform](http://moodletest.algorea.org/course/view.php?id=3#section-1)
  3. And I click on "Test dev.algorea.org (new platform) (quick-fix-disable-breadcrumbs-on-lti)"
  4. Then I see that the breadcrumbs are now shown during the loading


collettemathieu (approved):

collettemathieu:
No particular comment. Good job.

PR #1477, opened 2023-09-14
Fix web manifest file distribution

smadbe:
## Description

Currently, the asset path was not put in front of the reference to the web manifest file. Fix this as well as the image reference it contains.

## Test cases

- [ ] Case 1:
  1. Given I use Chrome
  2. I open network debug panel and disable cache
  3. When I go to [the home page](https://dev.algorea.org/branch/fix-webmanifest-distribution/en)
  4. And I see the webmanifest file is received
  5. Then I see in the response that the image it references are reachable as well

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -39,6 +39,11 @@
                 "glob": "**/*",
                 "input": "src/assets/images/",
                 "output": "/images/"
+              },
+              {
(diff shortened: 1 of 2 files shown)


Iloveall (approved):
👍

PR #1490, opened 2023-09-25
Feature/display status and status changing button

Iloveall:
## Description

Fixes #1482

## Notes (out of scope, known isues, hints for reviewing code, ...)  (optional)

...

## Test cases

- [ ] Case 1:
  1. Given I am the TEMP user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/display-status-and-status-changing-button/en/a/1357671178979168705;p=4702,1625159049301502151,5207993233955027100;pa=0)
  3. And I click on thread button
  4. Then I see the thread with caption "This thread is closed" and button is disabled with caption "Open this thread"
  5. Then I hover on button
  6. Then I see the caption: "You are not allowed to open this thread."

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/display-status-and-status-changing-button/en/a/1293400563985240032;p=7528142386663912287,7523720120450464843;a=0?watchedGroupId=117532200658920233&watchUser=0)
  3. And I click on thread button
  4. Then I see the thread with caption "This thread is closed" and button is enabled with caption "Open this thread"
  5. Then I click on button
  6. Then I see the error notification "Not implemented"
  7. Then I turn on "Offline" mode in Network tab
  8. And close open the thread
  9. Then I see "Error while loading thread info" as status message and no button

- [ ] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/display-status-and-status-changing-button/en/a/6379723280369399253;p=4702,1625159049301502151,5207993233955027100;a=0)
  3. And I click on thread button
  4. Then I see the thread with caption "This thread is open" and button is enabled with caption "Close this thread"
  5. Then I click on button
  6. Then I see the error notification "Not implemented"

- [ ] Case 4:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/display-status-and-status-changing-button/en/a/6379723280369399253;p=7528142386663912287,7523720120450464843;a=0?watchedGroupId=752024252804317630&watchUser=1)
  3. And I click on thread button
  4. Then I see the thread with caption "This thread is open" and button is disabled with caption "Close this thread"
  5. Then I hover on button
  6. Then I see the caption: "Only the user of the thread can close it."

- [ ] Case 5:
  1. Given I am the usual user
  2. And I open "Network" panel
[...]

PR #1505, opened 2023-10-02
Adds item title in thread header

### LINKED ISSUE #1500: Forum panel: display item title in header
## Overview
Add the item title in the forum panel header.
Screenshot 2023-09-29 at 15 49 46

## Acceptance criteria
- [ ] replace "Forum" by the discussion icon (used on the top bar) followed by the title of the item.
- [ ] for now, just fetch the item information related with the current `this.threadService.threadId$`  in the thread component 
- [ ] the item title should be clickable
- [ ] the exisiting `return this.getItemByIdService.get(threadId.itemId, isCurrentUserParticipant ? undefined : threadId.participantId).pipe(` line can reuse the fetched item if in the `isCurrentUserParticipant` case.

Iloveall:
## Description

Fixes #1500

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. And I open "Network" panel
  3. When I go to [this page](https://dev.algorea.org/branch/feature/forum-panel-display-item-title-in-header/en/a/home;pa=0/forum/others)
  4. And I click on "Show" button of thread list
  5. Then I see the Item Title in header of thread section
  6. Then I click on next "Show" button
  7. Then I see the Item Title has changed
  8. Then I click on "Hide" the thread and click on "Show" same thread
  9. And I see no request for item in "Network" panel
  10. And I click on next "Show" button
  11. Then I see no cancelled or duplicated calls in "Network" panel

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-panel-display-item-title-in-header/en/a/home;pa=0/forum/others)
  3.  And I click on "Show" button of "Blockly Basic Task" thread
  5. Then I see the "Blockly Basic Task" in header of thread section
  6. And I click on "Blockly Basic Task" title
  7. Then I see "Blockly Basic Task" page is opened

### DIFF
--- a/src/app/core/components/thread-container/thread-container.component.ts
+++ b/src/app/core/components/thread-container/thread-container.component.ts
@@ -1,7 +1,11 @@
 import { AfterViewInit, Component, Input, OnDestroy, ViewChild } from '@angular/core';
 import { DiscussionService } from 'src/app/modules/item/services/discussion.service';
-import { filter, Subscription } from 'rxjs';
+import { combineLatest, distinctUntilChanged, filter, Subscription, switchMap } from 'rxjs';
 import { ThreadComponent } from '../thread/thread.component';
+import { ThreadService } from '../../../modules/item/services/threads.service';
+import { UserSessionService } from '../../../shared/services/user-session.service';
+import { GetItemByIdService } from '../../../modules/item/http-services/get-item-by-id.service';
+import { isNotNull } from '../../../shared/helpers/null-undefined-predicates';
 
 @Component({
   selector: 'alg-thread-container',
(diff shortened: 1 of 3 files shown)


smadbe (changes requested):
As I wrote the issue, you can reuse the same request in the case where you do not use "watched_group_id" param, but not when it is given. We use the forbidden case to detect that the current user cannot watch the user. So you may get a forbidden instead of getting your title.
 Cfr scenario 7 in https://github.com/France-ioi/AlgoreaFrontend/pull/1490

In your case, as you do not even handle the error case, you get a crash. 


From what I see, you do not re-use the request done to check, for closed threads, that it can open it. So if you don't reuse, you just don't need the "watched_group_id" parameter to get the title...

BTW, your scenarios must also include a closed thread, as it is the case with the duplicated request.

src/app/core/components/thread-container/thread-container.component.ts
```diff
@@ -14,10 +19,26 @@ export class ThreadContainerComponent implements AfterViewInit, OnDestroy {
...
+    ),
+  );
 
   private subscription?: Subscription;
 
-  constructor(private discussionService: DiscussionService) {
+  constructor(
+    private discussionService: DiscussionService,
+    private threadService: ThreadService,
+    private userSessionService: UserSessionService,
```
smadbe, line 39: is it still used?

smadbe (approved):

src/app/core/components/thread-container/thread-container.component.ts
Iloveall: Fixed

PR #1510, opened 2023-10-04
migration to standalone components

smadbe:
Convert the whole app to standalone components. 
Manually get rid of every remaining modules.


Iloveall (approved):

Iloveall (changes requested):
There is issue with guard for forms

smadbe:
> There is issue with guard for forms
> 
> CASE:
> 
>     1. As usual
> 
>     2. Go to this [branch](https://dev.algorea.org/branch/migrate-to-standalone-comp/en/a/902356740789159624;p=;a=0/parameters)
> 
>     3. And I do some changes
> 
>     4. Then I click on "Content" tab
> 
>     5. Then I see confirmation modal and page is changed

I saw it... but the bug was already there before! ;-)

Iloveall (approved):

PR #1512, opened 2023-10-05
Feature/improve x ago dates

### LINKED ISSUE #1383: Improved "x ago" dates
## Overview
New "x ago" pipe for displaying dates on pages like item history. (like in github)

Can get inspired from: 
- angular pipe doing something very similar: https://github.com/tuhinkarmakar/ng-relative-time/blob/master/projects/relative-time/src/lib/pipes/relative-time.pipe.ts (I prefer reusing the code than important that as a full module for just this pipe, but just don't forget the license)
- github implem: https://github.com/github/relative-time-element
It may use the browser support (not a big deal if not well supported (just the date displayed) on the browser which do not support `Intl.RelativeTimeFormat` 

## Acceptance criteria
- [ ] Display <1min as "Just now"
- [ ] Display "yesterday" if yesterday
- [ ] Display "x minute(s)/hour(s)/day(s)/month(s)/year(s) ago" depending on the case
 - [ ] Make sure the days, years, ... get no "s" if no plural. ⚠️ 0 is plural in English and not plural in French so i18n should be handle correctly
- [ ] Should use the current locale of the user
- [ ] No need to implement the "future" for now
- [ ] There should be tests to be able to test every case (for the pipe)
- [ ] On hover, the full date (including the tz?) should be displayed (as in github)

Iloveall:
## Description

Fixes #1383

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/improve-x-ago-dates/en/a/6379723280369399253;p=7523720120450464843;a=0/progress/history)
  3. Then I see relative dates in table

Tested locally:
Screenshot at Oct 05 16-37-51

Other test cases tested in spec file


src/app/shared/pipes/relativeTime.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (absDiffInMS < MINUTES) {
+    return $localize`Just now`;
+  }
+
+  if (absDiffInMS > MINUTES && absDiffInMS < HOURS) {
+    return formatter.format(Math.trunc(diffInMS / MINUTES), 'minutes');
+  }
+
+  if (absDiffInMS > HOURS && absDiffInMS < DAYS) {
+    return formatter.format(Math.trunc(diffInMS / HOURS), 'hours');
```
smadbe, line 27: is it internationalized?

Iloveall: The `Intl.RelativeTimeFormat` object enables language-sensitive relative time formatting - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat

I've seen fr lang but I can't check if it's correct. But I assume as it part of js - must support the cases you described (plural)

smadbe (approved):

smadbe (changes requested):
But maybe you can use it somewhere to check that the behavior is as expected. In a history list maybe?

smadbe:
It will be generated once for all while the date does not change? It is not really a pure pipe actually as time runs... but we cannot recompute them all the time...

Iloveall:
> But maybe you can use it somewhere to check that the behavior is as expected. In a history-tab list maybe?

I've add test case

Iloveall:
> It will be generated once for all as long as the date does not change? It is not really a pure pipe actually as time runs... but we cannot recompute them all the time...

In libs kind this they keep both pipes, pure and non pure -> https://github.com/joanllenas/ngx-date-fns#pure-or-impure
Should we do same?

smadbe:
> > It will be generated once for all as long as the date does not change? It is not really a pure pipe actually as time runs... but we cannot recompute them all the time...
> 
> In libs kind this they keep both pipes, pure and non pure -> https://github.com/joanllenas/ngx-date-fns#pure-or-impure Should we do same?

They mention locale change... ok we can consider it will not change... but I was mentioning time change "just now" is not valid anymore after 2 hours.

smadbe (approved):
👍

PR #1520, opened 2023-10-11
Project file reorg

smadbe:
## Description

Reorganize the project structure (i.e. move files around) as discussed.

Just having the former modules (`items`, `groups`, `lti`) as features for now.

There are many "things" which are not at their ideal place, because there are doing several things, because there should not be shared as they are, ... I consider we'll improve this later.


Iloveall:
2 things about folders naming which not reflected for me what it does until opened it:
- containers - is it means layout part or just renamed components?
- data-access - data-access-services? But mb data-access more compact

Also I would suggest to start to describe folder architecture in readme so it would be helpful for new users devs and much easy if we start to do it now

Iloveall (approved):

PR #1534, opened 2023-10-23
Changes approach for open and close right thread panel

### LINKED ISSUE #1533: Fix the forum layout
**Describe the bug**
In some cases, we may get the forum displayed while it was not supposed to:
Screenshot 2023-10-20 at 14 44 35

As discussed on slack:

> I assume it’s because of flex approach. But even if something you have changed it’s should not broke layout. So definitely the layout must be improved
> Mb position absolute - more correct approach for avoid glitch like this.

Iloveall:
## Description

Fixes #1533

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/change-thread-panel-open-close-logic/en/a/6379723280369399253;p=7523720120450464843;a=0)
  3. And I click on "Thread" button in top bar
  4. Then I see the panel of thread is opened correct

- [ ] Case 2:
  1. Given I am the usual user
  2. From narrow size (mobile)
  3. When I go to [this page](https://dev.algorea.org/branch/feature/change-thread-panel-open-close-logic/en/a/6379723280369399253;p=7523720120450464843;a=0)
  4. And I click on "Thread" button in top bar
  5. Then I see the panel of thread is opened correct

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -3,7 +3,7 @@
     <alg-left-menu></alg-left-menu>
   </div>
   <ng-container *ngrxLet="isWatching$ as isWatching">
-    <div class="right-container" [ngClass]="{ collapsed: !leftMenu.shown, animated: leftMenu.animated, 'thread-opened': (isDiscussionVisible$ | async) }">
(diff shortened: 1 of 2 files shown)


smadbe (approved):
I cannot really say if it is better, but not worse at least.

PR #1552, opened 2023-11-17
Adds relative time format into forum messages

### LINKED ISSUE #1548: Forum: use relative time
## Overview

Use relative time in the forum

## Acceptance criteria
- [ ] Use relative time for time which is displayed in the forum thread
- [ ] Make sure it is updated every minute (?) so that the indication are quite correct
- [ ] Make sure it is not updated while the thread is not displayed

Iloveall:
## Description

Fixes #1548

For my point as non pure pipe - it makes actual time without any additional timers very well.

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/forum-use-relative-time/en/a/6379723280369399253;p=7523720120450464843;a=0)
  3. And I click on open the thread button
  4. Then I type message and click on send button
  5. Then I see new message with "Just now" caption
  6. Then I wait ~1 min
  7. Then I see the caption: "1 minute ago"

### DIFF
--- a/src/app/forum/containers/thread-message/thread-message.component.html
+++ b/src/app/forum/containers/thread-message/thread-message.component.html
@@ -67,5 +67,5 @@
 </div>
 
 <ng-template #sentAt>
-  <span i18n>{{ event.time | date : 'short' }}</span>
+  <span i18n>{{ event.time | relativeTime }}</span>
 </ng-template>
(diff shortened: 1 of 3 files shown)


smadbe (approved):
The problem with impure pipes is that it may impact performances. When we have a table with many dates, don't you think it may have a noticeable impact?  
Recomputing it as any mouse mouvement, isn't it very "expensive" for something which change very rarely ?

PR #1642, opened 2024-02-09
Replace io-ts by zod

smadbe:
## Description

Introduce "Zod" to replace io-ts+fp-ts. 
Lighter and more readable (no need for the functional part, the pipe and intersections).

I've tried to make it using automatic conversion tool and testing the new decoder with the former type before deleting it.. but I cannot promise it has no error.

## Test cases

As the usual user, I go to:
- [an activity log](http://localhost:4200/groups/by-id/2713577096475953687;p=)
- [an activity page](http://localhost:4200/a/home;a=0)
- [a group composition page](http://localhost:4200/groups/by-id/947670356997220543;p=/members) -> and the "sub-groups" tab
- [a progress matrix](http://localhost:4200/a/6707691810849260111;p=;pa=0/progress/chapter?watchedGroupId=672913018859223173)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
         "primeng": "^17.1.0",
         "rxjs": "~7.8.1",
         "tslib": "^2.3.1",
+        "zod": "^3.22.4",
         "zone.js": "~0.14.2"
       },
       "devDependencies": {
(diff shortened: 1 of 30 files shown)


Iloveall (approved):
Nice lib, I've used it once but not so deep 👍

PR #1648, opened 2024-02-15
Changes version of node_browsers_image

Iloveall:
## Description

Fixes circleci process

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
   # deps has to be run before the other jobs and basically run "npm install" and cache it
   deps:
     docker: &node_image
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ jobs:
   # deps has to be run before the other jobs and basically run "npm install" and cache it
   deps:
     docker: &node_image
-      - image: cimg/node:18.13
+      - image: cimg/node:21.6.1
```
smadbe, line 64: It is not only "node_browsers_image". 
Shouldn't we stick with the 20.11 "LTS" version for now?
It has impact on my other building / deployment tools as well.

PR #1660, opened 2024-02-26
Use playwright instead of Cypress

smadbe:
I wanted to give Playwright a try as I saw it recommended at many places for UI testing. Also because we do not use E2E in practice (we do not write new tests) while it would be very useful.

So my observation at the end: 
- Test writing using code gen are more straightforward and that is quite easy to write a new test
- the http interception to simulate failure is very useful, especially as this is typically the kind of cases that are even hard to test once for the review

What I did:
- rewrite roughly the same few tests written in cypress to playwright
- change the CI/package conf
- add some brief explanation in the readme
- showcase some important aspects of playwright:
  - simulating a http failure through http interception
  - soft assertions
- I upload the e2e report at the end
- (when there is a PR) it automatically posts a comment on the PR with a link to the report

It would be great if it could push us toward writing e2e tests with each PR in the same way we write scenarios (it could even replace scenarios).


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/playwright/playwright-report/index.html

Iloveall (approved):
Nice 👌 I like it

PR #1696, opened 2024-04-16
Feature/let managers edit approval setting of group part 1

### LINKED ISSUE #1693: Let managers edit approval setting of a group (part 1)
## Overview
Approvals required by group cannot currently be configured. So, create an place where it can be configured. 

## Acceptance criteria
- [ ] in the "settings" tab of a group, for managers, add a "Required approvals" section after "Group Information".
- [ ] in this section, add an explanation: "If you enable some of the rules below, users joining the group will have to approve them first."
- [ ] then add "Lock membership until a given date" with a on/off switch. If "on", a date-time field appears (can be textual if no widget already exists for that)
- [ ] then add "Managers can access member's personal information with a 3 options selector: "No"/"Read only"/"Read and edit"

This will only work for groups without members... the case with members will be handled in the "part 2" issue (#1694).

Iloveall:
## Description

Fixes #1693

I've add e2e tests and manual scenarios until build issue will be fixed.

The e2e tests for this feature also related on BE data so impossible to check in parallel on all browsers, so need to check in queue.

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/let-managers-edit-approval-setting-of-group-part-1/en/groups/by-id/8791610698648252339;p=/settings)
  3. Then I see "Required approvals" section
  4. Then I turn on "Lock membership until a given date" control
  5. Then I see input date control
  6. Then I put the date
  7. Then I select "Read only" button
  8. And I click on "Save" button
  9. The I see success notification

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/let-managers-edit-approval-setting-of-group-part-1/en/groups/by-id/8791610698648252339;p=/settings)
  3. Then I see "Required approvals" section
  4. Then I turn off "Lock membership until a given date" control
  5. Then I see datepicker control disappeared
  7. Then I select "No" button
  8. And I click on "Save" button
  9. The I see success notification

- [ ] Case 3:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/feature/let-managers-edit-approval-setting-of-group-part-1/en/groups/by-id/8791610698648252339;p=/settings)
  3. Then I see "Required approvals" section
  4. Then I turn on "Lock membership until a given date" control
  5. Then I see input date control
  6. Then I put past date in control
  7. Then I see validation message "Minimum date: ..."
  8. Then I put invalid date: "32/12/2024 12:00"
  9. Then I see "Invalid date" validation
  10. Then I put invalid time "12/12/2024 60:00"
  11. Then I see "Invalid date" validation
  12. Then I clear input date control and blur focus
  13. Then I see the validation message has disappeared

### DIFF
--- a/e2e/groups/group-edit-requires-approval.spec.ts
+++ b/e2e/groups/group-edit-requires-approval.spec.ts
@@ -5,10 +5,6 @@ import { DAYS } from 'src/app/utils/duration';
 
 test.describe.configure({ mode: 'serial' });
 
-test.use({
-  timezoneId: 'Africa/Accra',
-});
-
 test('Turn on approval rules', async ({ page }) => {
   await initAsUsualUser(page);
   await page.goto('/groups/by-id/6688692310502473984;p=/settings');
--- a/e2e/groups/group-edit-requires-approval.spec.ts
+++ b/e2e/groups/group-edit-requires-approval.spec.ts
@@ -86,9 +86,7 @@ test('Check invalid date validation', async ({ page }) => {
   await test.step('Fill past date', async () => {
     const currentDate = new Date();
     await inputDateLocator.fill(convertDateToString(currentDate));
-    console.log('currentDate', currentDate);
-    console.log('currentDate', convertDateToString(currentDate));
-    await expect.soft(page.getByText(`Minimum date: ${ convertDateToString(currentDate) }`)).toBeVisible();
+    await expect.soft(page.getByText(`The date must be grater than: ${ convertDateToString(currentDate) }`)).toBeVisible();
   });
 
   await test.step('Clear invalid date and check validation message is gone', async () => {
--- a/e2e/groups/group-edit-requires-approval.spec.ts
+++ b/e2e/groups/group-edit-requires-approval.spec.ts
@@ -1,34 +1,104 @@
 import { test, expect } from '@playwright/test';
 import { initAsUsualUser } from '../helpers/e2e_auth';
+import { convertDateToString } from 'src/app/utils/input-date';
+import { DAYS } from 'src/app/utils/duration';
 
 test.describe.configure({ mode: 'serial' });
 
-test('Setting group turn on approval rules', async ({ page }) => {
+test.use({
+  timezoneId: 'Africa/Accra',
+});
+
+test('Turn on approval rules', async ({ page }) => {
   await initAsUsualUser(page);
   await page.goto('/groups/by-id/6688692310502473984;p=/settings');
-  await expect(page.getByRole('heading', { name: 'Required approvals' })).toBeVisible();
-  await expect(page.getByText('Lock membership until a given date')).toBeVisible();
-  await expect.soft(page.getByTestId('switch-require-lock-until-enabled')).toBeVisible();
-  await page.getByTestId('switch-require-lock-until-enabled').click();
-  await expect.soft(page.getByTestId('require-lock-until-datepicker')).toBeVisible();
-  await expect(page.getByText('Managers can access member\'s personal information')).toBeVisible();
-  await expect(page.locator('li').filter({ hasText: 'Read only' })).toBeVisible();
-  await page.locator('li').filter({ hasText: 'Read only' }).click();
+
+  await test.step('Checks the lock membership until a given date control is visible', async () => {
+    await expect(page.getByRole('heading', { name: 'Required approvals' })).toBeVisible();
+    await expect(page.getByText('Lock membership until a given date')).toBeVisible();
+    await expect.soft(page.getByTestId('switch-require-lock-until-enabled')).toBeVisible();
+  });
+
+  await test.step('Enable lock membership until a given date control and fill the date', async () => {
+    await page.getByTestId('switch-require-lock-until-enabled').click();
+    const inputDateLocator = page.getByTestId('input-date').getByRole('textbox');
+    await expect.soft(inputDateLocator).toBeVisible();
+    await inputDateLocator.fill(convertDateToString(new Date(Date.now() + DAYS)));
+  });
+
+  await test.step('Switch managers can access member\'s personal information to "Read only"', async () => {
+    await expect(page.getByText('Managers can access member\'s personal information')).toBeVisible();
+    await expect(page.locator('li').filter({ hasText: 'Read only' })).toBeVisible();
+    await page.locator('li').filter({ hasText: 'Read only' }).click();
+  });
+
+  await test.step('Save changes and wait notification of success', async () => {
+    await page.getByRole('button', { name: 'Save' }).click();
+    await expect.soft(page.getByText('SuccessChanges successfully')).toBeVisible();
+  });
+});
+
+test('Turn off approval rules', async ({ page }) => {
+  await initAsUsualUser(page);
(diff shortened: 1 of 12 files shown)


smadbe (changes requested):
On my laptop, "small" screen, the date picker is not really usable, the year and month selectors are out of the screen, but if I scroll the date picker disappear.
https://github.com/France-ioi/AlgoreaFrontend/assets/1053150/f505a2ed-42ce-4eef-86ed-e0d7e2828306

src/app/groups/containers/group-edit/group-edit.component.ts
smadbe, line 1: A priori, I would have said that it would have been better out of this file as this file become long enough. But now that has been done...

e2e/groups/group-edit-requires-approval.spec.ts
smadbe, line 1: See [how I split the test so that they are more readable](https://github.com/France-ioi/AlgoreaFrontend/blob/master/e2e/items/item-routing.spec.ts#L61), by splitting in logic step, grouping set of action under a named step.
Your tests are more a monolith of rules that cannot be easily understood.

Iloveall: Fixed

e2e/groups/group-edit-requires-approval.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    await inputDateLocator.fill(convertDateToString(currentDate));
+    console.log('currentDate', currentDate);
+    console.log('currentDate', convertDateToString(currentDate));
+    await expect.soft(page.getByText(`Minimum date: ${ convertDateToString(currentDate) }`)).toBeVisible();
+  });
+
+  await test.step('Clear invalid date and check validation message is gone', async () => {
+    await inputDateLocator.fill('');
+    await expect.soft(page.getByTestId('input-date').getByPlaceholder('dd/mm/yyyy hh:mm')).toBeVisible();
+    await page.getByText('Lock membership until a given date').click();
```
Iloveall, line 101: Blur focus

src/app/ui-components/input-date/input-date.component.ts
```diff
@@ -0,0 +1,100 @@
...
+      this.control = this.injector.get(FormGroupDirective).getControl(injectedControl);
+    }
+  }
+
+  ngOnDestroy(): void {
+    setTimeout(() => {
+      this.control?.clearValidators();
+      this.control?.updateValueAndValidity();
+      this.control?.patchValue(null);
+    });
```
Iloveall, line 66: For my point when the control component has removed from parent form component - it should remove value and validation (primeng didn't that), otherwise it need to be handled in parent component which makes reusing this component complicated

But without timeout it doesn't work + it throws `ExpressionChangedAfterItHasBeenCheckedError` error

src/app/groups/containers/group-edit/group-edit.component.ts
Iloveall: I was thinking about separated control, but went the way we have. Definitely in future need to be improved

e2e/groups/group-edit-requires-approval.spec.ts
```diff
@@ -0,0 +1,104 @@
+import { test, expect } from '@playwright/test';
+import { initAsUsualUser } from '../helpers/e2e_auth';
+import { convertDateToString } from 'src/app/utils/input-date';
+import { DAYS } from 'src/app/utils/duration';
+
+test.describe.configure({ mode: 'serial' });
+
+test.use({
+  timezoneId: 'Africa/Accra',
+});
```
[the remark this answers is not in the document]
smadbe: I don't understand why. What was failing? If the website only works on UTC, that's a problem :-) so I would have had the tendency to avoid UTC for the tests

smadbe (changes requested):
Another question.
If I disable and re-enable the "lock", I get that (it is 10:53):
Screenshot 2024-05-29 at 10 53 58

- Why 8:53 while it is 10:53? Is the date configured in UTC? Should be in local time ?
- Are we correctly getting and setting date TZs? 
- If I quickly disable and re-enable lock, I would expect to keep the same date as configured before. (in "background", so date should not be nullified, it just be shown/hidden depending on the switch)

Actually, I modify add one minute to this "default time" I get the red message that the date is too low... so it shows there is problem.

e2e/groups/group-edit-requires-approval.spec.ts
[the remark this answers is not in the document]
Iloveall: Fixed

Iloveall:
Also need to add @no-parallelism label for tests

smadbe (changes requested):
Do we consider this behavior as "normal"?  (basically, saving it enabled without valid date) 
https://github.com/France-ioi/AlgoreaFrontend/assets/1053150/82026893-4bf0-433a-b07e-2fcfbd233388

src/app/groups/containers/group-edit/group-edit.component.ts
```diff
@@ -172,4 +213,21 @@ export class GroupEditComponent implements OnInit, OnDestroy, PendingChangesComp
...
+
+  updateRequireLockMembershipApprovalUntilValueAndValidity(): void {
+    const control = this.groupForm.get('requireLockMembershipApprovalUntil');
+    if (!control) throw new Error('Unexpected: Missed requireLockMembershipApprovalUntil control');
+    const initialRequireLockMembershipApprovalUntil = this.initialFormData?.requireLockMembershipApprovalUntil;
+    if (control.value !== null && initialRequireLockMembershipApprovalUntil) {
+      const currentDate = new Date();
+      this.minLockMembershipApprovalUntilDate = initialRequireLockMembershipApprovalUntil < currentDate
+        ? initialRequireLockMembershipApprovalUntil : currentDate;
+      // eslint-disable-next-line @typescript-eslint/unbound-method
```
Iloveall, line 225: Could be remove after merge with https://github.com/France-ioi/AlgoreaFrontend/pull/1697/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R106

src/app/ui-components/form-error/form-error.component.html
```diff
@@ -5,4 +5,10 @@
   <div *ngIf="control.errors.min as err"><small i18n>Invalid value, minimum: {{ err.min }}</small></div>
   <div *ngIf="control.errors.max as err"><small i18n>Invalid value, maximum: {{ err.max }}</small></div>
   <div *ngIf="control.errors.invalidDuration as err"><small i18n>Invalid duration</small></div>
+  @if (control.errors.inputDateInvalid; as err) {
+    <small i18n>Invalid date</small>
+  }
+  @if (control.errors.inputDateMinInvalid; as err) {
+    <small i18n>The date must be grater than: {{ err.minDate | date: 'dd/MM/yyyy HH:mm' }}</small>
```
smadbe, line 12:
```suggestion
    <small i18n>The date must be greater than: {{ err.minDate | date: 'dd/MM/yyyy HH:mm' }}</small>
```

smadbe (approved):
should be good 👍

PR #1727, opened 2024-07-11
move item datasource logic into store

smadbe:
## Description

Tried to do the minimal (yes minimal ;-)) change to move the item/breadcrumbs/result fetching from the item data source to the store.

I have also added e2e tests covering that. (and unit tests for effects)

## Notes

This requires to add a new concept of "identifier" on the fetch state, which tags for what content is this state. I thought about it already in the past for other usages but here it ensures the state that we select matches the current content (select from router, not via a reducer that we control)

## Test cases

- [ ] Case 1:
  1. Given I am the demo user
  2. When I go to [a chapter](https://dev.algorea.org/branch/item-content-fetching-in-store/en/a/944619266928306927;p=7528142386663912287;a=0)
  4. Then the behavior looks correct
  2. When I go to [a regular task](https://dev.algorea.org/branch/item-content-fetching-in-store/en/a/6379723280369399253;p=7528142386663912287,7523720120450464843;a=0)
  4. Then the behavior looks correct
  2. When I go to [a regular task with a wrong path](https://dev.algorea.org/branch/item-content-fetching-in-store/en/a/6379723280369399253;p=404,7523720120450464843;a=0)
  4. Then the behavior looks correct
  5. When I go to [a explicit-entry content](https://dev.algorea.org/branch/item-content-fetching-in-store/en/a/1480462971860767879;p=7528142386663912287,944619266928306927;pa=0)
  4. Then the behavior looks correct
  5. When I go to a [locked content](https://dev.algorea.org/branch/item-content-fetching-in-store/en/a/1667525548838093512;p=7528142386663912287,944619266928306927;a=0)
  4. Then the behavior looks correct
  
- [ ] Case 2:
  1. Given I am a temp user
  2. When I go to [a chapter](https://dev.algorea.org/branch/item-content-fetching-in-store/en/a/home;a=0)
  4. And I visit a few chapters
  5. Then the behavior looks correct (the results start as expected)

- [ ] Case 3
  1. Given I am a temp user
  2. When I go to [a content I cannot see](https://dev.algorea.org/branch/item-content-fetching-in-store/en/a/1667525548838093512;p=7528142386663912287,944619266928306927;a=0)
  3. I see an error message which invites me to go back to the home page


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/item-content-fetching-in-store/playwright-report/index.html

e2e/items/item-fetching.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  await page.goto('/a/7523720120450464843;p=7528142386663912287;a=0');
+
+  await test.step('check breadcrumbs', async () => {
+    await expect.soft(page.locator('.breadcrumb-list li:nth-child(1)')).toContainText('Algorea Testing Content for devs');
+    await expect.soft(page.locator('.breadcrumb-list li:nth-child(3)')).toContainText('Tasks Showcase');
+  });
+
+  await test.step('check content', async () => {
+    await expect.soft(page.locator('h1.item-title')).toContainText('Tasks Showcase');
+    await expect(page.locator('.right-container').getByText('Blockly Basic Task')).toBeVisible();
```
Iloveall, line 20: I would suggest to use testid in future

src/app/items/services/result-fetching.service.ts
```diff
@@ -0,0 +1,50 @@
...
+import { FullItemRoute, isRouteWithSelfAttempt } from 'src/app/models/routing/item-route';
+import { GetResultsService, Result } from '../data-access/get-results.service';
+import { EMPTY, Observable, map, of, switchMap } from 'rxjs';
+import { canCurrentUserViewContent } from '../models/item-view-permission';
+import { bestAttemptFromResults, implicitResultStart } from '../models/attempts';
+import { ResultActionsService } from 'src/app/data-access/result-actions.service';
+
+export function canFetchResults(item: Item): boolean {
+  return canCurrentUserViewContent(item);
+}
```
Iloveall, line 12: For better naming or suppose to be extended ?

Iloveall (approved):
First I was thinking that the selectors doing a bit a lot, but then I've found this article which has a good arguments https://medium.com/@thomas.laforge/ngrx-effect-vs-reducer-vs-selector-58337ab59043

e2e/items/item-fetching.spec.ts
smadbe: Ok, I agree. For, the right container, I feel it is still somehow valid. And it is still use at other places... so maybe a task maybe to fix all locators across the tests.

src/app/items/services/result-fetching.service.ts
smadbe: It is just that for the external point of view, we just want to know if we "can fetch results". 
And basically, it is "computed" (currently at least) as  whether the current user can view content. 
If I use `canCurrentUserViewContent` directly in the selector, the reader cannot understand why I use this function.

smadbe:
> First I was thinking that the selectors doing a bit a lot, but then I've found this article which has a good arguments https://medium.com/@thomas.laforge/ngrx-effect-vs-reducer-vs-selector-58337ab59043

Selectors are quite light.. it is just that they check that the current content is what we expect before returning the value. Then there is these 2 selectors aggregating several data, they are there for backward compat mainly, I didn't want to change directly all uses of `itemData`.

PR #1757, opened 2024-08-19
Item-content cleanup

smadbe:
## Description

Just cleanup the "item-content" component which is a big mess and where it is difficult to understand rules.

Use new control flows and signals; use proper pipes; fix the subject which wasn't released.

Thanks to the e2e tests on item-content, it looks the behavior hasn't changed.


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/item-content-cleanup/playwright-report/index.html

PR #1740, opened 2024-08-02
Allows to see Groups tab for all users

Iloveall:
## Description

Fixes #...

## Notes (out of scope, known isues, hints for reviewing code, ...)  (optional)

...

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](branch link)
  3. And I click on ...
  4. Then I see ...
  5. And

- [ ] Case 2:
  1. Given I am the usual user
  2. When ...

Issue #1735, opened 2024-07-25
Show "groups" even when not connected


smadbe:
## Overview

The goal is let, even anonymous/temp users, understand there are groups and they have to be connected to enter groups.

## Acceptance criteria
- [ ] Always display the "Groups" tab in the left menu
- [ ] The "My groups" page should just contain the title+subtitle and then a large (like usual (typically error) message) text displayed : "To join a group, you first have to log in or sign in using the top right button." using the login icon as icon.
- [ ] In practice, there is no case anymore when the left menu tabs are not shown, you can probably remove the logic around that

PR #1766, opened 2024-08-26
fix wrong left menu visibily condition when skills not displayed

smadbe:
## Description

Fix wrong condition when skills are disabled. That causes the tabs not to be displayed at all in this case.

Bug introduced by #1740. The [request issue](https://github.com/France-ioi/AlgoreaFrontend/issues/1735) mentioned " In practice, there is no case anymore when the left menu tabs are not shown, you can probably remove the logic around that".


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/fix-left-menu-no-skills/playwright-report/index.html

PR #1797, opened 2024-09-30
Fix too many redirections

smadbe:
## Description

We had in the crash reporting some "Too many redirections (unexpected)" reported from time to time for months. That looked like a bad smell as it is "unexpected".

Actually, the previous condition based on `this.itemState$.pipe(` was wrong as we may two successive "subscribe" without being the breadcrumbs failing twice, but just failing once and having another part of the state (item, results) changing. So it was wrongly considered as two failures.

Also, now, `hasRedirected` is resetted as soon as breadcrumb returns something else than a forbidden/not-found response, not just with a full success of the state.


smadbe:
E2e tests are not passing, but that looks like the usual problem. Isn't it?

Iloveall (approved):

PR #1805, opened 2024-10-07
E2e tests/resources

Iloveall:
## Description

- Fixes race conditions;
- Splits tests on groups chromium, firefox - to decrease time of execution and possible re run, also it would reduce loading on docker resource. 
- Updates playwright packages
- Moves simple tests to new test user - requires to move more cases but will take some time so in next PR
- Increased timeout - for possible cases when long waiting of multiple response before render needed part for test

Generally the issues with tests is race condition, it works very fast and can catch some cases the real user can't (kind filled form before BE returned all needed data). We need to see future errors and improve step by step our tests. Also I realised that need to test new cases with extra loading on BE (ex. 4-6 workers for 1 test case) and so when BE became in hot state and works faster - in result we getting unexpected behaviour.

### DIFF
--- a/e2e/groups/view-answer.spec.ts
+++ b/e2e/groups/view-answer.spec.ts
@@ -129,7 +129,7 @@ test('View answer in logs for current user', async ({ page }) => {
   await expect.soft(page.getByRole('link', { name: 'View answer' })).toBeVisible();
   await page.getByRole('link', { name: 'View answer' }).click();
   await expect.soft(page.locator('h1').getByText('Task with edit tab')).toBeVisible();
-  expect(page.url()).toContain('answerId=6216264238928088073');
+  await expect.soft(page).toHaveURL(new RegExp('answerId=6216264238928088073'));
 });
 
 test('View answer in logs for other user', async ({ page }) => {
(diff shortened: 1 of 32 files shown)


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/e2e-tests/resources/playwright-report/index.html

e2e/groups/view-answer.spec.ts
```diff
@@ -129,7 +129,7 @@ test('View answer in logs for current user', async ({ page }) => {
   await expect.soft(page.getByRole('link', { name: 'View answer' })).toBeVisible();
   await page.getByRole('link', { name: 'View answer' }).click();
   await expect.soft(page.locator('h1').getByText('Task with edit tab')).toBeVisible();
-  expect(page.url()).toContain('answerId=6216264238928088073');
+  await expect.soft(page).toHaveURL(new RegExp('answerId=6216264238928088073'));
```
smadbe, line 132: I guess you could write
```suggestion
  await expect.soft(page).toHaveURL(/answerId=6216264238928088073/);
```
which would be more readable (applies everywhere)

smadbe (approved):

e2e/groups/view-answer.spec.ts
Iloveall: The reason I'm using it is the path with `/`, example:
`new RegExp('/a/home;pa=0')` will be `/\/a\/home;pa=0/`, would be difficult to copy paste

smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/e2e-tests/resources/playwright-report/index.html

smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/e2e-tests/resources/playwright-report/index.html

PR #1787, opened 2024-09-19
Adds supporting guard form changes for tabs

Iloveall:
## Description

Fixes #1767

## Notes (out of scope, known isues, hints for reviewing code, ...)  (optional)

The problem is the tab service switching view directly without router

## Test cases

Issue #1767, opened 2024-08-26
Bug: changes lost when changing tabs from parameters to dependencies


smadbe:
**Describe the bug**
The changes are lost when changing tabs from parameters to dependencies. The "Confirm Navigation" modal is displayed after the navigation. 

**To Reproduce**
1. As the usual user
2. Go to the a parameter page https://dev.algorea.org/en/a/39530140456452546;p=4702,4102,1980584647557587953;a=0/parameters
3. Click add "!" to the title
4. Change to the "dependencies" tab
5. Select "no" into the modal to "Do you want to leave this page and lose its changes?"
6. Check in the parameter tab that the change has been lost

It is important to add a E2E test here to prevent re-breaking that without noticing.

PR #1806, opened 2024-10-07
fix tab handling 1803 and 1787

smadbe:
## Description

Fix tab handling: the problem #1767 fixed by #1787 and also the problem #1787 introduces.

As tab changes is quite complex, especially with tasks, I am not completely sure all cases are ok. Tests I can think of (that I should have test but for which a e2e test might be useful):
- on front page without prior navigation (chapter)
- on task: after loading the first tab is selected
- from first task tab to another
- from a task tab to history
- from history a task tab
- site loaded on the history page url
- site loaded on the history page url, then switch to the content tab which loaded custom tabs
- task tabs appearing/disappearing on resize: check task selection is consistent
- leaving the form with changes: tested through the existing e2e


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/fix-item-tabs/playwright-report/index.html

Iloveall (approved):

PR #1812, opened 2024-10-18
Crash reporting: avoid several dialog

smadbe:
## Description

Apparently, if several errors are triggered, several crash reporting dialog were showing up, on top of each other so not allowing the user to enter his feedback.
So now: 
- the dialog is not displayed if a dialog is already being displayed
- the dialog is not redisplayed if the same error happens several time (was already done by sentry but needs to be reimplemented with the previous change)

## Test cases

I had to cause crashes in a branch to trigger that :-/


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/crash-reporting-avoid-several-dialog/playwright-report/index.html

Iloveall (approved):

PR #1814, opened 2024-10-18
allow non-users (teams) in log view

smadbe:
## Description

The log service returns a optional "user" property but we were using it as it was always given...
It is case we cannot really reproduce on our data and that we cannot really create with the UI.. with in prod legacy data, it happens... 
So the fix just display the team name without link or anything... the goal for now is just that it does not crash anymore.


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/fix-allow-non-user-in-logs/playwright-report/index.html

Iloveall (approved):

PR #1819, opened 2024-10-24
save the current answer on submit

smadbe:
## Description

Save the current answer on submit.

Test:
on https://dev.algorea.org/branch/save-current-answer-on-submit/en/a/3862371645338198698;p=4702,7528142386663912287,7523720120450464843;pa=0
when I submit, I see there is a PUT on the current answer.


smadbe:
I think I will do it differently.

PR #1823, opened 2024-10-28
Answer handling improvements

smadbe:
## Description

Fixes several issues with answer saving:
1) if there is a change when changing task, it should be saved (was not if we hadn't waiting the 60sec delay)
2) on submit, there should be a current answer save (was not if we hadn't waiting the 60sec delay)

Recommendation: test it on https://dev.algorea.org/branch/answer-handling-improvements/en/a/773396873861865944;p=4702,7528142386663912287,7523720120450464843;a=0

You can see if a change as been saved because 1) you see the call (PUT), 2) if you refresh the page, you see the change

## Test cases
- there is no save at task launch
- if there is change immediately, it is saved by autosave
- if there one failed auto save (every 60sec), a message is displayed, but it is  retried at the next interval
- if there is no change, no autosave is done 
- submitting triggers a current state save
- changing task triggers a (visible) current state save when the state has changed
  - if this save fails, the user can choose to retry
  - if this save fails, the user can choose to loose changes
- changing task does not trigger a current state save when the state has not changed
  - even just after the app was launched (past bug: was saving just after app launch if even if no change)
- on a task which cannot be loaded (e.g., https://dev.algorea.org/branch/answer-handling-improvements/en/a/1414822370876733593;p=4702,100575556387408660,1788359139685642917;pa=0), the next-prev navigation still work (may be blocked by unload protection if buggy)
- if I change task while the autosave is request is running, the "current" answer must still have been saved somehow

### DIFF
--- a/src/app/items/models/answers.ts
+++ b/src/app/items/models/answers.ts
@@ -1,0 +1,7 @@
+import { Answer } from '../data-access/get-answer.service';
+
+type StateAnswer = Pick<Answer, 'answer'|'state'>;
+
+export function areStateAnswerEqual(x: StateAnswer, y: StateAnswer): boolean {
+  return (x.answer ?? '') === (y.answer ?? '') && (y.state ?? '') === (y.state ?? '');
+}
--- a/src/app/items/models/answers.ts
+++ b/src/app/items/models/answers.ts
@@ -3,5 +3,5 @@ import { Answer } from '../data-access/get-answer.service';
 type StateAnswer = Pick<Answer, 'answer'|'state'>;
 
 export function areStateAnswerEqual(x: StateAnswer, y: StateAnswer): boolean {
-  return (x.answer ?? '') === (y.answer ?? '') && (y.state ?? '') === (y.state ?? '');
+  return (x.answer ?? '') === (y.answer ?? '') && (x.state ?? '') === (y.state ?? '');
 }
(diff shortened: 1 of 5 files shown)


src/app/items/models/answers.ts
```diff
@@ -0,0 +1,7 @@
+import { Answer } from '../data-access/get-answer.service';
+
+type StateAnswer = Pick<Answer, 'answer'|'state'>;
+
+export function areStateAnswerEqual(x: StateAnswer, y: StateAnswer): boolean {
+  return (x.answer ?? '') === (y.answer ?? '') && (y.state ?? '') === (y.state ?? '');
```
Iloveall, line 6: Incorrect condition: `(y.state ?? '') === (y.state ?? '')`

Iloveall:
When saving, and you click navigate to another item - it will save again as lot as you clicking

https://github.com/user-attachments/assets/9cd07cae-eebd-47a0-a03f-2f139cc55f66

smadbe:
> When saving, and you click navigate to another item - it will save again as lot as you clicking
> Screen.Recording.2024-11-06.at.5.14.19.PM.mov

I thought about it. It is not a big deal and it would be quite difficult to fix. So let's keep it as it is.

src/app/items/models/answers.ts
smadbe: :+1: fixed

PR #1824, opened 2024-10-29
Adds overflow param for message info, adds text overflow for observat…

### LINKED ISSUE #1817: Do not overflow "messages"
## Overview

The message boxes (blue/yellow) have their text overflowing when the window is to narrow to fit everything. Fix that.
image

image

## Acceptance criteria
- [ ] Add a parameter which determine what should be the behavior of the message box: multiline or overflow 
- [ ] Implement both: 
  - [ ] a flexible height for message so that we can see the full message when a narrow screen
  - [ ] overflow cut for case such as the observation widget in the top bar
- [ ] check all usage to set the right parameter

Iloveall:
## Description

Fixes #1817

1. The observation bar doesn't use message info component, I've fixed it seprated
2. Overflow type I've tested locally as we have no cases in app:

Screenshot at Oct 29 16-44-31

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  4. When I go to [this page](https://dev.algorea.org/branch/feature/upgrade-message-info/en/groups/by-id/8894246582084793867;p=?watchedGroupId=8894246582084793867)
  5. Then I see truncated text in observation bar

### DIFF
--- a/src/app/containers/observation-bar-with-button/observation-bar-with-button.component.scss
+++ b/src/app/containers/observation-bar-with-button/observation-bar-with-button.component.scss
@@ -1,5 +1,9 @@
 @import 'src/assets/scss/functions';
 
+:host {
+  overflow: hidden;
+}
+
 .container {
   display: flex;
   justify-content: flex-start;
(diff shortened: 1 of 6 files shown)


smadbe (approved):
Ah I forgot there is the case where there is scrolling as well (as for "As a manager of
Test 1, Test 2, and Test 3 you can view this group."). I don't really know how to do one or the other... I would have expected 3 different classes (with one default).. .but probably that is something like that which has been done. :+1:

smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/feature/upgrade-message-info/playwright-report/index.html

PR #1840, opened 2024-11-23
Fixes e2e issues with input date

Iloveall:
## Description

Fixes e2e issue with input date

The problem was with mask that put placeholder as value. As playwright working very fast (unlike human) sometimes it filled the value before p-input emits mask and as result the mask override the value. For self, I can see the best way how to reproduce kind race condition issues in future - is the run tests 50-100 times

image


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/e2e/input-date/playwright-report/index.html

smadbe (approved):

PR #1841, opened 2024-11-27
Fixes navigation for other group pages

Iloveall:
## Description

Fixes navigation for groups without `id`

...


smadbe:
fixed in #1842

PR #1842, opened 2024-11-28
Fix group navigation when going to the mine/managed pages

smadbe:
## Description

Handle correctly the non-content group pages in the group routing process.

+ use constants for user/group contentType and path parts.


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/fix-group-nav/playwright-report/index.html

PR #1845, opened 2024-11-29
"automatic" Observation on groups

smadbe:
## Description

Change of behavior: Now when arriving a group page that we can observe, we enter automatically in observation mode on that group. If we change change group, the observed group changed. If we switch to a group we cannot watch, it leaves observation mode.

@Iloveall  Could you review the code (better commit per commit in this case) and fix the tests? 
If not existing yet, there should be tests covering that:
- when going to our own user page, it does not show observation
- when going to a group we can observe,, observation is enabled,
- when switch group,, observation is changed to that new group,
- when going to a group we cannot observe from a group we can observe, observation is disabled,
- when going to a user page, there is observation mode on that user


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/observing-auto-on-group/playwright-report/index.html

smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/observing-auto-on-group/playwright-report/index.html

smadbe:
> The only small thing I've found is with cross icon when switching between group

That's not a bug, that's because you cannot leave observation of group G if you are on the page of that group.  So when that is still the former group, there is the cross but when we are on the page, we cannot leave it. We'll see if we improve that behavior in the future.

PR #1865, opened 2025-01-13
"Load answer as current" as a state

smadbe:
## Description

Change the way "load answer as current" is handled. Don't pass it in the route anymore as it requires removing the parameter immediately after parsing. That makes the route handling quite complex.

(Part of continuous improvement / refactoring of routes etc)

Open question: if we have a pet task (typically a task where you enter the score you want to get), would it become easy to have an e2e test for that? @Iloveall 

## Test cases

- [ ] Case 1:
  1. Given I am a temp user
  2. When I go to [the blockly test task](https://dev.algorea.org/branch/loadAsCurrent-as-state/en/a/6379723280369399253;p=4702,7528142386663912287,7523720120450464843;a=0)
  3. Then I solve the 1 star problem
  4. Then I solve the 2 stars problem
  5. Then I go the "History" tab and reload the solution which scored 25pts
  6. And I see I am back at the state without the 2 starts solved

### DIFF
--- a/e2e/items/item-logs.spec.ts
+++ b/e2e/items/item-logs.spec.ts
@@ -42,7 +42,7 @@ test('checks reload link in logs on task page', async ({ page }) => {
   await expect.soft(reloadAnswerLocator).toBeVisible();
   await Promise.all([
     reloadAnswerLocator.click(),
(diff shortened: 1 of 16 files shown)


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/loadAsCurrent-as-state/playwright-report/index.html

PR #1885, opened 2025-01-27
Item route cleanup2

smadbe:
## Description

Few small refactoring to item route (to be reviewed one commit at a time): 
- improve how the current page is obtained 
- use only array of strings for item pages
- clarify usage of url vs command in item router
- move functions related to default item in a specific file


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/item-route-cleanup2/playwright-report/index.html

PR #1907, opened 2025-03-12
add more debug info so that we can understand better crash reports

smadbe:
Just add more debug info so that we can understand better crash reports


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/add-debug-info-on-task-config-change/playwright-report/index.html

PR #1922, opened 2025-04-29
Adds input for add extra time in item

### LINKED ISSUE #1915: Allow changing user extra time
## Overview
Currently, it is possible to view user extra time, but not to change it. So provide a way to change it.

## Acceptance criteria
On the extra-time page (for instance https://telecom-paris.france-ioi.org/en/a/1480462971860767879;p=4702,7528142386663912287,944619266928306927;a=0/extra-time as the usual user): (this is a proposal of UI, fell free to suggest something ele)
- [ ] Transform the "Participant-specific" column into, for each row, an input field with the current value.
- [ ] When the value is changed from the initial value, a ✅ validation button, and a cancel ❌ button are shown next to it. And the value in the total column is removed for that row.
- [ ] When the cancel button is pressed, the value is restored to its initial value (so the action of the changes are reverted)
- [ ] When the validation button is pressed:
    - [ ] the input field, and buttons are disabled
    - [ ] a call to the `itemSetAdditionalTime` service is done to change the value
    - [ ] on success, a green toast is displayed and the full table is refreshed.
    - [ ] on failure, a red toast is displayed

Iloveall:
## Description

Fixes #1915

...

## Test cases

Branch for [test](https://dev.algorea.org/branch/feature/allow-changing-user-extra-time/en/a/1480462971860767879;p=4702,7528142386663912287,944619266928306927;a=0/extra-time?watchedGroupId=672913018859223173)

### DIFF
--- a/src/app/items/containers/item-extra-time/item-extra-time-input/item-extra-time-input.component.ts
+++ b/src/app/items/containers/item-extra-time/item-extra-time-input/item-extra-time-input.component.ts
@@ -1,4 +1,4 @@
-import { Component, effect, input, model, output } from '@angular/core';
+import { Component, effect, input, output, signal } from '@angular/core';
 import { FormsModule } from '@angular/forms';
 import { InputNumberModule } from 'primeng/inputnumber';
 import { ButtonIconComponent } from 'src/app/ui-components/button-icon/button-icon.component';
@@ -16,7 +16,7 @@ import { ButtonIconComponent } from 'src/app/ui-components/button-icon/button-ic
 })
 export class ItemExtraTimeInputComponent {
   initialValue = input.required<number>();
-  value = model<number | null>(null); // The p-inputNumber value is nullable
+  value = signal<number | null>(null); // The p-inputNumber value is nullable
   disabled = input(false);
   saveEvent = output<number>();
 


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/feature/allow-changing-user-extra-time/playwright-report/index.html

smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/feature/allow-changing-user-extra-time/playwright-report/index.html

smadbe (changes requested):
If you test it, you immediately realize there is a problem: currently, there is currently "0s" participant specific, and 10sec total... if I change the participant specific value, it is the total value which is changed and the participant specific stays at "0sec". So clearly, you are changing the group additional time and not the participant one. 
If I check the PUT request, that's indeed what's happening, the participant id does not enter into account in the request.

smadbe:
> Yes, i've noticed it, but wasn't sure is it business logic or bug.
If you noticed it, why don't you write it down in the PR description at least?

> Is the issue on my side or BE?
I saw the PUT service that is called only have the top group in parameters, not the user... so I assume you call the service with the wrong parameters.

smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/feature/allow-changing-user-extra-time/playwright-report/index.html

smadbe:
Would it be easy to make pressing 'Enter' in the input field trigger a save of the value?

src/app/items/containers/item-extra-time/item-extra-time-input/item-extra-time-input.component.ts
```diff
@@ -0,0 +1,32 @@
...
+  imports: [
+    FormsModule,
+    InputNumberModule,
+    ButtonIconComponent,
+  ],
+})
+export class ItemExtraTimeInputComponent implements OnChanges {
+  value = model.required<number>();
+  disabled = input(false);
+  defaultValue = signal(0);
```
smadbe, line 20: I would call it `savedValue`, it would make thing clearer I think.

src/app/items/containers/item-extra-time/item-extra-time-input/item-extra-time-input.component.ts
```diff
@@ -0,0 +1,32 @@
...
+  saveEvent = output<number>();
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (!changes['disabled'] || changes['value']) {
+      this.defaultValue.set(this.value());
+    }
+  }
```
smadbe, line 27: You should not use `ngOnChange` anymore with signals. But in this case, I expect this "savedValue" to be updated by the parent on refresh... so that should not be required to update the input from inside the component (bad smell anyway)

smadbe (changes requested):

smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/feature/allow-changing-user-extra-time/playwright-report/index.html

src/app/items/containers/item-extra-time/item-extra-time-input/item-extra-time-input.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  standalone: true,
+  imports: [
+    FormsModule,
+    InputNumberModule,
+    ButtonIconComponent,
+  ],
+})
+export class ItemExtraTimeInputComponent {
+  initialValue = input.required<number>();
+  value = model<number | null>(null); // The p-inputNumber value is nullable
```
smadbe, line 19: In that option, I think the point was that value was no longer an input... so no longer a model but just a signal.

smadbe (changes requested):

src/app/items/containers/item-extra-time/item-extra-time-input/item-extra-time-input.component.ts
smadbe: I have just tested it that works.
```suggestion
  value = signal<number | null>(null); // The p-inputNumber value is nullable
```

smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/feature/allow-changing-user-extra-time/playwright-report/index.html

smadbe (approved):

PR #1941, opened 2025-06-04
Add ancestors to group manager table

smadbe:
## Description

It was hard to realize what were the managers of a group as only the manager of the current group were displayed but not managers from ancestors groups, while they also act as manager on the current group.

So using what the existing backend service offers, we now also display the managers from ancestors groups. If a manager has an higher permission level from ancestors than what it has on the current group (including it has none on the current group), what the manager gets from the ancestor vs current group is displayed.

The first commits corresponds to general cleaning around that component. Only the last 2 commits are about the feature.

Note that the row selection is a bit weird/buggy when some rows not selectable, but let's say it is not crucial for this not-very used page. Also, the design of that new feature is not great, don't hesitate to improve if you have an idea.

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go on a [group with no manager with higher perm through ancestors](https://dev.algorea.org/branch/group-managers/en/groups/by-id/4628755348638868810;p=460263955289148254/managers)
  4. Then that behaves exactly as before

- [ ] Case 2:
  1. Given I am the usual user
  2. When I go on a [group with higher perm through ancestors](https://dev.algorea.org/branch/group-managers/en/groups/by-id/4322568739774621128;p=460263955289148254/managers)
  4. Then that I see that some manager have the ancestor permission indicated as well.


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/group-managers/playwright-report/index.html

PR #1946, opened 2025-06-11
merge current user page with other's user page. Adapt content.

smadbe:
## Description

- merge the current user page with the other's user page. 
- adapt the header to display a message depending on the personal info perm
- only display the firstname/lastname max
- translated to fr 

## Test cases

See e2e. [Test deployment](https://dev.algorea.org/branch/user-page-merge/en/).


smadbe:
@Iloveall  Do you understand the cause of the e2e test failure?

PR #1948, opened 2025-06-12
Fix the caching problem with the left menu.

smadbe:
## Description

Fix the caching problem with the left menu which was causing the left menu not to be refreshed on group creation/deletion. Two changes have been done:
- the cache is invalidated when there is request to force reload the left menu, even if done within the 10sec cache validity
- the http service is called with rxjs `defer`, otherwise the function is executed once and so the cache check is not redone when replaying

## Test cases

- [ ] Case 1:
  1. Given I am the demo user
  2. When I go to [the manage group page](https://dev.algorea.org/branch/leftmenucachebusting/en/groups/manage)
  3. And I create a group "123group"
  4. Then I see it directly in the left menu
  5.  And when I delete it (via settings tab, bottom)
  6. It disappears from the menu directly

### DIFF
--- a/src/app/data-access/group-navigation.service.ts
+++ b/src/app/data-access/group-navigation.service.ts
@@ -6,12 +6,13 @@ import { inject } from '@angular/core';
 import { decodeSnakeCase } from 'src/app/utils/operators/decode';
 import { Cacheable } from 'ts-cacheable';
 import { z } from 'zod';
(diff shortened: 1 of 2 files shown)


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/leftmenucachebusting/playwright-report/index.html

PR #1960, opened 2025-06-25
do not treat forum effects when forum is disabled

smadbe:
## Description

Do not treat forum effects when forum is disabled.


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/disable-forum-effects-when-forum-disabled/playwright-report/index.html

PR #1961, opened 2025-06-26
does not init the current thread state when forum is disabled

smadbe:
## Description

Does not init the current thread state when forum is disabled


smadbe:
[auto-generated message] E2E test report on: https://algorea-static.s3.eu-central-1.amazonaws.com/branch/forum-current-thread-handle-disabled/playwright-report/index.html

PR #1974, opened 2025-08-07
Replaces path suggestion primeng overlay with cdk overlay

### LINKED ISSUE #1873: Migration: Path Suggestion
## Overview

Need to create own component with using angular cdk overlay, ex: https://medium.com/@JoaoPoggioli/creating-a-custom-popover-with-overlay-in-angular-dfb330cfd124

## Acceptance criteria
- [ ] Replace path suggestion logic with angular cdk overlay
- [ ] Migrate styles
- [ ] Remove p-overlayPanel
- [ ] Test

Iloveall:
## Description

Fixes #1873

I didn't put triangle in this version as the position depends on viewport and can be too varius. Also I've added scroll inside path suggestion content, so less chance that the user can't reach content, as it was before.

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/migration-ui/path-suggestion-overlay/en/groups/by-id/9213510114403290476;p=/history)
  3. And I hover on content item in table
  4. Then I see new path suggestion

### DIFF
--- a/src/app/containers/left-search-result/left-search-result.component.html
+++ b/src/app/containers/left-search-result/left-search-result.component.html
@@ -24,10 +24,12 @@
 <ng-container *ngFor="let item of data;">
   <div class="container">
     <ul class="result-list alg-search-result">
-      <li class="result-list-item"
-          [algShowOverlay]="op"
-          (overlayOpenEvent)="itemId.set(item.id)"
-          (overlayCloseEvent)="itemId.set(undefined)"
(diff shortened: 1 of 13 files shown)


smadbe (approved):
That's a bit strange without the "triangle" as we don't really know what "This content can be found in..." is for. But let's say that's ok for now.

PR #1984, opened 2025-08-25
Localize app to German and Italian

smadbe:
Adapt the app so that it can be localized into German and Italian.

And so: 
- change the workflow so that translations are not stored in the project anymore, but managed by an external service where we push our strings and from which we can retrieve the translations
- we do not build other lang than "en" on the branches
- we de retrieve the translation when building master or a version
- refactor the circle CI config file to prevent repetitions of job for each language (use parameter matrices)


smadbe:
[auto-generated message] E2E test report on: https://assets.algorea.org/branch/localize-app-to-de-it/playwright-report/index.html

smadbe:
[auto-generated message] E2E test report on: https://assets.algorea.org/branch/localize-app-to-de-it/playwright-report/index.html

PR #1967, opened 2025-07-02
Fixes extra calls of ng changes in item edit wrapper component

Iloveall:
## Description

Fixes #1956

## Notes (out of scope, known isues, hints for reviewing code, ...)  (optional)

Seems it's old problem (after route store has been implemented). The cancel route action (see on screen) - emits the whole state again. I couldn't find better solution

image

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/bugfix/lost-changes/en/a/6707691810849260111;p=;a=0/parameters)
  3. And I do change the input of title to "Changed"
  4. Then I see confirmation modal
  5. And I click on "Cancel"/"close" button or click on background
  6. Then I still can see "Changed" title in input

PR #1993, opened 2025-09-10
Fix bug on item state

smadbe:
Fix a bug that may cause the item by id to be stuck on fetching if we change route while staying on the same item

It was introduced by https://github.com/France-ioi/AlgoreaFrontend/pull/1967.

Explanation: 
> Actually the fix is not correct, that may create some new bugs in some scenarios (maybe not possible when the code was written): If we stay on the same item page with the same data (navigate from an item to the same item id) but have a route error in the middle (for instance the path or attempt is missing): as the data hasn't changed, it does not re-emit and so state$ stays on what was last emitted from this.store.select(fromItemContent.selectActiveContentRouteErrorHandlingState).pipe(filter(isNotNull) which is a fetching state; and so the item page stays on fetching forever while the actual item state is ready.

Tested: the bug fixed by #1967 did not reappeared.


smadbe:
[auto-generated message] E2E test report on: https://assets.algorea.org/branch/fix-item-state/playwright-report/index.html

PR #1994, opened 2025-09-10
Item log view fixes

smadbe:
2 improvements / fixes: 
- if we stay on the same item, use the already known attempt (instead of doing an extra request to fetching result)
- force navigating to the content tab when reloading an answer


smadbe:
[auto-generated message] E2E test report on: https://assets.algorea.org/branch/item-log-view-fixes/playwright-report/index.html

PR #1999, opened 2025-09-12
on narrow screen, when only the left menu is displayed, hide the menu when selecting the 'join' or 'manage' button

smadbe:
when selecting the 'join' or 'manage' button

## Test cases

- [ ] Case 1:
  1. Given I am any user
  2. I go the website https://dev.algorea.org/branch/close-left-menu-on-summary-group-pages/en/
  3. If I use a narrow screen and display the left menu 
  4. If I select the "join" or "manage" button in the left menu
  5. The left menu closes


smadbe:
> When the screen is narrow and I do click on "Group" tab - it's also closes the menu. Not consistently with other tabs

I am being told that it is already better than before, we'll see if we need to rechange that later.

PR #2019, opened 2025-10-10
Forum: major refactoring

smadbe:
## Description

Major changes on how the forum work.:

- do not push activity events to the forum anymore: all users now fetch the latest activities from the backend directly
- creating a message is now a POST rest service
- getting all previous messages is now GET rest service

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/forum-v2/en/a/7528142386663912287;p=4702;a=0/forum/others)
  3. And I click on the " Task with localization 	usr_5p020x2thuyu " thread
  4. Then I see the messages and activities
  5. And can post messages

### DIFF
--- a/src/app/forum/containers/thread/thread.component.html
+++ b/src/app/forum/containers/thread/thread.component.html
@@ -35,7 +35,7 @@
             </div>
             } @else {
               @if (threadMetadata.id) {
-                @for (event of state().data; track event) {
+                @for (event of state().data; track event.time) {
                   <alg-thread-message
                     class="thread-message"
                     [threadId]="threadMetadata.id"


src/app/forum/containers/thread/thread.component.html
```diff
@@ -1,119 +1,128 @@
...
-                  alg-button
-                  (click)="changeThreadStatus({ open: !thread.threadStatus.data.open, threadId: threadId })"
-                  *ngIf="thread.threadStatus && thread.threadStatus.data"
-                >
-                  <span *ngIf="thread.threadStatus.data.open; else openThreadCaption" i18n>Close this thread</span>
-                  <ng-template #openThreadCaption>
-                    <span i18n>Open this thread</span>
+            } @else {
+              @if (threadMetadata.id) {
+                @for (event of state().data; track event) {
```
Iloveall: As the `mergeEvents` returns new array, I assume that the track by event (so link) won't work here

smadbe:
> > Bug: After sending the message the text area doesn't clean the typed message
> 
> It does not clear out if there is an error. But so I need to know what you did. Ah I see, if you send as the demo user... it's because you get a 5xx from the server, I'll check why..

this one is fixed, it was an issue on the backend side

src/app/forum/models/thread-events.ts
```diff
@@ -1,77 +1,75 @@
...
 export function isSubmissionEvent(event: ThreadEvent): event is SubmissionEvent {
-  return event.label === EventLabel.Submission;
+  return event.kind === EventKind.Submission;
+}
+
+/**
+ * Merge multiple event lists into a single list, sorted by time.
+ */
+export function mergeEvents(eventLists: ThreadEvent[][]): ThreadEvent[] {
+  const events = eventLists.reduce((acc, events) => [ ...acc, ...events ], []);
```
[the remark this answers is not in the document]
smadbe: I didn't know about that JS fonction ;-) :+1:

src/app/forum/containers/thread/thread.component.html
smadbe: The content of the array would be the same objects (as long as there is not a new refresh)... so I assume it should work.
But I can use the `time` as well maybe.

PR #2061, opened 2025-12-09
hide attribute which are not currently supported

smadbe:
## Description
Just hide the item settings which are actually not currently implemented. Now actual users use the platform, it is more clear to hide them.

## Test cases

- [ ] Case 1:
  1. Given I am the usual user
  2. When I go to [this page](https://dev.algorea.org/branch/hide-nonimplemented/en/a/39530140456452546;p=4702,4102,1980584647557587953;a=0/parameters)
  4. Then I see the "not implemented" attribute and the "Team" section has disappeared, without glitches


smadbe:
[auto-generated message] E2E test report on: https://assets.algorea.org/branch/hide-nonimplemented/playwright-report/index.html

